PR #10, opened 2016-03-11
Show Workshop banner on blog post pages

marcoow:


ganeshbsub:
Personally, I think the banner might be nice at the top, just above the Blog date.. Some blog posts are really huge and people might miss the banner.

PR #32, opened 2016-09-07
Add Twitter and FB OG image metadata

ganeshbsub:


_includes/head.html
```diff
@@ -6,6 +6,8 @@
   <meta name="language" content="en"/>
   <meta name="content-language" content="en"/>
   <meta name="publisher" content="simplabs" />
+  <meta content="/images/logo@2x.png" property="og:image"/>
+  <meta name="twitter:image:src" content="/images/logo@2x.png">
```
marcoow, line 10: I guess these should be added by the seo gem? Are we using an old version maybe?

ganeshbsub: I checked the meta data on our site. Everything else(og: title, twitter: title, etc.) but the images are being generated. I tried this explicitly to see if this solved the issue.

ganeshbsub:
Not sure why it still doesn't show an image. https://cards-dev.twitter.com/validator

ganeshbsub:
[This](https://github.com/jekyll/jekyll-seo-tag/issues/84) might be the issue.

Issue #29, opened 2016-09-07
Missing simplabs logo on link previews


oliverbarnes:
Would be nice to have the simplabs logo where the page icon shows up:

![image](https://cloud.githubusercontent.com/assets/21290/18308292/dc446f40-74eb-11e6-9fc6-6f57789eba28.png)

ganeshbsub:
Works now. Apparently, it needed the full path to the image in the meta tag. 

https://twitter.com/digiberber/status/773482613797445632

PR #33, opened 2016-09-07
Fix site preview image

ganeshbsub:


ganeshbsub:
Closes #29

_includes/head.html
```diff
@@ -6,8 +6,8 @@
   <meta name="language" content="en"/>
   <meta name="content-language" content="en"/>
   <meta name="publisher" content="simplabs" />
-  <meta content="/images/logo@2x.png" property="og:image">
-  <meta name="twitter:image:src" content="/images/logo@2x.png">
+  <meta content="http://simplabs.com/images/apple-144.png" property="og:image"/>
+  <meta name="twitter:image" content="http://simplabs.com/images/apple-144.png"/>
```
marcoow, line 10: Do we need to include these manually when we have the above config in `_config.yml`?

PR #38, opened 2016-09-08
Use logo as default OG/Twitter image for pages with no image

ganeshbsub:


_config.yml
```diff
@@ -28,16 +28,26 @@ navigation:
     url: /blog/
 
 title: simplabs
-author: simplabs GmbH
+author: simplabs
```
marcoow, line 31: GmbH is fine actually

ganeshbsub: It resolves into a twitter tag **@simplabs GmbH**. That's why I changed it.

PR #39, opened 2016-09-08
Workaround for htmlproofer

ganeshbsub:


ganeshbsub:
@marcoow This is the only way if we want `html-proofer` to work with `jekyll-seo-tag`, I guess. At least the only way I could think of.

PR #52, opened 2017-01-12
Blog: New features for ember-test-selectors

Turbo87:
[Rendered](https://github.com/simplabs/simplabs.github.io/blob/test-selectors/blog/_posts/2017-01-12-ember-test-selectors.md)


blog/_posts/2017-01-12-ember-test-selectors.md
```diff
@@ -0,0 +1,314 @@
...
+
+#### Automatic binding of `data-test-*` properties
+
+As you may know from our [previous blog post](http://simplabs.com/blog/2016/03/04/ember-test-selectors.html)
+on this topic, the goal of this addon is to let you use attributes starting
+with `data-test-` in your templates:
+
+```handlebars
+{% raw %}<article>
+  <h1 data-test-selector="post-title">{{post.title}}</h1>
```
pangratz: This should be `data-test-post-title` to work with the selector below...

blog/_posts/2017-01-12-ember-test-selectors.md
```diff
@@ -0,0 +1,314 @@
...
+#### Stripping out `data-test-*` attributes in templates
+
+Our initial goal with this library was stripping our `data-test-*` attributes
+from HTML tags in your templates. In the previous section we implemented
+automatic bindings for `data-test-*` properties on components now too, but
+these properties were not stripped from the template yet.
+
+To modify templates from within an addon our best bet was to use an
+<abbr title="Abstract Syntax Tree">AST</abbr> transform on the Handlebars
+AST, that we get from the template parser. This can be accomplished using by
```
pangratz: `s/using//`

blog/_posts/2017-01-12-ember-test-selectors.md
```diff
@@ -0,0 +1,314 @@
...
+      }
+    });
+
+    return ast;
+  }
+}; 
+```
+
+You can try out what this transform does in the
+[AST explorer](https://astexplorer.net/#/5ZDpdTbKwL).
```
pangratz: ❤️

blog/_posts/2017-01-12-ember-test-selectors.md
```diff
@@ -0,0 +1,314 @@
...
+      },
+    },
+  });
+};
+```
+
+With the Babel plugin done, all we had left to do was making sure that your
+app actually uses that plugin at build time. While this is not quite public
+API and may change in the future we have found a way to accomplish that in the
+official [ember-cli-htmlbars-inline-precompile](https://github.com/ember-cli/ember-cli-htmlbars-inline-precompile/blob/master/index.js#L64-L69)
```
pangratz: I think we should use a permalink here...

pangratz: Can be achieved by by pressing `y` when viewing the above link

pangratz (dismissed):
Awesome writeup!! 💯 👍 

Since the post is quite long, I think a TL;DR at the beginning might be a good idea... 🤔

marcoow:
@Turbo87: need to add you [picture](https://github.com/simplabs/simplabs.github.io/tree/master/images/authors) here as well so it renders corretly at the bottom.

blog/_posts/2017-01-12-ember-test-selectors.md
```diff
@@ -0,0 +1,314 @@
...
+author: "Tobias Bieniek"
+github-handle: Turbo87
+twitter-handle: tobiasbieniek
+---
+
+In March 2016 we have released the first version of
+[ember-test-selectors](https://github.com/simplabs/ember-test-selectors) and
+today we are proud to present you with our next milestone: `v0.1.0`.
+
+While `v0.1.0` does not sound like much has changed, the addon has actually
```
marcoow: I think we should not include the `v` prefix in the version numbers in the text actually - it's basically just the tags that use the prefix while users will always deal with the bare versions.

blog/_posts/2017-01-12-ember-test-selectors.md
```diff
@@ -0,0 +1,314 @@
...
+with `data-test-` in your templates:
+
+```handlebars
+{% raw %}<article>
+  <h1 data-test-selector="post-title">{{post.title}}</h1>
+  <p>{{post.body}}</p>
+</article>{% endraw %}
+```
+
+... so that you can use these as CSS/jQuery selectors in your acceptance and integration tests:
```
marcoow: I'd just say _"you can use these as selectors"_

blog/_posts/2017-01-12-ember-test-selectors.md
```diff
@@ -0,0 +1,314 @@
...
+function, which we were now able to [unit test](https://github.com/simplabs/ember-test-selectors/blob/v0.1.0/tests/unit/utils/bind-data-test-attributes-test.js)
+separately.
+
+As we are committed to not including any unnecessary code in your
+production builds, we also had to make sure to not include the
+initializer and utility function in there. Since both of those are part
+of our `addon` and `app` folders, which are included in your builds by
+default, we borrowed a ["trick"](https://github.com/ember-cli/ember-cli-chai/blob/master/index.js#L119-L123)
+from [ember-cli-chai](https://github.com/ember-cli/ember-cli-chai/) which
+only includes both folders if we are in testing mode.
```
marcoow: I think we should be clearer about the fact that this is not only for testing mode but for all modes that we're not stripping in, e.g. also `development` env. Otherwise that might lead people to believe they wouldn't get these things in `development` neither.

blog/_posts/2017-01-12-ember-test-selectors.md
```diff
@@ -0,0 +1,314 @@
...
+If you try the same example template in the
+[AST explorer](https://astexplorer.net/#/8wkrolD3V0), but with the modified
+transform code, you will notice that the `data-test-comment-id=comment.id` part
+of the `some-component` invocation is now gone.
+
+
+#### Stripping out `data-test-*` properties in JS files
+
+Now that we are able to automatically bind `data-test-*` properties, maybe it
+makes sense to also use them inside the class as *computed* properties.
```
marcoow: I'd say sth. like _"While one way of assigning data attributes to a component is in the template, data attributes can also be defined as properties on the component context. Data attributes defined that way are now auto-bound as well and of course also need to be stripped for production (or production-like) builds."_

Your wording sounds a bit like we're suggesting to define data attributes on the template context instead of the template which I don't think we should do. For acceptance and integration tests your perspective is actually more centered on the template instead of the context and it's probably simpler in most cases to define the data attributes there as well then.

blog/_posts/2017-01-12-ember-test-selectors.md
```diff
@@ -0,0 +1,314 @@
...
+  comment: null,
+  'data-test-comment-id': Ember.computed.readOnly('comment.id'),
+});
+```
+
+Unfortunately we have now have a property that is not stripped by the AST
+transform described in the previous section. At this point we could have
+used a similar strategy as before and added a JavaScript preprocessor, that
+strips all those properties from the code, but instead we hooked into the
+existing JavaScript processing pipeline using [Babel](http://babeljs.io/).
```
marcoow: _"There now is a property definition inside of the template context that will not be stripped by the Handlebars AST transform described above as it is defined in JS actually. At this point we could have used a similar strategy a…"_

marcoow (commented):
Great post!

blog/_posts/2017-01-12-ember-test-selectors.md
Turbo87: I tried to make this clearer in the "testing in production" section. Mentioning this including all the details everywhere will probably make it harder to read.

marcoow: I guess you could just say sth. like _"…only includes both folders unless building for production or a production-like environment where none of ember-test-selector's code should be present."_ or so.

Turbo87:
updated the PR with the suggestions

blog/_posts/2017-01-12-ember-test-selectors.md
```diff
@@ -0,0 +1,315 @@
...
+CSS/jQuery selector strings used for `find()` or `this.$()` in your tests.
+Previously you had to import that function from `<app-name>/tests/helpers/ember-test-selectors`,
+but since our `addon` folder is now removed from the build in production
+we were able to simplify that import to just this:
+
+```js
+import testSelector from 'ember-test-selectors';
+```
+
+----
```
marcoow: We shouldn't be using `<hr>`s in the content as that conflicts with the layout

marcoow (changes requested):

marcoow (approved):

blog/_posts/2017-01-13-ember-test-selectors.md
```diff
@@ -0,0 +1,313 @@
...
+[ember-test-selectors](https://github.com/simplabs/ember-test-selectors) and
+today we are proud to present you with our next milestone: `0.1.0`.
+
+While `0.1.0` does not sound like much has changed, the addon has actually
+gained a lot of new functionality and should be considered our release candidate
+for `1.0.0`.
+
+This blog post will highlight the major changes in this release, and will
+give you a short introduction into *how* we have implemented these new
+features.
```
pangratz, line 20: IMHO we should have a TL;DR here with a summary of what happened:

- binding of `data-test-*` attributes within components in templates and JS files
- stripping of `data-test-*` attributes within templates and JS files
- simplified import of test selector

marcoow: Sorry, had dismissed your review a second earlier. This sounds like a good idea though. @Turbo87: want to create a follow-up PR adding this?

PR #61, opened 2017-02-17
install intercom for workshop catalog

marcoow:
might be helpful if people have questions


Turbo87 (approved):
bildschirmfoto 2017-02-17 um 10 20 02

should we put the simplabs logo in there instead of the "M"?

marcoow:
@Turbo87: that should actually show my face and name - that's configured on intercom's side anyway though.

PR #63, opened 2017-02-17
remove legacy transition rules

marcoow:


gr0uch (changes requested):
I think there were actually a few more instances of `-webkit-transition` and `-moz-transition`, can all of them be removed?

marcoow:
These were the only ones actually: https://github.com/simplabs/simplabs.github.io/search?utf8=✓&q=-webkit-transition

There were other introduced in another PR but that has not been merged to `master` and I removed them from the PR again.

gr0uch (approved):

PR #65, opened 2017-02-23
Don't load via http

ganeshbsub:
Based on https://github.com/simplabs/elixir-phoenix-workshops/commit/798ef30bcbfb5e0e4b1c39a129a13a995e18fd29


ganeshbsub:
@marcoow Also, can we disable scroll on the map here as well?

marcoow (approved):

PR #70, opened 2017-03-09
Minor correction in Job description

ganeshbsub:


marcoow (approved):

pangratz:
@ganeshbsub this is not based of the latest `master` and hence the `or --> of` correction I made in #69 would be overwritten again...

PR #71, opened 2017-03-09
ignore vimeo url

marcoow:
This particular URL seems to cause problems during tests - maybe Vimeo doesn't allow requests that are not coming from real browsers or so…

### DIFF
--- a/script/cibuild
+++ b/script/cibuild
@@ -2,4 +2,4 @@
 set -e # halt script on error
 
 bundle exec jekyll build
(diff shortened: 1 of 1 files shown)


ganeshbsub:
@marcoow  If that's the case, we should probably ignore all vimeo URLs then?

marcoow:
I'd ignore this for now and see whether we continue to have problems in the future.

ganeshbsub:
Cool. But just in case, for future reference - [Regex](http://stackoverflow.com/questions/29678646/can%C2%B4t-get-htmlproof-to-accept-an-ignore-pattern)

ganeshbsub (approved):

PR #75, opened 2017-03-14
Wording fixes

marcoow:
This fixes and improves some of the wording on simplabs.com


about/index.html
```diff
@@ -19,19 +19,19 @@
     <div class="col-md-10 col-md-offset-1 col-sm-12">
       <div class="row content-section split-section separated">
         <div class="col-sm-6">
-          <h3>Bavarian Engineering</h3>
+          <h3>International Engineering</h3>
```
Turbo87, line 22: "International Engineering" sounds a little strange. I think I'd keep the previous title here.

about/index.html
```diff
@@ -41,7 +41,7 @@
           <h3>Community Engagement</h3>
         </div>
         <div class="col-sm-6">
-          <p>We <strong>sponsor conferences</strong> like <a href="http://emberconf.com">EmberConf</a> and  <a href="http://rubyday.it">RubyDay.it</a>, <strong>organize the <a href="http://www.meetup.com/Ember-js-Munich/">Munich Ember.js Meetup</a></strong> and help girls and women learn Rails at <a href="http://railsgirls.com">Rails Girls</a>.</p>
+          <p>Our team will regularly be present at <strong>international conferences</strong>. We give talks, engage in the community and help by sponsoring conferences like <a href="http://emberconf.com">EmberConf</a>, <a href="http://www.elixirconf.eu">ElixirConf.EU</a>, <a href="https://emberfest.eu">Emberfest</a>, <a href="https://www.enterjs.de">enterjs</a> and others. We also organize local User Groupe and help girls and women learn Rails at <a href="http://railsgirls.com">Rails Girls</a>.</p>
```
Turbo87: "Our team is regularly present ..."

Turbo87: "local user groups"

clients/index.html
```diff
@@ -8,7 +8,7 @@
     <div class="row">
       <div class="col-md-6 col-md-offset-3 col-sm-10 col-sm-offset-1">
         <h2>Clients</h2>
-        <p>We have a track record of successful projects for clients <strong>from small startup to big enterprise</strong>.</p>
+        <p>We have a track record of <strong>projects for international clients</strong> from small startups to big enterprises.</p>
```
Turbo87, line 11: I like the old wording better here

trainings/index.html
```diff
@@ -19,12 +19,12 @@
     <div class="col-md-10 col-md-offset-1 col-sm-12">
       <div class="row content-section separated">
         <div class="col-sm-12">
-          <p>We offer a variety of Workshops to help teams advance their skills in many frontend and backend topics. <strong>All of the Workshops can be customized to meet your team's specific needs</strong> - <a href="/contact/">contact us</a> to discuss how we can help you best.</p>
+          <p>We offer a <strong>variety of Developer Trainings</strong> to help teams advance their skills in many frontend and backend topics. Professional Developer trainings can <strong>dramatically improve developer's productivity</strong> (we're usually told between 8% and 20%) and thus provide a manifold Return on Investment. All of the <strong>Workshops can be customized to meet your team's specific needs</strong> - <a href="/contact/">contact us</a> to discuss how we can help you best.</p>
```
Turbo87: > we're usually told between 8% and 20%

can that actually be quantified like this? I'd probably leave that part out

work/index.html
```diff
@@ -0,0 +1,57 @@
...
+            </div>
+            <div class="col-sm-4 col-xs-6 tool">
+              <a href="http://rubyonrails.org" class="icon"><i class="rails-logo"></i></a>
+            </div>
+          </div>
+        </div>
+      </div>
+      <div class="row content-section separated">
+        <div class="col-sm-12">
+          <p><strong>Ember.js, Phoenix and Ruby on Rails</strong> are what we work with most of the time. The <strong>conventional approach</strong> that they all share make them a great fit for successful <strong>projects that stay healthy over a long time</strong> and are simple for our clients to take over once we left.</p>
```
Turbo87: "once we left" sounds strange, I'd probably leave that part out.

work/index.html
```diff
@@ -0,0 +1,57 @@
...
+          <p>We will however choose whatever technology is best suited to get a particular job done, which in some cases might not be one of the above three. We keep our eyes open and <strong>love to explore new technical opportunities</strong>.</p>
```
Turbo87: "love to explore" -> "love exploring"

work/index.html
```diff
@@ -0,0 +1,57 @@
...
+        </div>
+      </div>
+      <div class="row content-section">
+        <div class="col-sm-12">
+          <h3>Our Process</h3>
+          <p>We are <strong>enthusiastic about agile practices</strong> and a flexible development process. While we don't think planning is evil per se, <strong>we don't overplan and aim to keep the process lean</strong>. Projects start off with a kick-off workshop to get everybody on the same page and from that continue with one or two week long sprints. We <strong>require clients to actively participate</strong> in projects and will communicate openly and often - not only to give status updates but also to explain <strong>the how and why behind design decisions we make</strong>.</p>
```
Turbo87: "on the same page and from that continue with one or two week" -> "on the same page and continue with one or two week"

about/index.html
marcoow: Hm, we're actually more international than bavarian though - both in terms of team as well as in terms of clients.

clients/index.html
marcoow: I think it's important though to mention we work for international clients - people might have the feeling we're a local german company while that's not actually the case. If we leave _"successful"_ in, the text is actually too long and breaks into 3 lines which doesn't work with the current layout. Also we shouldn't have to say the projects were successful?

trainings/index.html
marcoow: That's taken from https://mike.works actually. I think it's important to explain why investments in trainings are almost always worth the money but there might be better ways to do that of course…

work/index.html
marcoow: I think this is an important point though - explain why clients don't depend on us forever which is usually one of the biggest fears when hiring an external consultancy. Maybe there's a better way to put it?

work/index.html
marcoow: What about _"get everybody on the same page and then iterate in one or two week long sprints"_?

about/index.html
Turbo87: sure. but you'd also label something "Made in Germany" even if 90% were produced in Asia :P

clients/index.html
Turbo87: yeah, maybe it's just because of seeing the diff that I missed the "successful"

_config.yml
```diff
@@ -19,14 +19,14 @@ excerpt_separator: <!--break-->
 navigation:
   - title: Services
     url: /services/
-  - title: Process
-    url: /process/
+  - title: Work
+    url: /work/
+  - title: Trainings
```
oliverbarnes: Training, singular

about/index.html
oliverbarnes: I see Tobi's point, but I think stating the company is based in Munich gets the point across

services/index.html
```diff
@@ -32,16 +33,17 @@
           <h3>Technology Consulting</h3>
         </div>
         <div class="col-sm-6">
-          <p>We offer <strong>consulting services to help clients successfully implement their own projects</strong>. Having years of experience with Ruby on Rails, Ember.js, REST interfaces and scaling of large systems, we can <strong>help find performance bottlenecks</strong> in existing applications, review architectures or <strong>design complex client server systems</strong>.</p>
+          <p>We offer consulting services to <strong>help clients to successfully implement their own projects</strong> based on decisions and designs that we help make and validate.</p>
```
oliverbarnes: `help clients successfully` is better, I think - the `to` isn't needed

work/index.html
oliverbarnes: "once we leave" or "once we have left"

oliverbarnes (changes requested):

marcoow:
@oliverbarnes: addressed all your remarks, can you have another look?

oliverbarnes (approved):

oliverbarnes:
Looks great 👍

PR #78, opened 2017-03-16
improve some wording

marcoow:
some more improvements

### DIFF
--- a/_includes/blog_header.html
+++ b/_includes/blog_header.html
@@ -3,7 +3,7 @@
     <div class="row">
       <div class="col-sm-6 col-sm-offset-3">
         <h2>Blog</h2>
-        <p>We write about <strong>Ember.js</strong>, <strong>Ruby on Rails</strong> as well as <strong>Elixir</strong> and <strong>Phoenix</strong>.</p>
+        <p>We write about Web Engineering, <strong>Ember.js</strong>, <strong>Ruby on Rails</strong> as well as <strong>Elixir</strong> and <strong>Phoenix</strong>.</p>
     	</div>
     </div>
   </div>
--- a/_includes/blog_header.html
+++ b/_includes/blog_header.html
@@ -3,7 +3,7 @@
     <div class="row">
       <div class="col-sm-6 col-sm-offset-3">
         <h2>Blog</h2>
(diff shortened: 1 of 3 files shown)


_includes/blog_header.html
```diff
@@ -3,7 +3,7 @@
     <div class="row">
       <div class="col-sm-6 col-sm-offset-3">
         <h2>Blog</h2>
-        <p>We write about <strong>Ember.js</strong>, <strong>Ruby on Rails</strong> as well as <strong>Elixir</strong> and <strong>Phoenix</strong>.</p>
+        <p>We write about Web Engineering, <strong>Ember.js</strong>, <strong>Ruby on Rails</strong> as well as <strong>Elixir</strong> and <strong>Phoenix</strong>.</p>
```
oliverbarnes: I think the same thing I've been noting about capitalized common nouns applies here. But it'd be good to see other examples of this, do you have any references?

services/index.html
```diff
@@ -22,7 +22,7 @@
           <h3>Software Engineering</h3>
         </div>
         <div class="col-sm-6">
-          <p>simplabs builds solutions for the web and mobile devices for everyone from small startups to big enterprises. Our solutions are <strong>well architected and easy to maintain and extend</strong> by our clients once we're gone.</p>
+          <p>simplabs builds solutions for the web and mobile devices for everyone from small startups to big enterprises. Our solutions are <strong>well architected and easy to maintain and extend</strong> by our clients once we have left.</p>
```
oliverbarnes, line 25: if we're worried about "gone" and "left", we can also say "once our engagement ends" or something like it

oliverbarnes (changes requested):

_includes/blog_header.html
marcoow: Hm, no idea really. If you say lowercase I'll lowercase it.

services/index.html
marcoow: I think I had changed _"we're gone"_ to _"we have left"_ at a few other places and simply forgot this one so this is just aligning this paragraph with the other text.

_includes/blog_header.html
oliverbarnes: lowercase it then :)

services/index.html
oliverbarnes: 👍

_includes/blog_header.html
marcoow: fixed

oliverbarnes (approved):

Issue #74, opened 2017-03-10
Order Workshops better


marcoow:
We should order the workshops so that the frontend and backend workshops show up first (order among those doesn't really matter) and everything else comes below that (e.g. Sass, SEO etc.). Currently the SEO workshop shows up first which isn't really ideal.

ganeshbsub:
@marcoow both of them are also categorised as "Front end Dev" based on the data from Mike North's API I think.

marcoow:
Hm, is there any other way to identify the workshops that we'd treat as more important than other ones?

ganeshbsub:
We'd have to manually introduce a new tag I suppose.  Don't see another way from the data that's currently available. Maybe @EmanuelaTrabaldoLena can think of something else?

marcoow:
Hm, that's not ideal. @mike-north: is there any way we can identify the ones that are more important for us (potentially - I'm not actually sure whether that is actually true but I assume it is)?

marcoow:
fixed by #81

PR #81, opened 2017-03-21
WIP: workshops sorted by weight, added to every post

trabaldini:
Every post has a weight now so it's possible to sort them as wanted

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -45,11 +45,12 @@ facebook:
 exclude: [vendor]
 
 defaults:
-  - scope:
-    path: ""
+  -
+    scope:
+      path: ""
     values:
-    image: 
-    twitter: /images/apple-144.png
-    facebook: /images/apple-144.png
-    height: 144
-    width: 144
+      image: 
+        twitter: /images/apple-144.png
+        facebook: /images/apple-144.png
+        height: 144
+        width: 144
--- a/_config.yml
+++ b/_config.yml
@@ -45,12 +45,11 @@ facebook:
 exclude: [vendor]
 
 defaults:
-  -
-    scope:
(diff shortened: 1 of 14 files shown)


marcoow (changes requested):
The weight approach is good but I think we should auto-assign a weight based on the workshop title in the update script (otherwise we have to manually add the weights again every time we update the workshops). I'd make it so that we assign a weight of 1 for all workshops that have _"ember"_ in the title, 2 for all that have _"elixir"_ or _"phoenix"_ in the title and 3 for all others (make sure that the match is case-insensitive).

_config.yml
```diff
@@ -45,12 +45,11 @@ facebook:
...
-      image: 
-        twitter: /images/apple-144.png
-        facebook: /images/apple-144.png
-        height: 144
-        width: 144
+    image: 
+    twitter: /images/apple-144.png
+    facebook: /images/apple-144.png
+    height: 144
+    width: 144
```
marcoow: These changes seem unrelated?

marcoow (commented):
I think it would be easier to do this in the [update script](https://github.com/simplabs/simplabs.github.io/blob/master/script/update-workshops#L29) so that we don't have to add logic to the templates:

```rb
weight = if title =~ /ember/i
  1
else if title =~ /elixir|erlang|phoenix/
  2
else
  3
end

course_yaml = {
  layout:      'workshop',
  title:       title,
  weight: weight,
  permalink:   "/training/#{slug}",
  category:    category,
  description: description,
  image:       "/#{image_name}"
}
```

trabaldini:
closes #74

marcoow (approved):

PR #82, opened 2017-03-25
New blog post: "On Computed Properties vs. Helpers"

marcoow:
@kellyselden: would be great to hear your thoughts as well before releasing this!

### DIFF
--- a/blog/_posts/2017-03-21-on-computed-properties-vs-helpers.md
+++ b/blog/_posts/2017-03-21-on-computed-properties-vs-helpers.md
@@ -280,11 +280,10 @@ The intention of this post is not to say template helpers are bad and computed
 properties should be used for everything. That would be a pretty ignorant
 statement to make. __There are original concerns of the template layer that are
 best implemented in helpers__ - typical examples being translating strings,
-formatting numbers and dates etc. We use template helpers almost exclusively
-for these things in our projects (besides things like the `action`, `hash` or
-`mut` helpers of course). This is usually only a small part of any application
-though - we will often only have a few helpers defined in pretty big and
-involved applications.
+formatting numbers or dates etc. We use template helpers almost exclusively for
+this kind of functionality which is usually only a small part part of any
+application though - we often only have a few helpers defined even in pretty
+big and involved applications.
 
 ##### Missing a property context
 
--- a/blog/_posts/2017-03-21-on-computed-properties-vs-helpers.md
+++ b/blog/_posts/2017-03-21-on-computed-properties-vs-helpers.md
@@ -7,20 +7,25 @@ github-handle: marcoow
 twitter-handle: marcoow
 ---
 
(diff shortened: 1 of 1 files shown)


kellyselden:
When viewing https://github.com/simplabs/simplabs.github.io/blob/helpers-vs-computed-properties/blog/_posts/2017-03-21-on-computed-properties-vs-helpers.md, the markdown code escaping gets out of whack throughout the post. Is that expected with Jekyll?

marcoow:
Yep, all renders fine with Jekyll actually.

Turbo87 (approved):
LGTM 👍

blog/_posts/2017-03-21-on-computed-properties-vs-helpers.md
```diff
@@ -0,0 +1,355 @@
...
+```hbs {% raw %}
+{{userToDisplay.name}}
+{% endraw %}```
+
+Comparing this to a solution that moves all of the logic for selecting the user
+into the template via helpers illustrates how much harder it becomes to
+understand the template now:
+
+```hbs {% raw %}
+{{#each (take 1 (filter-by (shuffle users) 'state' 'active')) as |user|}}
```
kellyselden, line 130: I would do the filter-by first, then the shuffle to match the above order.

blog/_posts/2017-03-21-on-computed-properties-vs-helpers.md
```diff
@@ -0,0 +1,355 @@
...
+user's `age` property changes. While arguments that are passed to helpers are
+observed by Ember automatically and the __helper will be re-executed when any
+of these arguments change to a different value, the same is not true when any
+property on any of the arguments changes__. In this case changes of the user's
+`age` property would go unnoticed and the DOM would get out of sync with the
+underlying data.
+
+##### Performance
+
+In general, __performance should be more or less on par for computed properties
```
kellyselden, line 201: I've heard that this used to be true, until glimmer 2. Now templates are more performant. More research is needed to validate these claims. Perhaps a once-over on CP implementation can yield a similar performance boost.

blog/_posts/2017-03-21-on-computed-properties-vs-helpers.md
marcoow: Good point - will change

blog/_posts/2017-03-21-on-computed-properties-vs-helpers.md
marcoow: I'm not sure Glimmer 2 does actually make a difference here. The code that's being executed is pretty much the same both for the CP and the helper solution with the difference that with helpers all arguments will be eagerly evaluated so they might even be slower. I think the difference that Glimmer 2 makes is mostly how it renders the value that it gets from either the CP or the helper and how it makes sure updates are done in a performant way.

PR #90, opened 2017-04-13
navbar collapses to hamburger-menu

### LINKED ISSUE #86: Turn main navi into hamburger menu for narrow windows
We should not always display the main navi as a vertical list for narrow windows (e.g. mobiles). Instead, we should hide the menu and only display it on click on a hamburger icon or so (e.g. like http://www.201-created.com does it).

I think Bootstrap has built-in mechanisms for achieving that.

trabaldini:
closes #86

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -4,7 +4,7 @@
       <div class="col-sm-12 col-xs-12">
         <div class="navbar-header">
           <label for="menu-toggle" class="navbar-toggle collapsed">
-            <i class="fa fa-bars fa-2x"></i>
+            <i class="fa fa-bars"></i>
           </label>
         </div>
         <input type="checkbox" id="menu-toggle" class="hidden"/>
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -2,10 +2,12 @@
   <div class="container">
     <div class="row">
       <div class="col-sm-12 col-xs-12">
-        <button type="button" class="navbar-toggle" data-toggle="collapse" data-target="#collapsingNavbar">
-          ☰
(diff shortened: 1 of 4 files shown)


_includes/nav.html
```diff
@@ -2,23 +2,28 @@
...
-          </li>
-          {% for link in site.navigation %}
-            {% assign active = false %}
-            {% if page.url contains link.url %}
-              {% assign active = 'active' %}
-            {% endif %}
-            <li class="page">
-              <a href="{{link.url}}" class="{{active}}"><span>{{link.title}}</span></a>
+        <button type="button" class="navbar-toggle" data-toggle="collapse" data-target="#collapsingNavbar">
+          ☰
```
marcoow: I think we should use the respective character from font-awesome here instead.

marcoow (commented):
I think this should also be possible with CSS only which would be preferrable so we don't have to introduce additional JS dependencies.

See https://siongui.github.io/2017/03/23/css-only-bootstrap-responsive-navbar/

marcoow (commented):
Hm, this doesn't work for me - nothing happens when I click on the hamburger.

trabaldini:
Ok strange, on my localhost it's working. The simplabs-icon disappears and I can open the menu and the pages in it

marcoow:
Maybe you haven't committed some CSS changes? The changes in this PR seems pretty small. Also we should use `cursor: pointer;` for the hamburger menu.

marcoow:
Ah, sorry - didn't have the latest code actually. The hamburger menu seems a bit big though…

PR #103, opened 2017-05-22
Higher header

### LINKED ISSUE #102: Make header higher on mobile devices
Currently the header isn't high enough so that on most pages the content overflows the blue background. We should just make it higher for all pages.

img_0262

trabaldini:
closes #102

### DIFF
--- a/css/app.scss
+++ b/css/app.scss
@@ -591,12 +591,15 @@ ul {
   }
 
   .primary-banner {
-    padding: 20px 0;
-    height: 250px;
+    height: 200px;
+
+    h2 {
+      font-size: 36px;
+    }
 
     &.main-page {
       h2 {
-        font-size: 46px;
+        font-size: 36px;
       }
     }
   }
--- a/css/app.scss
+++ b/css/app.scss
@@ -185,7 +185,7 @@ ul {
 
 .primary-banner {
   padding: 20px 0;
-  height: 220px;
+  height: 250px;
 
   &.main-page {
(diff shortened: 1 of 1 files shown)


css/app.scss
```diff
@@ -185,7 +185,7 @@ ul {
 
 .primary-banner {
   padding: 20px 0;
-  height: 180px;
+  height: 250px;
```
marcoow: This makes the banner higher for all devices while the height should only be increased on mobile devices. See below - there's a section for devices with a specific max width which targets mobile devices with small screens.

marcoow (changes requested):
We need to limit this change to devices with small screens.

css/app.scss
```diff
@@ -589,4 +589,15 @@ ul {
...
+  .primary-banner {
+    padding: 20px 0;
+    height: 250px;
+
+    &.main-page {
+      h2 {
+        font-size: 46px;
+      }
+    }
+  }
```
marcoow: Isn't only changing the height sufficient? Also I think 200px should be enough or do we need the additional space at the bottom?

trabaldini: For the work section we need at least 240px for example on an iPhone 5. 
The size fits now for an iPhone, what kind of other changes are there that i could use?

marcoow: I think we can simply reduce the font size for `.primary-banner h2` to 36px on small screens.

marcoow (approved):

PR #105, opened 2017-06-08
updated script for 2 new workshops

trabaldini:


marcoow (changes requested):
We shouldn't add the TypeScript one yet as it's more or less empty /cc @mike-north

marcoow (approved):

PR #120, opened 2017-08-16
magic data post

geekygrappler:
Rough draft of a blog post on using data in tests.

I don't like the title.

It's quite chatty, which is my style, but probably not appropriate for simplabs so happy to make it more 'professional'.

I kinda want to remove the ember stuff from it, as it's not really relevant to the principle. Might reach a wider audience with vanilla JS.


marcoow (commented):
This is pretty nice! Also a great example of something coming up while working on a project turned into a blog post.

We should indeed make this sound a bit more _"professional"_ though ;) I'd also make the example much simpler. I guess you can illustrate the main point with very little code potentially, e.g. staying with the country picker component but simply using an array of country codes as the fixture data (`['de', 'fr', ''it]`) that you'll want to arrange in the test instead of depending on the fixture data.

blog/_posts/magic-test-data.md
```diff
@@ -0,0 +1,258 @@
+# Magic Data in Tests
+
+The AAA principle for testing is Arrange Act Assert and it's Amazing. The TL;DR is:
```
jayjayjpg: The alliteration here is `Amazing` :thumbsup: Do you know if there is any source for the AAA principle as well? I know most tech blogs are not so source-driven, but I'm always curious to learn more about where new terms come from and think this could work here as well. What do you think?

blog/_posts/magic-test-data.md
```diff
@@ -0,0 +1,258 @@
...
+If you have a test that doesn't Arrange, your test is brittle and hard to understand.
+
+Don't believe me? Lets go through a real life example. If you are tasked with making a country/language picker for a website that looks something like this.
+
+[screen shot of a country picker]
+
+You will probably have a hardcoded list of countries that your website supports somewhere in your app.
+
+```json
+[
```
jayjayjpg: Afaik the tricky part here is that the `.json` file cannot be imported later on as `import COUNTRIES from 'countries';` if it is in plain `JSON` format, but writing it as an ES6 module (`.js`) works out of the box:

```
// countries.js

export default [
{ locale: "en",
 country: "GB",
 // .....
}
];
```

blog/_posts/magic-test-data.md
```diff
@@ -0,0 +1,258 @@
...
+What you'll need to do is write a function that munges these two lists together and probably adds a key to each country that points to the flag image.
+
+So you'll write a component
+
+```js
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+
+/* the countries!! */
+import COUNTRIES from 'ct/countries';
```
jayjayjpg: This can be imported, once the `countries` file follows the ES6 module syntax

blog/_posts/magic-test-data.md
```diff
@@ -0,0 +1,258 @@
...
+
+This test will pass.
+
+// Celebration gif
+
+I have loads of problems with it.
+
+// Fry eyes gif
+
+1) The test is brittle.
```
jayjayjpg: That's a pretty neat structure to go through the pitfalls of the first test implementation step-by-step :thumbsup:

blog/_posts/magic-test-data.md
```diff
@@ -0,0 +1,258 @@
...
+    "locale": "de",
+    "region": "europe",
+    "name": "Germany"
+  }
+  /** other countries **/
+]
+```
+The test will now fail without us having changed the code. No code change, no behaviour change, but failing tests. FML. Sure it may only take you 5 mins to figure out why, but it would probably take me 15 minutes and that grinds my gears.
+
+2) The test doesn't actually tell me what the function is doing. By reading the test, I am none the wiser as to what the function does. The implication is there, but it's not explicit. The difference between implying and explicity defining is subtle but makes a whole world of difference. Let's say I didn't write that component, but I did add Bulgaria and now I've got to fix this breaking test. From reading the test what can I deduce? I can see that the function provides some regions and each country in the region will have a flag key, and that for some reason the first regions first country will be Deutschland...
```
jayjayjpg: I believe it would be cool to also state in a short sentence or example how the explicit test description differs from the implicit description that we have now. What do you think?

jayjayjpg (commented):
This post sounds fresh and I like the structure from going from the brittle test example over to the improved one step-by-step a lot! I left a few comments, let me know what you think

blog/_posts/magic-test-data.md
marcoow: @jessica-jordan: http://wiki.c2.com/?ArrangeActAssert

jayjayjpg: @marcoow Is this the original source for the principle (if there is any)?

marcoow: I don't know but http://wiki.c2.com generally has lots of good information.

blog/_posts/magic-test-data.md
geekygrappler: 👍 sounds good. I'll add it.

blog/_posts/magic-test-data.md
geekygrappler: Ah yeah good catch.

blog/_posts/magic-test-data.md
geekygrappler: http://xp123.com/articles/3a-arrange-act-assert/ he's staked a claim to it. But I did look at wiki.c2 when writing this (as the example is copy & pasted from there)

jayjayjpg: Wow, that's cool - I believe either a reference to that or to the `Test-Driven Development By Example` book that they mention would be really neat for the introduction into this blog post's topic 🤓

blog/_posts/magic-test-data.md
```diff
@@ -0,0 +1,242 @@
...
+  let groupCountriesByRegions = this.subject().get('groupCountriesByRegions');
+
+  /* This is assert */
+  assert.deepEqual(groupCountriesByRegions[0].countries[0], { name: 'Deutschland' locale: 'de', country: 'DE', region: 'europe' });
+});
+```
+
+This test will pass.
+
+// Celebration gif
```
geekygrappler: Are we good to keep these two gifs and the game of thrones one? I think gifs are fun and not too unprofessional.

blog/_posts/magic-test-data.md
```diff
@@ -0,0 +1,242 @@
...
+If you have a test that doesn't Arrange, your test may be brittle and hard to understand.
+
+Don't believe me? Lets go through a real life example. If you are tasked with making a country/language picker for a website that looks something like this.
+
+[screen shot of a country picker]
+
+You will probably have a hardcoded list of countries that your website supports somewhere in your app.
+
+```js
+export default [
```
marcoow: I'd add the (potential) name of this file as a comment at the top - that helps people understand the example better, e.g.

```js
// config/supported-locales.js
```

blog/_posts/magic-test-data.md
```diff
@@ -0,0 +1,242 @@
...
+];
+```
+
+And you will have a hard-coded list of regions that you support
+
+```js
+export default ['europe', 'asia-pacific'];
+```
+
+What you'll need to do is write a function that munges these two lists together.
```
marcoow: I'd make the example even simpler - assume you have a list of supported locales

```js
export default [
  'en',
  'fr',
  'de'
]
```

and a test that asserts the component renders each of these locales:

```js
assert.ok(find('button.en'));
assert.ok(find('button.de'));
assert.ok(find('button.fr'));
```

blog/_posts/magic-test-data.md
marcoow: Let's add some gifs first and see how it looks on the website.

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,243 @@
...
+  /* Assert */
+  assertEquals("cba", result);
+}
+```
+
+If you have a test that doesn't Arrange, your test may be brittle and hard to understand.
+
+Don't believe me? Lets go through a real life example. If you are tasked with making a country/language picker for a website that looks something like this.
+
+<!--break-->
```
marcoow: I'd move this to line 16

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,243 @@
+---
+layout: article
+section: Blog
+title: Magic Data in Tests
+author: "Marco Otte-Witte"
+github-handle: marcoow
+twitter-handle: marcoow
```
marcoow: This should be you @geekygrappler, not me

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,243 @@
...
+---
+
+
+The [AAA principle](http://wiki.c2.com/?ArrangeActAssert) for testing is Arrange Act Assert and it's Amazing. The TL;DR is:
```
marcoow: I'd add an introductory sentence maybe.

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,243 @@
...
+
+- Setup the data/inputs for the code you're testing (Arrange)
+- Make the code do something (Act)
```
marcoow: That sounds like you're changing some pre-existing code to me while actually what you mean is do/invoke something.

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,243 @@
...
+  /* Arrange */
+  String input = "abc";
+
+  /* Act */
+  String result = Util.reverse(input);
+
+  /* Assert */
+  assertEquals("cba", result);
+}
+```
```
marcoow: Can we use JS here? Java looks a bit strange in the context of the rest of the post.

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,243 @@
...
+<!--break-->
+
+[screen shot of a country picker]
+
+You will probably have a hardcoded list of countries that your website supports somewhere in your app.
+
+```js
+// config/countries.js
+export default [
+    {
```
marcoow: wrong indentation 👀

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,243 @@
...
+```
+
+And you will have a hard-coded list of locales that you support:
+
+```js
+// config/supported-locales.js
+export default ['en', 'de', 'zh'];
+```
+
+What you'll need to do is write a function that munges these two lists together.
```
marcoow: I still find this example too complex. The second list alone would be enough to get the point across. You have the list of supported locales and assert the component renders a selection button for each of them. If you depend on the list of locales in `config/supported-locales.js` your test is brittle as it depends on that global that is define outside of the test.

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,243 @@
...
+
+{: .center}
+![](/images/blog/celebrate.gif)
+
+I don't think it's a good test though.
+
+{: .center}
+![](/images/blog/fry.gif)
+
+1) The test is brittle.
```
marcoow: I'd make these sub-headlines.

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,243 @@
...
+```
+
+```js
+// config/supported-locales.js
+export default ['bg', 'de', 'en', 'zh'];
+```
+
+The test will now fail without us having changed the code. No code change, no behaviour change, but failing tests. The definition of a brittle test. It may only take the original writer of the test minutes to figure out why, but it might take any one new to the code unit a bit longer to figure out why.
+
+2) The test doesn't actually tell me what the function is doing. By reading the test, I am none the wiser as to what the function does. Of course this is a trivial example, but imagine if the function was merging three lists and adding specific fields based on conditionals? Then explicit tests are important. If I didn't write that component, but I did add Bulgaria and now I've broken the test and need to fix it the test isn't helpful. From reading the test what can I deduce? I can see that the function provides a lists of countries and the first one is supposed to be Deutschland... but why is it breaking?!?
```
marcoow: I'm not sure this relates to AAA really tbh. I think the main point of the post is you don't want to depend on anything global in a test.

css/app.scss
```diff
@@ -604,3 +604,7 @@ ul {
     }
   }
 }
+
+.center {
```
marcoow: 😱

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,243 @@
...
+
+I feel much better about this test. It deals with all three of my concerns with the previous test:
+
+1) The test is no longer brittle as it's not dependent on an external `JSON` file. 
+2) Reading from top to bottom I can now see, we have two lists on our component, `countries` and `locales`. We get the `supportedCountries` property from our component. That will be a list of the countries that are also supported locales. It's so transparent it could be the tax authority in [Norway](http://www.bbc.co.uk/news/magazine-40669239). It's much easier for developers to understand, much easier to debug the test as the test data is in the test, not hidden in another file.
+3) No more randomly selecting elements of lists to see if they are the same.
+
+And I get to commit GoT & LOTR to the code base.
+
+{: .center}
```
marcoow: All images should be centered already. Did that not work here?

marcoow (commented):
This is going to be a nice post!

I still find the example to complex though - you can do with something way simpler I think (see comment below). Also I'd focus more on the main point which is not depending on anything global and leave out the other ones.

Also while I think writing in an informal style is fine, the beginning doesn't read too nice - it feels a bit like non-connected sentences (in the first part, later that changes). I'd maybe add an introductory sentence and work a bit on the flow.

blog/_posts/2017-08-24-magic-test-data.md
geekygrappler: naww yeah but that's the point about making sure you arrange data. Both these lists are the same, they both come from some config file, but in the test both should be arranged to look like some short, easy to follow example. 

I take your point we could just have one list, and not filter one list based on another. Except I want to avoid ember specific code as much as possible, and if I just had one list, I'd test that the component rendered an input for `DE`, then I'm using `hbs{{country-picker` etc. and then a non-ember person will be thinking, what's this hbs thing? 

By having two list I think it simplifies it overall because I'm testing a plain JS function that filters a list based on another list. That function is more complex than simply rendering a list in an ember component, but doesn't require much ember specific knowledge to follow.

marcoow: Avoiding Ember-specific code is a good point. The problem is only that with this example you have to understand why/how these lists are munged together. I think you can probably find an example that uses a unit test (as opposed to an integration test that asserts on rendered DOM) but does something simpler than what you have here.

After all, the point is only that you don't rely on the list(s) from the config file in the test but define the list that is used in the test in the test file.

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -44,77 +44,46 @@ You will probably have a hardcoded list of countries that your website supports
...
-  },
-  {
-    'country': 'JP',
-    'locale': 'jp',
-    'region': 'asia-pacific',
-    'name': 'Japan'
+    country: 'GB',
+    locale: 'en',
+    region: 'europe',
+    name: 'United Kingdom'
```
jayjayjpg: The shorter data example is a nice change indeed :thumbsup:

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -44,77 +44,46 @@ You will probably have a hardcoded list of countries that your website supports
...
   }
 ];
 ```
+We need to write a function that adds a url for the country flag so that the component can display that flag image.
```
jayjayjpg: I think reading the full sentence, it already becomes clear, but it might be worth to further specify which url type should be added, to make it clearer from the start what the url is for, e.g.

> We need to write a function that adds an image url for the country flag [...]

or 

> We need to write a function that adds a url pointing to an image of the country flag [...]

or something similar - what do you think?

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -44,77 +44,46 @@ You will probably have a hardcoded list of countries that your website supports
...
-
-  regions: LOCALES,
-  
-  supportedCountries: function() {
-    const { countries, locales } = this.getProperties('countries', 'locales');
-
-    return countries.filter(country => locales.includes(country.locale));
+export default Component.extend({ 
+  displayCountries: function() {
+    return COUNTRIES.map(country => Object.assign({}, country, { flag: `/assets/images/flags/${country.country}.png` }));
```
jayjayjpg: Not sure, but would it be 
```
flag:  `/assets/images/flags/${country.locale}.png`
```
similar to the previous version of the example?

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -134,93 +103,73 @@ Let's say our business development team have made inroads into Bulgaria and now
...
   },
-  /** other countries **/
 ];
 ```
 
-```js
-// config/supported-locales.js
-export default ['bg', 'de', 'en', 'zh'];
-```
+The test will now fail without us having changed the code. No code change, no behaviour change, but failing tests. The definition of a brittle test. The test relies on a magic data from an external file, namely `COUNTRIES`. It may only take the original writer of the test minutes to figure out why the test fails, but it might take any one new to the code unit a bit longer to figure out why.
```
jayjayjpg: Nice explanation on why  magic data makes maintaining tests so cumbersome 👍  
I think  in the third sentence, either the article `a` can be saved or the wording `magic data set` could be used here:

> The test relies on magic data from an external [...]

> The test relies on a magic data set from an external [...]

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -134,93 +103,73 @@ Let's say our business development team have made inroads into Bulgaria and now
...
 
-The test will now fail without us having changed the code. No code change, no behaviour change, but failing tests. The definition of a brittle test. The test relies on a magic data, namely `COUNTRIES` and `LOCALES`. It may only take the original writer of the test minutes to figure out why, but it might take any one new to the code unit a bit longer to figure out why.
+What you should do is *Arrange* the data. When you do this it becomes clear that the function is simply adding a key, and it won't break due to external data changes. All you care is that given a starting set of data, after applying your function you get the resultant set of data, as clearly defined in the test.
```
jayjayjpg: Is there an `about` missing with 
> All you care is that a given [...]

or is it equivalent with

> All you care about is that a given [...]

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -134,93 +103,73 @@ Let's say our business development team have made inroads into Bulgaria and now
...
 
-Also `supportedCountries[0]` is [code smell](https://martinfowler.com/bliki/CodeSmell.html).
+First we need to stop using a constant in our component.
```
jayjayjpg: Sounds dramatic, but I enjoy it :thumbsup:

jayjayjpg (commented):
That post looks really cool with the examples and the gifs already! I just left a few comments mostly only related to wordings 👍

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,185 @@
...
+export default Component.extend({
+  countries: COUNTRIES,
+  
+  displayCountries: function() {
+    let countries = this.get('countries'); // This is equivalent to `this.countries` but for Ember objects.
+    return countries.map(country => Object.assign({}, country, { flag: `/assets/images/flags/${country.country}.png` }));
+  })
+});
+```
+
```
marcoow: I'd add something here like _"Then we can override that property in our test so that the test actually **arranges**"_ or something like that.

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,185 @@
...
+layout: article
+section: Blog
+title: Magic Data in Tests
+author: "Andy Brown"
+github-handle: geekygrappler
+twitter-handle: geekygrappler
+---
+
+Often when working on large codebases, my changes break some existing tests. While I would prefer my coding to be perfect, it's highly unlikely that I'll ever acheive the state of coding zen, so it's nice to know I have a test suite to catch me when I fall. Given that the codebase is large and in the majority not written by me, I tend to be introduced to code via the test files. One important principle I've started to follow when writing and refactoring tests is AAA. 
+
```
marcoow: maybe move the `<!-- break -->` here?

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,185 @@
...
+    country: 'GB',
+    locale: 'en',
+    region: 'europe',
+    name: 'United Kingdom'
+  }
+];
+```
+We need to write a function that adds a url for the country flag so that the component can display that flag image.
+
+
```
marcoow: extra newline 👀

css/app.scss
```diff
@@ -349,6 +349,7 @@ ul {
       max-width: 100%;
       margin-right: auto;
       margin-left: auto;
+      display: block;
```
marcoow, line 352: `block` should be the default for images anyway?

css/app.scss
```diff
@@ -603,4 +604,4 @@ ul {
       }
     }
   }
-}
+}
```
marcoow: 👀

blog/_posts/2017-08-24-magic-test-data.md
```diff
@@ -0,0 +1,185 @@
...
+export default Component.extend({
+  countries: COUNTRIES,
+  
+  displayCountries: function() {
+    let countries = this.get('countries'); // This is equivalent to `this.countries` but for Ember objects.
+    return countries.map(country => Object.assign({}, country, { flag: `/assets/images/flags/${country.country}.png` }));
+  })
+});
+```
```
marcoow: This is example is much nicer I think - much simpler than the previous one and still gets the point across.

marcoow (commented):
Nice! Few points only:

* your image is missing (see https://github.com/simplabs/simplabs.github.io/tree/master/images/authors)
* we should organize the images for posts in separate directories - see e.g. https://github.com/simplabs/simplabs.github.io/tree/master/images/posts/2017-08-30-creating-web-components-with-glimmer
* can you run http://imageoptim.com over all images?
* the country picker screenshot looks a bit weird on white background (esp. with the line on the right), can we include some of the background (or just make it use a different background color than white)?

blog/_posts/2017-08-24-magic-test-data.md
geekygrappler: :+1:

blog/_posts/2017-08-24-magic-test-data.md
geekygrappler: 👍

blog/_posts/2017-08-24-magic-test-data.md
geekygrappler: 👍

css/app.scss
geekygrappler: `inline` is the default for some reason.

blog/_posts/2017-08-24-magic-test-data.md
geekygrappler: :+1:

css/app.scss
geekygrappler: :+1:

blog/_posts/2017-08-24-magic-test-data.md
geekygrappler: No I think the about is missing. :+1:

blog/_posts/2017-08-24-magic-test-data.md
geekygrappler: 👍

blog/_posts/2017-08-24-magic-test-data.md
geekygrappler: Could be either, I think in the TL code base we do have it as `locale` e.g. `en.svg` but in this theoretical codebase it's `EN.png`.

css/app.scss
marcoow: 👍

blog/_posts/2017-08-24-magic-test-data.md
geekygrappler: 👍

blog/_posts/2017-08-24-magic-test-data.md
jayjayjpg: Ok I see, then either way works indeed 👍

geekygrappler:
@marcoow should be ready to publish once the tests pass.

marcoow (approved):

PR #121, opened 2017-08-16
WIP: feat(blog-post): adds "Creating Web Components w/ Glimmer" post

jayjayjpg:
My write-up of parts of my [EmberCamp talk](https://speakerdeck.com/jessicajordanjs/the-modern-state-of-web-components-a-glimmer-of-hope)

I'm still working on the parts marked with `?????` , but feel free to already review the rest and also give comments on general structure or writing style.

You can find the repo for the street map component example on here: https://github.com/jessica-jordan/glimmer-map (still needs some ✨ as well)

Current:

creating web components with glimmer simplabs


blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+   ```
+(??? This part hasn’t even been implemented in any browser yet. Not really sure if this is actually a promising thing of the spec anymore and this might better be left out of the blog post after all ????)
+
+Finally, a custom element that has been registered via the `CustomElementRegistry` can simply be used anywhere in HTML templates as a tag:
+
+
+    <my-customelement></my-customelement>
+
+Being able to encapsulate functionality as components and to easily reuse them via HTML is so powerful for several reasons: ????
+First, it allows….
```
geekygrappler: Cross compatibility of components, e.g. using components written in React in your Vue project without having to understand React or write any further code to get it to work.

geekygrappler: It allows the community to all start building and testing 'standard' components. At first there will be 10s of different 'calendar-picker' Web Components, but eventually they'll be 3 then 2 then 1. We'll reach the "best" `calendar-picker` for the any project based on actual usage by developers, rather than mandated by developers writing standards. And this battle tested component can be adopted by browsers once we have consensus. I think this is the best part about web components.

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+      |-- ui
+         |-- components
+            |-- glimmer-map
+               |-- component.ts
+               |-- template.hbs
+   ```
+
+Leveraging the benefits of [TypeScript](https://www.typescriptlang.org/) we are even able to use types while developing our component, but we are not forced to do so. Since TypeScript is a mere super set of JavaScript, syntactically correct, plain JavaScript will evaluate as valid TypeScript as well.
+
+Also, if we have a closer look on our app setup, we will find out, how our Glimmer app can be bother booted up, but also be rendered as a custom element later on when reused.
```
geekygrappler: type bother -> both

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+Creating a Glimmer Element Class for further reusage
+
+In our `src/ui/components/glimmer-map/component.ts` file, we can setup our map for display by taking advantage of the component lifecycle hooks - similar to the ones we became familiar with in Ember components.
+
+First, let’s import all the modules we will need for creating the component. We can import the default export of the `@glimmer/component` module, as well as from the `leaflet` package to get started:
+
+```
+    // src/ui/components/glimmer-map/component.ts
+    import Component from '@glimmer/component';
+    import L from 'leaflet';
```
geekygrappler: I presume there's a step to install `leaflet`? Is that `yarn add`/`npm install`?

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+        attribution: 'Map data &copy; <a href="http://openstreetmap.org">OpenStreetMap</a> contributors, <a href="http://creativecommons.org/licenses/by-sa/2.0/">CC-BY-SA</a>, Imagery © <a href="http://mapbox.com">Mapbox</a>',
+        maxZoom: 18,
+        id: 'mapbox.streets',
+        accessToken: 'your.mapbox.access.token'
+    }).addTo(this.map);
+      }
+    }
+```    
+
+We also aim to make our map respond to user input. Specifically, we would like to be able to set the marker to different locations on the map using a property for the horizontal position (`x`) and the vertical position `y` easily.  To allow a re-render of the component on any changes to these properties, we can make use of the `@tracked` decorators:
```
geekygrappler: (`x`) and `y` different formats

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+    <head>
+        <!-- ... -->
+        <link rel="stylesheet" href="https://unpkg.com/leaflet@1.2.0/dist/leaflet.css"
+          integrity="sha512-M2wvCLH6DSRazYeZRIm1JnYyh22purTM+FDB5CsyxtQJYeKq83arPe5wgbNmcFXGqiSH2XR8dT/fJISVA1r/zQ=="
+          crossorigin=""/>
+          <script src="/assets/webcomponentsjs/webcomponents-lite.js"></script>
+    </head>
+    <body>
+        <glimmer-map></glimmer-map>
+        <script src="app.js"></script>
```
geekygrappler: find this strange, shouldn't it be `glimmer-map.js`?

So I can just set up a html page and drop in `<glimmer-map></glimmer-map>` with that script and it'll work? That's cool.

geekygrappler (commented):
Looks good! I like the map example. I'd be interested to see how complicated it would be to make it support google api or leaflet.js.

blog/_posts/creating-web-components-with-glimmer.md
jayjayjpg: I especially like your second thought about the community focusing on a small subset of really well-developed components which solve a specific use case nicely @geekygrappler

blog/_posts/creating-web-components-with-glimmer.md
jayjayjpg: The assets in the `dist/` folder after `ember build` are not name spaced specifically, so it'd be as simple as dropping in an `app.js` file. And yes, I find it pretty cool that it's that straight forward, too

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
+# Creating Web Components with Glimmer
+
+At [this year's EmberConf the Ember core team officially announced](https://youtu.be/TEuY4GqwrUE?t=58m43s) the release of [Glimmer](https://glimmerjs.com/) - a light-weight JavaScript library aimed to provide an useful toolset for creating fast and reusable UI components. Powered by the already battle-tested Ember-CLI, developers can now build their Glimmer apps in an easy and efficient manner as they already came to love building applications in Ember.js before.
```
marcoow: I'd say _"can build Glimmer apps"_ - the _"now"_ isn't necessarily really as this is not a new way of building Glimmer apps (as there was previously no way to build Glimmer apps at all ;) )

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+
+Additionally, Glimmer allows the creation of components according to the Custom Elements v1 specification, making it possible to build native web components which may be used in applications of ????? I dunno all kinds of technological backgrounds?????.
```
marcoow: > Additionally, Glimmer allows

_"In addition to building standalone Glimmer apps…"_ might be clearer.

marcoow: > making it possible to build native web components which may be used in applications of ????? I dunno all kinds of technological backgrounds?????.

maybe something like _"making it possible to build native web components that can be reused across all kinds of frontend stacks.."_

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+
+Before diving into how we can build a Glimmer powered web components, let’s have a quick recap on what custom elements in general:
```
marcoow: _"Before diving into how to build Glimmer powered web components, here is a quick recap of what Custom Elements (aka Web Components) are:"_

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+
+
+Creating native, reusable web components with the Custom Elements spec v1
```
marcoow: I think this might not be necessary at all?

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+
+Custom Elements describes the capability to create customized HTML elements which can used with the traditional angle-bracket notation: `<my-customelement>`.
```
marcoow: _"The Custom Elements spec describes capabilities for creating custom HTML elements that can be used alongside native elements."_?

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+The current version v1 of this specification includes the definition of a `CustomElementRegistry` interface that allows to register custom elements based on extensions from the HTML element base class:
```
marcoow: > custom elements based on extensions from the HTML element base class:

Maybe _"custom elements as extensions of the `HTMLElement` base class"_?

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+
+    ```
+    class CustomElementClass extends HTMLElement {
+      // ...
+    }
+
+    customElements.define('my-customelement', CustomElementClass);
+    ```
```
marcoow: This code snippet does not actually use the above mentioned `CustomElementRegistry` class. Is `customElements.define` somehow Glimmer specific?

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+
+???? And it even proposes the ability to inherit from other DOM interfaces, allowing for the customization of built-in elements: ????
+
+   ```
+    class AutoPlayVideoClass extends HTMLVideoElement {
+      // ...
+    }
+    customElements.define('my-video`, AutoPlayVideoClass, { is: 'video'});
+   ```
+(??? This part hasn’t even been implemented in any browser yet. Not really sure if this is actually a promising thing of the spec anymore and this might better be left out of the blog post after all ????)
```
marcoow: It might be better to leave this out as it is also not really necessary to understand the topic of this post.

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+
+Finally, a custom element that has been registered via the `CustomElementRegistry` can simply be used anywhere in HTML templates as a tag:
```
marcoow: _"…can be used anywhere in HTML like any other tag:"_

blog/_posts/creating-web-components-with-glimmer.md
marcoow: I think one key point is that components are standalone and self-contained. It actually doesn't matter what was used to build the web component as it includes all its dependencies (and is actually a black box).

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+[INSERT IMAGE OF COMPONENT HERE]
+
+You can also check out the project on [Github](???? link to the repo, currently: https://github.com/jessica-jordan/glimmer-map ???? ).
+
+To get started with an initial, running project setup, we will be using [Ember CLI](https://ember-cli.com/) for scaffolding and configuration of our Glimmer app. From  `ember-cli@2.14.beta` onwards, we can get started to create a Glimmer-backed web component by using the `ember new` generator, the [respective glimmer blueprint](https://github.com/glimmerjs/glimmer-blueprint) and the `--web-component` flag:
+
+```
+    ember new glimmer-map -b @glimmer/blueprint --web-component
+```
+generating the needed scaffolding for our first Glimmer app. As soon as the Glimmer app is booted up via a simple `ember serve` and we navigate to the typical `http://localhost:4200`, we will find that our first component project is not only served and rendered, but inspecting the Page Source we will also see that the `glimmer-map` is not represented as a custom element; instead a placeholder element `#app` as well as a script tag containing `app.js` are loaded only, generating the DOM finally on runtime.
```
marcoow: Here you say that although we passed `--web-component` to `ember new` we don't actually see a custom elements in the DOM but a placeholder element `#app`. In the following you're focusing on other aspects of Glimmer like TypeScript and the project structure though. It might be better to either mention the placeholder element later in the post and explain why that looks how it looks then or explain that right here.

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+              placeholder.remove();
+              assignAttributes(customElement, glimmerElement);
+            });
+          }
+        }
+      });
+
+      window.customElements.define(name, GlimmerElement);
+    }
+```
```
marcoow: This is a pretty long and complex code example. Is it required in its completeness to get the point across? Maybe it could be shortened to only the relevant parts?

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+    import L from 'leaflet';
+
+    export default class GlimmerMap extends Component {
+      didInsertElement() {
+        this.createMapInstance();
+        this.renderMap();
+      }
+
+      createMapInstance() {
+        const map = L.map(this.mapTarget).setView([41.08, 11.068], 12);;
```
marcoow: I think this is missing a definition for `mapTarget`?

blog/_posts/creating-web-components-with-glimmer.md
marcoow: I think it would be good to show what the component renders in its current state before going to the next topic (enabling interactivity).

blog/_posts/creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,268 @@
...
+    <body>
+        <glimmer-map></glimmer-map>
+        <script src="app.js"></script>
+    </body>
+    </html>
+```
+
+And finally, we can see our Glimmer-based street map being rendered just as seen below:
+
+[INSERT COMPONENT HERE]
```
marcoow: This is a great way of showing off the custom component!

marcoow (commented):
This is a great post! I left some comments on a bunch of smaller issues but in general this is awesome. 🎉

blog/_posts/creating-web-components-with-glimmer.md
jayjayjpg: I think it is good to recap on this one, so people have a context when reading about the `customElements.define(...)` API mentioned later on in the article when looking at the Glimmer app setup [here](https://github.com/simplabs/simplabs.github.io/pull/121#discussion_r133544572)

blog/_posts/creating-web-components-with-glimmer.md
jayjayjpg: The `customElements` represents the `CustomElementRegistry`: 
https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define
https://www.w3.org/TR/2016/WD-custom-elements-20161013/#custom-elements-autonomous-example

It might be good to highlight this even more in [the recap on Custom Elements](https://github.com/simplabs/simplabs.github.io/pull/121#discussion_r133544057), so readers don't miss it?

blog/_posts/creating-web-components-with-glimmer.md
marcoow: Yeah, covering that makes sense but this section headline seems a bit out of place.

blog/_posts/creating-web-components-with-glimmer.md
marcoow: > It might be good to highlight this even more in the recap on Custom Elements, so readers don't miss it?

Yeah, you're looking for `CustomElementRegistry` to appear somewhere but it never does which seems strange. A comment next to `customElements.define` would probably already solve that.

blog/_posts/creating-web-components-with-glimmer.md
jayjayjpg: > Custom Elements (aka Web Components) are:

Could this wording be a bit difficult, as `custom elements` is technically only a sub part of the bigger specification of `web components` which is used as an umbrella term for templates + Shadow DOM + HTML imports + Custom Elements? I believe it's fair to use the words `custom elements` and `web components` interchangeably, when we are talking about creating the example component in this post (as a web component can technically be built upon only one of the four technologies it consists of - in this case, custom elements), but I wouldn't agree that `Custom Elements` == `Web Components`

What about 
""Before diving into how to build Glimmer powered web components, here is a quick recap of what Custom Elements are:"

And in the beginning of the recap:

"Custom Elements describes a technology which is - alongside of HTML template, Shadow DOM and HTML import - an integral part of the [current Web Component specification](https://www.w3.org/wiki/WebComponents/)." ?

marcoow: > but I wouldn't agree that Custom Elements == Web Components

Yeah, that's right of course. I just though that's how people use the terms in reality but you're right - it's not great to encourage that wrong usage and we shouldn't suggest it's correct.

> What about _"Before diving into how to build Glimmer powered web components, here is a quick recap of what Custom Elements are:"_

> And in the beginning of the recap: _"Custom Elements describes a technology which is - alongside of HTML template, Shadow DOM and HTML import - an integral part of the current Web Component specification."_?

👍

blog/_posts/creating-web-components-with-glimmer.md
jayjayjpg: That's a good point for making it easier to follow along the example - I'll update that

blog/_posts/creating-web-components-with-glimmer.md
jayjayjpg: Hm, why exactly? I think it makes sense to separate the theoretical part about the `custom elements` from the practical part of the code example with a section headline. Or do you suggest that the wording can be improved?

marcoow: It just seemed a bit strange to me that the previous paragraph ended with _"here is a quick recap of what Custom Elements are:"_ followed by a section headline and the actual content after that. I think it's better to either introduce the next paragraph in the previous paragraph (_"here is a quick recap of what Custom Elements are:"_) *or* use a section headline.

jayjayjpg:
@geekygrappler and @marcoow Can you take another look when you have time? Merci beaucoup

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
+---
+layout: article
+section: Blog
+title: Creating Web Components with Glimmer
+author: "Jessica Jordan"
+github-handle: jessica-jordan
+twitter-handle: jjordan_dev
+---
+
+At [this year's EmberConf the Ember core team officially announced](https://youtu.be/TEuY4GqwrUE?t=58m43s) the release of [Glimmer](https://glimmerjs.com/) - a light-weight JavaScript library aimed to provide an useful toolset for creating fast and reusable UI components. Powered by the already battle-tested Ember-CLI, developers can build their Glimmer apps in an easy and efficient manner as they already came to love building applications in Ember.js before.
```
marcoow: _"an useful toolset"_

**a** useful toolset

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+
+<!--break-->
+
+In addition to building standalone Glimmer applications, the library allows the creation of components according to the Custom Elements v1 specification, making it possible to build native web components which can be reused across all kinds of front end stacks.
+
+#### What's in the Custom Elements v1 specification
+
+Custom Elements describes a technology which is - alongside of HTML template, Shadow DOM and HTML import - an integral part of the [current Web Component specification](https://www.w3.org/wiki/WebComponents/).
```
marcoow: _"Custom Elements describes"_

How about _"The Custom Elements spec…"_?

marcoow: _"alongside of HTML template, Shadow DOM and HTML import"_

_"alongside ~of~ HTML template**s**, Shadow DOM and HTML import**s**"_

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+The Custom Elements spec describes capabilities for creating custom HTML elements which can be used just like native HTML elements: `<my-customelement>`.
+The [current version of this specification](https://www.w3.org/TR/custom-elements/) defines  a `CustomElementRegistry` interface; in the browser environment the registry is represented by the global `customElements` object and using its `define` method, custom elements as extensions of the HTMLElement base class can be registered as follows:
```
marcoow: I'd make this a new paragraph as this is a new topic.

marcoow: _"The [current version of this specification](https://www.w3.org/TR/custom-elements/) defines  a `CustomElementRegistry` interface; in the browser environment the registry is represented by the global `customElements`"_

_"The [current version of this specification](https://www.w3.org/TR/custom-elements/) defines  a `CustomElementRegistry` interface which is available as the `customElements` global in the browser. Its `define` method can be used to register custom elements:"_

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+
+```js
+customElements.define(elementName, constructor, options);
+```
+
+A specific custom element, for example `my-customelement`, can then be registered like so:
```
marcoow: Custom elements extend the native `HTMLElement ` base class:

marcoow: I'd probably switch the order of these 2 things - first show how a custom element is defined, than show how it is registered. Otherwise you need to repeat the registration here although it has been introduced before already.

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+
+```html
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>My Custom Element Demo</title>
+  </head>
+  <body>
+    <my-customelement></my-customelement>
+    <script src="/path/to/script/registering/custom-element.js"></script>
```
marcoow: I think this would be easier to understand if you used the tag after the `script` tag. Otherwise people might wonder whether this actually works in this order.

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+</html>
+```
+
+Being able to encapsulate functionality this way and to be able to reuse them via HTML markup is powerful for several reasons:
+
+Frame work agnosticism
+Wider community effort to create a small set of widgets to solve common problems
+
+
+First, it allows the free usage of these components beyond the boundaries of a wide range of front-end tech stacks, including the diverse set of client-side JavaScript frameworks out there.
```
marcoow: > _"it allows the free usage of these components"_

_"it allows ~the free~ usage of these components"_

marcoow: > _"beyond the boundaries of a wide range of front-end tech stacks"_

_"beyond the boundaries of a single front-end tech stack"_

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+Imagine being able to develop a menu header once and to reuse it across all the different applications that are built in the scope of your project,
+no matter if these applications were built upon React or Ember. This also furthers even greater efforts to create well-maintained and highly flexible widgets, not only on a project's or company's scale, but also in terms of open-source as an even greater community of developers will be interested in creating and using that e.g. one well-built `date-picker` component. Imagine if the community behind the `ember-power-datepicker`, `react-datepicker`, `angular-datepicker` and others could funnel their work into creating one component that is of great benefit to all of these JavaScript communities at once.
+Second, the standalone and self-contained nature of web components creates an API, that is not only straight-forward to use, but is also easy to reason about; a well-structured web component
```
marcoow: I'd make this a new paragraph, e.g. add another newline above.

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+will bring in all the dependencies needed for its core functionality and will be configured - if needed at all - with string based HTML attributes alone - making the final mark up very descriptive. Advanced users will still be able to configure additional, more specific functionality, as e.g. event listeners, via JavaScript themselves on top of that.
+Third, due to the fact that the knowledge of the HTML markup language alone is needed, the usage of these components becomes very easy. Using the basic and likely most well-known language of the web alone, it empowers an even larger community of developers to build their websites and web apps with and upon web components.
```
marcoow: This should be a new paragraph as well.

marcoow: > _"Third, due to the fact that the knowledge of the HTML markup language alone is needed, the usage of these components becomes very easy."_

_"Third, using web components is very easy as only knowledge of the HTML markup language is required to do so."_

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+In the following we will create a simple street map based on [Leaflet.js](http://leafletjs.com/) which can be re-used as such a custom element in any other application or static website:
+
+![Screenshot of Final Glimmer Map Component Example - Open Street Map View Centered on Munich](/images/glimmer-map-screenshot-final.png)
+
+
+You can also check out the project on [Github](https://github.com/jessica-jordan/glimmer-map).
+
+##### Starting a new Glimmer Web Component Project
+
+To get started with an initial, running project setup, we will be using [Ember CLI](https://ember-cli.com/) for scaffolding and configuration of our Glimmer app. From  `ember-cli@2.14.beta` onwards, we can get started to create a Glimmer-backed web component by using the `ember new` generator, the [respective glimmer blueprint](https://github.com/glimmerjs/glimmer-blueprint) and the `--web-component` flag:
```
marcoow: > _"From  `ember-cli@2.14.beta` onwards…"_

Ember CLI 2.14 has been released already - mentioning the beta here makes it sound like you need a beta release. Maybe _"From  `ember-cli@2.14.0` onwards…"_?

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+
+```bash
+ember new glimmer-map -b @glimmer/blueprint --web-component
+```
+generating the needed scaffolding for our first Glimmer app. As soon as the Glimmer app is booted up via a simple `ember serve` and we navigate to the typical `http://localhost:4200`, we will find that our first component project is already rendered with a "Hello Glimmer!" headline:
```
marcoow: > _"rendered with a "Hello Glimmer!" headline:"_

_"rendered with a "Welcome to Glimmer!" headline:"_

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+Let's have a look at our project's file structure as well; following the new folder structure planned out in the [module unification RFC](https://github.com/emberjs/rfcs/pull/143) we can already find our `component.ts` and `template.hbs` files for creating our component co-located in the same directory below the `src` directory:
+
+
+```html
+-- src
+  |-- ui
+     |-- components
+        |-- glimmer-map
+           |-- component.ts
+           |-- template.hbs
```
marcoow: You can use the output of `tree app` here

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+```
+
+Leveraging the benefits of [TypeScript](https://www.typescriptlang.org/) we are even able to use types while developing our component, but we are not forced to do so. Since TypeScript is a mere super set of JavaScript, syntactically correct, plain JavaScript will evaluate as valid TypeScript as well.
```
marcoow: I'd leave this out - this could be a long article by itself.

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+
+Also, if we have a closer look on our app setup, we will find out, how our Glimmer app can be both booted up and also be rendered as a custom element later on when reused.
+Under the hood, Glimmer’s `renderComponent` API, is taking care of initial setup and rendering of the component into the aforementioned placeholder in the DOM and the `initializeCustomElements` API for registering the component as a native custom element:
+
+In `glimmer-web-component/src/initialize-custom-elements.ts`:
+
+```js
```
marcoow: This should be ` ```ts `

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+  |-- ui
+     |-- components
+        |-- glimmer-map
+           |-- component.ts
+           |-- template.hbs
+```
+
+Leveraging the benefits of [TypeScript](https://www.typescriptlang.org/) we are even able to use types while developing our component, but we are not forced to do so. Since TypeScript is a mere super set of JavaScript, syntactically correct, plain JavaScript will evaluate as valid TypeScript as well.
+
+Also, if we have a closer look on our app setup, we will find out, how our Glimmer app can be both booted up and also be rendered as a custom element later on when reused.
```
marcoow: > _"Also, if we have a closer look on our app setup, we will find out, how our Glimmer app can be both booted up and also be rendered as a custom element later on when reused."_

_"Having a closer look at the app setup in `glimmer-web-component/src/initialize-custom-elements.ts`, we can see how the Glimmer app is started up and rendered as a custom element."_

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+Under the hood, Glimmer’s `renderComponent` API, is taking care of initial setup and rendering of the component into the aforementioned placeholder in the DOM and the `initializeCustomElements` API for registering the component as a native custom element:
+
+In `glimmer-web-component/src/initialize-custom-elements.ts`:
```
marcoow: I'd remove this and just mentioned it in the paragraph above: _"Having a closer look at the app setup in glimmer-web-component/src/initialize-custom-elements.ts…"_

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+}
+```
+
+The final line makes use of the `CustomElementRegistry`'s `define` method to register the `GlimmerElement` class as a custom element, just as we expect it from the Custom Elements specification. Interesting to note here as well is the call to the `assignAttributes` method, ensuring that any top-level attributes defined on our custom element are later on also mapped to attributes on our rendered Glimmer component.
+
+So, after having that quick dive into how our component is spun up in Glimmer, let’s get started developing it:
+
+##### Developing a Web Component with Glimmer's API
+
+In our `src/ui/components/glimmer-map/component.ts` file, we can setup our map for display by taking advantage of the component's lifecycle hooks - similar to the ones we became familiar with in Ember components.
```
marcoow: I wouldn't mention the component lifecycle hooks here as the following code sample shows a very basic component blueprint only.

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+
+So far the blueprint for our component module has already been setup by Ember CLI's generator to import the default export of the `@glimmer/component`:
```
marcoow: Maybe just change this and the previous paragraph to

_"The Glimmer app's main component is defined in `src/ui/components/glimmer-map/component.ts`:"_

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+import Component from '@glimmer/component';
+
+export default class GlimmerMap extends Component {
+
+}
+```
+As we would like to create a map based on `Leaflet.js`, let's also get that dependency installed in our project:
+
+```bash
+yarn add leaflet
```
marcoow: I guess this should be a dev dependency (`yarn add --dev leaflet`)?

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+      plugins: [
+        resolve({ jsnext: true, module: true, main: true }),
+        commonjs()
+      ]
+    }
+  });
+
+  return app.toTree();
+};
+```
```
marcoow: I'd explain what this enables quickly, e.g.:

_"This lets us import the Leaflet API as `import L from 'leatlet';`"_

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+
+Now for creating and rendering our map, let’s use the `didInsertElement` hook for our first render, ensuring that the component is readily waiting in the DOM for further setup:
```
marcoow: _"Now for creating and rendering the map, we can use the `didInsertElement` hook in the component so that the map is rendered after the Glimmer component has first been inserted into the DOM."_

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+    attribution: 'Map data &copy; <a href="http://openstreetmap.org">OpenStreetMap</a> contributors, <a href="http://creativecommons.org/licenses/by-sa/2.0/">CC-BY-SA</a>, Imagery © <a href="http://mapbox.com">Mapbox</a>',
+    maxZoom: 18,
+    id: 'mapbox.streets',
+    accessToken: 'your.mapbox.access.token'
+}).addTo(this.map);
+  }
+}
+```
+
+Also, let's add the needed markup into our `template.hbs`
```
marcoow: > _"into our `template.hbs`"_

_"into the component's template:"_

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+
+```hbs {% raw %}
+// src/ui/components/glimmer-map/template.hbs
+<div class="glimmer-map">
+  <div id="map"></div>
+</div>
+```
```
marcoow: This is missing an `{% endraw %}`

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+
+With this setup, our map already renders for the map points set intially:
+
+![Screenshot of Open Street Map View of First Pass on the Glimmer Map Component](/images/glimmer-map-screenshot.png)
+
+We also aim to make our map respond to user input. Specifically, we would like to be able to set the marker to different locations on the map using a property for the horizontal position (`x`) and the vertical position (`y`) easily.  To allow a re-render of the component on any changes to these properties, we can make use of the `@tracked` decorators:
```
marcoow: I'd call the properties `lat` and `lon` as that's what they are really.

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,319 @@
...
+  }
+}
+```
+
+With this we are already done and can get started with distributing our component.
+
+
+#### Reusing Glimmer components as custom elements
+
+As mentioned in the beginning of this article, the current blueprint for Glimmer web components comes with a wrapper for being able to package and reuse our Glimmer components as custom elements. To create our assets needed for reusage, let’s build those as we are already used to when building Ember apps:
```
marcoow: > _"as we are already used to when building Ember apps"_

_"~as~ **like** we are already used to when building Ember apps"_

css/app.scss
```diff
@@ -95,6 +95,10 @@
       margin-bottom: 0;
     }
   }
+
+  img {
+    width: 100%;
+  }
```
marcoow: Why is this change necessary?

marcoow (commented):
This is going to be an awesome article!

I left some comments - mostly on styling issues and wording.

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
jayjayjpg: I honestly wasn't aware of that :thumbsup:

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
jayjayjpg: I actually don't know how this would work with other custom element implementations, but in the case of `Glimmer` the script actually should be placed after the custom element. Should we rather leave it this way so the Glimmer example at the end of the post is more comprehensible?

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
jayjayjpg: Yes, I agree after looking at the rendered view :thumbsup:

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
marcoow: Ok if this is required by Glimmer that's fine of course. I just wonder why that might be. Could you add a link to the docs saying that?

css/app.scss
jayjayjpg: The images in the blog post were not resizing to the width of the container yet

marcoow: I assume that's because they were too small though and we should better use larger ones (in addition to keeping this change)?

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
jayjayjpg: I can't find a doc reference for that, the only way I found out was to copy the setup from the `index.html` that you will get when running `ember build --prod` and it works, moving the script to the top for e.g. the demo in this blog post fails to load the component

marcoow: 👍

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,341 @@
...
+
+The [current version of this specification](https://www.w3.org/TR/custom-elements/) defines  a `CustomElementRegistry` interface global which is available as the `customElements` global in the browser. Custom elements are based on extensions of the native `HTMLElement` base class:
+
+```js
+class CustomElementClass extends HTMLElement {
+  // ...
+}
+```
+
+The `CustomElementRegistry`s `define` method can subsequently be used to register custom elements, e.g. a custom element named `my-customelement`, would be registered as follows:
```
marcoow: > The `CustomElementRegistry`s `define`

The `CustomElementRegistry`'s `define` 👀

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,341 @@
...
+    <title>My Custom Element Demo</title>
+  </head>
+  <body>
+    <my-customelement></my-customelement>
+    <script src="/path/to/script/registering/custom-element.js"></script>
+  </body>
+</html>
+```
+
+Being able to encapsulate functionality this way and to be able to reuse them via HTML markup is powerful for several reasons:
```
marcoow: > and to be able to reuse them via HTML

and to be able to reuse **it** via HTML

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,341 @@
...
+
+Frame work agnosticism
+Wider community effort to create a small set of widgets to solve common problems
```
marcoow: I don't think this list is needed as you have _"First, …"_, _"Second, …"_ below anyway. I'd just remove these 2 lines.

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,341 @@
...
+
+
+First, it allows the of these components beyond the boundaries of a single front-end tech stack, including the diverse set of client-side JavaScript frameworks out there.
```
marcoow: > First, it allows the of these components

First, it **enables usage** of these components

css/app.scss
marcoow: Ah sorry, I think the idea behind the layout would not actually be as wide as the text necessarily. I guess narrower images that are centered would be fine as well in this case?

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,341 @@
...
+│   ├── index.html
+│   └── styles
+│       └── app.scss
+└── utils
+    └── test-helpers
+        └── test-helper.ts
+```
+
+Having a closer look at the app setup in `glimmer-web-component/src/initialize-custom-elements.ts`, we can see how the Glimmer app is started up and rendered as a custom element.
+Under the hood, Glimmer’s `renderComponent` API, is taking care of initial setup and rendering of the component into the aforementioned placeholder in the DOM and the `initializeCustomElements` API for registering the component as a native custom element:
```
marcoow: > Under the hood, Glimmer’s `renderComponent` API, is taking

Under the hood, Glimmer’s `renderComponent` API ~,~ is taking

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,341 @@
...
+import L from 'leaflet';
+
+export default class GlimmerMap extends Component {
+  didInsertElement() {
+    this.createMapInstance();
+    this.renderMap();
+  }
+
+  createMapInstance()
+    const element = this.element.children[0];
```
marcoow: I'd use

```js
const element = this.element.querySelector('#map');
```

here to make clear you're selecting the `#map` element.

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
```diff
@@ -0,0 +1,341 @@
...
+<div class="glimmer-map">
+  <div id="map"></div>
+</div>
+```{% endraw %}
+
+With this setup, our map already renders for the map points set intially:
+
+![Screenshot of Open Street Map View of First Pass on the Glimmer Map Component](/images/glimmer-map-screenshot.png)
+
+We also aim to make our map respond to user input. Specifically, we would like to be able to set the marker to different locations on the map using a property for the longitude (`lon`) and the latitude (`lat`) easily.  To allow a re-render of the component on any changes to these properties, we can make use of the `@tracked` decorators:
```
marcoow: > To allow a re-render of the component on any changes to these properties, we can make use of the `@tracked` decorators:

_To make sure any changes to these property trigger a re-render of the component's DOM, we can make use of the `@tracked` decorator:_

(_"To allow a…"_ sounds like this allows a re-render but actually it forces a re-render which is what we want; also there's 2 spaces in front of the sentence 👀 )

marcoow (commented):
Awesome, this is really, really good 🙌 

Few points:

* we need to shrink/optimize the images a bit - some of them are 4MB 😱 In addition to using a smaller resolution we should always run all images that we add through https://imageoptim.com/mac
* I'dd some `<strong>` (I believe `**` in Markdown) here and there to highlight important points. Especially in longer paragraphs that helps with reading the text.
* I think we should put images for blog posts into dedicated folders to stay organized: `images/posts/2017-08-21-creating-web-components-with-glimmer/`

css/app.scss
jayjayjpg: I think it really looks best and most clean if the images align together with the text at both sides of the container. I also think with the code examples already breaking out of the container layout, it's nice to have the images align with the text correctly to not make the overall layout feel too chaotic. 
The images I saw that didn't fit the container were too large for the container layout, so to ensure that this looks good on all screen sizes, I'd assume the `width: 100%` is a solid fix for it. 
What do you think?

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
jayjayjpg: I actually like that this example highlights the fact that there is an `element` property that gives access to the main HTML element of the component. I'd rather go for dropping the `#map` id from the template example to avoid further confusion, but leave the selection via `this.element.children[0]` in here.
What do you think?

css/app.scss
marcoow: I'd like to stay with the original layout actually. If we force all images to be 100% we're blowing up smaller images as well which won't look good. I'm fine with us looking at this again but should be done in a separate PR.

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
jayjayjpg: Yes, that shouldn't be in there at all 🙈

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
marcoow: `querySelector` is an API of the DOM element that `this.element` returns. It still shows that you can access the DOM, just makes it a bit clearer which DOM element is being selected.

css/app.scss
jayjayjpg: Ah okay, so you're rather concerned that this might affect other pages as well? I could scope the css for my blog post, if you like, since I know that I won't have images in it that are smaller than the container and will blow up and loose resolution. I really like the full-width layout of the images and would like to keep it for this post. I can see though that this shouldn't apply to other posts as well and should rather be re-evaluated in another PR.

marcoow: We shouldn't add post-specific CSS please. I think changing this to `max-width: 100%` (in a different PR) would be fine and should have the effect you want.

marcoow: This should be scoped to the `.article` parent element though.

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
jayjayjpg: Yes, I can see your point and I'll update that :thumbsup:

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
jayjayjpg: I find the wording ok so far, as it highlights that a re-render simply is not possible without any `@tracked` properties earlier. Also with multiple spaces it's handled in markdown similarly to e.g. `latex` so it shouldn't be an issue with the actual text rendering:

screen shot 2017-08-28 at 10 49 26

marcoow: > I find the wording ok so far, as it highlights that a re-render simply is not possible without any @tracked properties earlier.

You could always re-render but it would not update the DOM without the `@tracked` decorator. The point is that `@tracked` is not a prerequisite really for allowing re-renders but it **is** Glimmer's one-way-binding mechanism.

> Also with multiple spaces it's handled in markdown similarly to e.g. latex so it shouldn't be an issue with the actual text rendering:

Sure, just a housekeeping thing

jayjayjpg:
@marcoow I updated the post, lmkwyt 🤓

blog/_posts/2017-08-21-creating-web-components-with-glimmer.md
jayjayjpg: > You could always re-render but it would not update the DOM without [...]

I guess the wording I used then is sloppy, as `re-render` not necessarily equals `property update` or `DOM update`, would something like 

_To allow any property changes to be reflected in the component's DOM, we can make use of the @tracked decorator: [...] [EXAMPLE] Anytime there is a change to one of the `tracked` properties' values, a re-render of the component with a newly updated DOM will follow. _ 

work better in your opinion?

marcoow: Yeah I think that works better. But also marking something as `@tracked` will cause any change to that property to **trigger** a re-render as well.

marcoow (approved):

marcoow (approved):
🎉 :shipit:

PR #117, opened 2017-08-04
Add success stories

trabaldini:
This adds case studies for DD-WRT, Timify and trainline (unclear whether we'll be allowed to publish trainline):

@EmanuelaTrabaldoLena: can you check please whether all the texts are up to date with the latest texts in basecamp? Can you also add `alt` tags to the images in the success stories (see output of failing build: https://travis-ci.org/simplabs/simplabs.github.io/builds/277406311?utm_source=github_status&utm_medium=notification)?

localhost_4000_cases_dd-wrt ipad pro
localhost_4000_cases_timify ipad pro
localhost_4000_cases_trainline ipad pro

PR #128, opened 2017-09-18
Add landing page for targeting startups

trabaldini:
depends on #117 

The build will be successful again once #117 is merged and this is rebased.

screenshot

### DIFF
--- a/css/app.scss
+++ b/css/app.scss
@@ -130,6 +130,18 @@ blockquote {
   }
 }
 
+blockquote {
+  border: none;
+  width: 83.33333333%;
+  margin: 0 auto;
+  padding: 0;
+
+  &:before {
+    display: block;
+    content: "\201d";
+  }
+}
+
 .highlight {
   border: none;
   padding: 20px;
--- a/css/typography.scss
+++ b/css/typography.scss
@@ -42,6 +42,21 @@ code {
   white-space: normal;
 }
 
+blockquote {
+  text-align: center;
+  font-style: italic;
+  font-size: 16px;
+
+  &:before {
+    font-size: 90px;
+    line-height: 30px;
+  }
+
+  cite {
+    font-weight: bold;
+  }
+}
+
 .highlight {
   word-wrap: normal;
 
--- a/disruptors/index.html
+++ b/disruptors/index.html
@@ -37,7 +37,7 @@ title:
         </div>
         <div class="col-sm-6 content-teaser">
           <h4>Timify</h4>
-          <p>Timify is an <strong>online appointment scheduling calendar</strong> for businesses that connects service providers with clients. We <strong>helped them to ship a stable, high quality app</strong> that paved the way for them at an early stage.</p>
+          <p>Timify is an <strong>online appointment scheduling calendar</strong> for businesses that connects service providers with clients. We <strong>helped them ship a stable, high quality app</strong> that paved the way for them at an early stage.</p>
           <a href="/cases/timify" class="custom-button">learn more</a>
         </div>
       </div>
--- a/disruptors/index.html
+++ b/disruptors/index.html
@@ -1,0 +1,48 @@
+---
+layout: default
+title:
+---
+<div class="primary-banner main-page">
+  <div class="container">
+    <div class="row">
+      <div class="col-md-10 col-md-offset-1 col-sm-12 col-sm-offset-0">
+        <h2>You do the disrupting<br>we do the engineering</h2>
+        <p>We build <strong>custom web and mobile solutions</strong> for our clients to rely on.</p>
+    	</div>
+    </div>
+  </div>
+</div>
+
+<div class="container index landing-page" id="main-section">
+  <div class="row">
+    <div class="col-md-10 col-md-offset-1 col-sm-12">
+      <div class="row content-section separated">
+        <div class="col-sm-12">
+          <p>Disrupting a market is hard enough - <strong>let an experienced partner handle the engineering for you so you can concentrate on your core business</strong>. Our team of technology experts builds ambitious solutions for clients across the world, from startups to long-established businesses.</p>
+        </div>
+      </div>
+      <div class="row content-section separated">
+        <div class="col-xs-12">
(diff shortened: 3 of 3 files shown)


Gemfile.lock
```diff
@@ -220,4 +235,4 @@ DEPENDENCIES
   jekyll-seo-tag
 
 BUNDLED WITH
-   1.15.1
+   1.16.0.pre.2
```
marcoow: These changes are all unrelated

_config.yml
```diff
@@ -3,7 +3,7 @@ name: simplabs.com
 sass:
   sass_dir: css
 
-gems:
+plugins:
```
marcoow: This is unrelated

disruptors/index.html
```diff
@@ -0,0 +1,54 @@
...
+            <div class="col-md-10 col-md-offset-1 col-xs-12">
+              <h1>&rdquo;</h1>
+              <h4>It has been an absolute pleasure to work with simplabs. Their team of experienced JavaScript engineers quickly slotted in to our in-house development teams. Their expertise with modern, agile software development best practices and tools meant they were able to work with our processes to deliver fantastic customer experiences</h4><br>
+              <h6><strong>Mark Holt, </strong>Trainline CTO</h6>
+            </div>
+          </div>
+        </div>
+      </div>
+      <div class="row content-section success">
+        <h3>Recent Case Studies</h3>
```
marcoow: This header is not needed

disruptors/index.html
```diff
@@ -0,0 +1,54 @@
...
+        <div class="col-sm-6 content-teaser">
+          <h6>ruby on rails, custom engineering, consulting</h6>
```
marcoow: We should not be using these tags

disruptors/index.html
```diff
@@ -0,0 +1,54 @@
+---
+layout: default
+title: Disruptors
```
marcoow: We should use the same title that we use for the home page here

marcoow (commented):
Nice. This should really look the same as the home page though, only with different text and the quote instead of the client logos (without the blue background).

marcoow (approved):

disruptors/index.html
```diff
@@ -0,0 +1,48 @@
...
+    </div>
+  </div>
+</div>
+
+<div class="container index landing-page" id="main-section">
+  <div class="row">
+    <div class="col-md-10 col-md-offset-1 col-sm-12">
+      <div class="row content-section separated">
+        <div class="col-sm-12">
+          <p>Disrupting a market is hard enough - <strong>let an experienced partner handle the Engineering for you so you can concentrate on your core business</strong>. Our team of technology experts builds ambitious solutions for clients across the world, from startups to long-established businesses.</p>
```
pangratz: Not a native speaker, but the capitalized `Engineering` caught my eye... cc @oliverbarnes @geekygrappler

oliverbarnes: **engineering** (lowercase) :)

disruptors/index.html
```diff
@@ -0,0 +1,48 @@
...
+      </div>
+      <div class="row content-section success">
+        <div class="col-sm-6 content-teaser">
+          <h4>DD-WRT</h4>
+          <p>DD-WRT is a Linux based <strong>firmware for wireless routers</strong>, running on millions of devices worldwide. We helped embeDD &mdash; the company behind DD-WRT &mdash; <strong>migrate the frontend to a modern tech stack</strong>.</p>
+          <a href="/cases/dd-wrt" class="custom-button">learn more</a>
+        </div>
+        <div class="col-sm-6 content-teaser">
+          <h4>Timify</h4>
+          <p>Timify is an <strong>online appointment scheduling calendar</strong> for businesses that connects service providers with clients. We <strong>helped them to ship a stable, high quality app</strong> that paved the way for them at an early stage.</p>
```
pangratz: IMHO `for businesses that connects service providers with clients` reads a bit halting (if that's the right word https://www.dict.cc/?s=holprig). My suggestion is something like:

> Timify is a calendar system which allows clients to schedule an appointment online.

marcoow: _" connecting sellers and clients"_ is almost stereotypic startup talk though…

oliverbarnes (changes requested):
Also, the contact footer would also be better as a regular sentence with only the first word capitalized: `Do you have a project in mind?` instead of `Do you Have a Project in Mind?`

disruptors/index.html
marcoow: fixed

marcoow:
> instead of Do you Have a Project in Mind?

Ouch, wasn't aware we were writing it like that 🤦‍♂️ Will fix in a separate PR.

oliverbarnes (approved):

disruptors/index.html
```diff
@@ -0,0 +1,48 @@
...
```
niklaslong: This is a detail, but I think `We helped them ship a stable, high quality app` reads better than `to ship`.

marcoow: agree

jayjayjpg (approved):
:sparkles: This looks great @EmanuelaTrabaldoLena

css/app.scss
```diff
@@ -130,6 +130,18 @@ blockquote {
...
+  width: 83.33333333%;
+  margin: 0 auto;
+  padding: 0;
+
+  &:before {
+    display: block;
+    content: "\201d";
+  }
+}
+
```
marcoow, line 144: This is duplicate now: https://github.com/simplabs/simplabs.github.io/blob/master/css/app.scss#L121

css/typography.scss
```diff
@@ -42,6 +42,21 @@ code {
...
+
+  &:before {
+    font-size: 90px;
+    line-height: 30px;
+  }
+
+  cite {
+    font-weight: bold;
+  }
+}
```
marcoow, line 58: This is now duplicate as well: https://github.com/simplabs/simplabs.github.io/blob/master/css/typography.scss#L26

PR #135, opened 2017-10-11
add contact-form

trabaldini:
Better text? Solution with jQuery ok?


marcoow (commented):
Works well - we should improve the styling a bit and render the success/error messages inline but this is nice.

@ninaarndt: do you think we need to get a design from Flo?

marcoow (approved):

PR #164, opened 2017-11-27
fix quote-styling

### LINKED ISSUE #160: Styling quirks with quotes
https://simplabs.com/training/ then `Ember Data Basic` (https://simplabs.com/training/2016-12-17-ember-data-basics.html):

screen shot 2017-11-22 at 17 51 02

trabaldini:
closes #160

### DIFF
--- a/training/_posts/2016-12-17-ember-data-basics.md
+++ b/training/_posts/2016-12-17-ember-data-basics.md
@@ -15,10 +15,10 @@ stages:
   description: "[JSON-API](http://jsonapi.org/) is a spec for JSON contracts between
     APIs and their clients. Because ember-data uses JSON-API by default, and as a
     guide for its internal data structures, it's useful to be familiar with it when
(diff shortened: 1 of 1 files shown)


marcoow (changes requested):
Let's fix the workshop markdown instead - after all those should not be quotes anyway. Also can you run the workshop content update script again?

trabaldini:
The problem is when I fix the markdown and then run the script the update is undoing my changes

marcoow:
Yeah, seems like the content is broken on @mike-north's side. You need to run the script and fix the content afterwards.

trabaldini:
This should fix the quotes.
I would also leave the changes in css because if we use quotes in markdown they won't be changed and it's easier to maintain.

marcoow:
Let's not add the CSS changes until we need them. Once a CSS is added it tends to never get removed anymore although it might not be necessary or even used.

marcoow (approved):

PR #172, opened 2017-12-07
Add `ember-freestyle` blog post

Turbo87:


blog/_posts/2017-12-07-ember-freestyle.md
```diff
@@ -0,0 +1,261 @@
+---
+layout: article
+section: Blog
+title: "Using ember-freestyle as a component playground"
+author: "Tobias Bieniek"
+github-handle: Turbo87
+twitter-handle: tobiasbieniek
+---
+
+A component playground is an application that you can use to test out and
```
marcoow: Is _"application"_ the best term here? It's actually not an application but part of an app both in Storybook and ember-freestaly. Maybe say _"…is a part of an application that you can…"_

blog/_posts/2017-12-07-ember-freestyle.md
```diff
@@ -0,0 +1,261 @@
...
+
+  deactivate() {
+    this._super(...arguments);
+    this.controllerFor('application').set('onFreestyleRoute', false);
+  },
+});
+```
+
+If you check `/freestyle` again you should see that now the only thing that is
+displayed is the `ember-freestyle` guide. 🎉
```
marcoow: Is this whole section actually so relevant? It introduces a rather obscure concept only to solve a problem that most real world apps don't have anyway - the `application` route rendering a huge component rather than just some UI chrome that would normally not be a problem to have on top and/or below the playground?

marcoow: I'd maybe just replace this with a sentence _"Remove the `ember-welcome-page` dependency"_.

blog/_posts/2017-12-07-ember-freestyle.md
```diff
@@ -0,0 +1,261 @@
...
+If you now restart the development server and check the `/freestyle` route
+you should see that everything is working as before, but if instead you run
+`ember serve -prod` you will notice that you only see a blank page because
+the `freestyle` controller, route and template are no longer available.
+
+This leaves us with only the single condition in the `application` template and
+the `this.route('freestyle')` call in the `router.js` file that we ship to
+production. While we could put in some effort to remove those too the situation
+is good enough and we can finally focus on putting content into our new
+component playground! 🎉
```
marcoow: I guess it would be good to mention that the route should be wrapped in `if (env !== 'production')` so we're not routing something that we can'd handle anyway (and it would be correctly handled by a potential catchall route or so).

marcoow (commented):
Nice post! I'm not sure about the paragraph on the application route though and would put in some `<strong>`s here and there - I generally find that these make especially longer paragraph easier to read, e.g.

bildschirmfoto 2017-12-07 um 16 06 28

blog/_posts/2017-12-07-ember-freestyle.md
Turbo87: well, it's an application within an application and I wanted to keep the language as simple as possible

blog/_posts/2017-12-07-ember-freestyle.md
Turbo87: have a look at webapp... this is very much a real-world-problem

blog/_posts/2017-12-07-ember-freestyle.md
Turbo87: that would result in more bytes being shipped to production though, so I'm not sure if that's really necessary.

blog/_posts/2017-12-07-ember-freestyle.md
marcoow: In the context of this post it's only a problem since `ember-welcome-page` is not removed - this adds a pretty high mental hurdle for everyone to step over that could be avoided by removing the `ember-welcome-page` dependency (and maybe talk about this briefly a bit further down).

blog/_posts/2017-12-07-ember-freestyle.md
marcoow: I think the number of bytes isn't relevant in this case and I don't think it's great to advise just leaving it in there although it's guaranteed not to work.

blog/_posts/2017-12-07-ember-freestyle.md
Turbo87: again, a lot of real-world-apps have wrapper UI in their `application` template. this section is teaching the readers how to work around that.

blog/_posts/2017-12-07-ember-freestyle.md
Turbo87: it does "work", it shows an empty page 😉

marcoow: > and it would be correctly handled by a potential catchall route or so

blog/_posts/2017-12-07-ember-freestyle.md
marcoow: Yes but I don't get why the wrapper UI is a) a problem and b) that problem is worth a lengthy paragraph at the beginning of the post. The problem is this gives people an impression of _"Oh wow, I have to do all these things and even use `controllerFor` that I heard to avoid to use ember-freestyle"_.

blog/_posts/2017-12-07-ember-freestyle.md
```diff
@@ -0,0 +1,261 @@
...
+
+<!--break-->
+
+
+#### Component Playgrounds
+
+You might be wondering "Why would would I need something like this? I can just
+test my components in the app!" and you're right. But imagine building a
+reasonably large application with dozens of routes, hundreds of components
+and thousands of possible component-states. Checking all of these
```
jayjayjpg: I think this can be `[...] thousands of possible component states. [...]` , as hyphens are not very common as part of composed words as in German. Regarding this term specifically see also e.g. [here](https://reactjs.org/docs/faq-state.html) or [here](https://docs.oracle.com/cd/B40099_02/books/SysDiag/SysDiagSysMonitor20.html)

blog/_posts/2017-12-07-ember-freestyle.md
```diff
@@ -0,0 +1,261 @@
...
+by Dominic Nguyen that explains the benefits very well.
+
+[ui-component-explorers]: https://blog.hichroma.com/the-crucial-tool-for-modern-frontend-engineers-fb849b06187a
+
+
+#### Installing `ember-freestyle`
+
+Installing and configuring `ember-freestyle` correctly is unfortunately a
+little complicated right now so if you want to take a look at the final
+outcome you can skip forward to the ["Using `ember-freestyle`"](#using-ember-freestyle)
```
jayjayjpg: I think it would be `take a look at the final outcome, you can [...]` with the comma separating the conditional and the main sentence

blog/_posts/2017-12-07-ember-freestyle.md
```diff
@@ -0,0 +1,261 @@
...
+section for now.
+
+Before we start installing anything we should make sure we know what situation
```
jayjayjpg: That's a great intro into the "How to" part 👍

blog/_posts/2017-12-07-ember-freestyle.md
```diff
@@ -0,0 +1,261 @@
...
+
+![Screenshot](/images/posts/2017-12-07-ember-freestyle/freestyle-underneath-welcome-page.png)
+
+This is obviously not what we want. To fix this we will need to adjust the
+`app/templates/application.hbs` file and wrap the existing contents in a
+condition:
+
+```diff
+{% raw %}+{{#if onFreestyleRoute}}
++  {{outlet}}
```
jayjayjpg: These git modifiers look pretty neat. How do these look like in the end?

blog/_posts/2017-12-07-ember-freestyle.md
```diff
@@ -0,0 +1,261 @@
...
++{{else}}
+ {{!-- The following component displays Ember's default welcome message. --}}
+ {{welcome-page}}
+ {{!-- Feel free to remove this! --}}
+
+ {{outlet}}
++{{/if}}{% endraw %}
+```
+
+`onFreestyleRoute` is a propery on the `application` controller that will be
```
jayjayjpg: I think this would be 
```
[...] is a property on the [...]
```

blog/_posts/2017-12-07-ember-freestyle.md
```diff
@@ -0,0 +1,261 @@
...
+set to `true` once we visit the `/freestyle` route and and back to `false`
+once we leave it again. This can be implemented in the `app/routes/freestyle.js`
+file, and since that does not exist yet we can generate it using
```
jayjayjpg: I think this would be 
```
since that does not exist yet, we can [...]
```
to separate the main sentence from the rest

blog/_posts/2017-12-07-ember-freestyle.md
```diff
@@ -0,0 +1,261 @@
...
+});
+```
+
+If you check `/freestyle` again you should see that now the only thing that is
+displayed is the `ember-freestyle` guide. 🎉
+
+##### Removing `ember-freestyle` from the production build
+
+Once you try to ship this to production you might notice another issue: your
+asset size has increase significantly, so it seems that `ember-freestyle` is
```
jayjayjpg: I think this would be `asset size has increased significantly [...]`

blog/_posts/2017-12-07-ember-freestyle.md
```diff
@@ -0,0 +1,261 @@
...
+
+Next we will add our new `freestyle` in-repo-addon to the `pluginsToBlacklist`
+list above:
+
+```js
+let pluginsToBlacklist = environment === 'production' ? ['ember-freestyle', 'freestyle'] : [];
+```
+
+Finally we need to adjust the paths that `ember-freestyle` uses to generate
+usage code snippets. For that we will add a `snippetSearchPaths` property
```
jayjayjpg: I think it could be useful to elaborate more what the term `usage code snippets` describes with a short relative sentence or something, e.g. 

```
the paths that "ember-freestyle" uses to generate code snippets which show how the component is used in a host application. [...]
```
or something similar. What do you think?

jayjayjpg (commented):
That's a great blog post and I really like the flow of the full walkthrough on how to use the project. I mostly left only minor grammar / spelling comments for now 👍

blog/_posts/2017-12-07-ember-freestyle.md
Turbo87: `+` colors the line green, and `-` colors it red

Turbo87:
@jessica-jordan fixed, thanks :)

blog/_posts/2017-12-07-ember-freestyle.md
marcoow: seems we're stuck here - let's get some more opinions from @jessica-jordan, @pangratz, @geekygrappler

blog/_posts/2017-12-07-ember-freestyle.md
marcoow: seems we're stuck here - let's get some more opinions from @jessica-jordan, @pangratz, @geekygrappler

pangratz: I'm in favor of the stripping-in-production approach. No side effects in production and I guess the bytes are negligible.

blog/_posts/2017-12-07-ember-freestyle.md
pangratz: Hmm, so I can see both arguments. It feels like a rather large introduction "just" for clarifying how the styleguide is added to the app so it's available via `/freestyle` without any application wrapping view stuff.
But I can see how this is information is useful in apps which have a large `application.hbs`... 🤔 

How about modifying the introduction of the `` Installing `ember-freestyle` `` section so it emphasizes on the current issues with the official approach for `ember-freestyle` (wrapped within application layout, not stripped from production) and that this section is only relevant if you care about this?

Turbo87: The whole point of this blog post is to teach about stuff that the `ember-freestyle` docs don't teach you by default, and this is exactly one of those things and why I'm emphasizing it.

marcoow: Maybe that's the problem then - to me it read like a post about how `ember-freestyle` is introduced into a project in general (vs. focussing on things that are not mentioned in the addon's docs already) and then this whole block while of course it makes sense seems like a lot of upfront stuff that's not strictly necessary for using the addon.

jayjayjpg: I think having a full separate route for displaying the `freestyle` styleguide makes sense in an application and it can be useful to also have this as part of the blog post. 
In my opinion it might be nice to distinguish this approach from the minimal setup though, e.g. through re-wording the section title from describing "what" is done to "why" it's done or to "which outcome" it has, for example:

```
Separating the styleguide from your application through Routes
```

```
Convenient access to the styleguide through separate Routes
```

or something similar. Another short sentence in the first paragraph of this section describing why this additional setup is useful can make it also easier for readers to quickly grasp why this is done and decide if they want to go through this section or skip it instead.

chriskrycho: @Turbo87 linked me this doc b/c I was wondering about, well… exactly this idea. This post did *exactly* what I needed it to: I understood quite well how to integrate freestyle in general—as I suspect most people will be, as its own directions are pretty solid.

But figuring out how to get the style guide on its own route, with a totally different template context than everything else in the application, was *really* what I needed. Our app has a *boatload* of stuff in the application template (it's 65 lines long), and we want exactly *none* of it on our living style guide! Similarly, the info below about how to split it into its own in-repo add-on to keep it out of production was perfect, and again *exactly* what I was looking for.

@jessica-jordan's and @pangratz's most recent suggestions about headings and tweaks would be helpful in clarifying intent and direction—but I'm mostly here to say "This resource *as is* is already incredibly important and valuable; please publish it so I can link others to it!"

blog/_posts/2017-12-07-ember-freestyle.md
chriskrycho: FWIW, In my own app, even though it wasn't mentioned here, I did the `env !== 'production'` tweak independently—my own take is that it's important not to hit a blank page like that but to explicitly 404. I also took the bytes as negligible there.

Turbo87: @marcoow I've added a small section about adding a conditional to `router.js`

marcoow (approved):

Issue #161, opened 2017-11-22
You can hit send without entering info at the contact form


pangratz:
There is no validation on the contact form, so when nothing is entered and you hit `Send`, then `Sending ...` is shown which is acknowledged with a `Thanks, we'll be in touch`.

marcoow:
We could use https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Form_validation but I don't think this is really important

PR #173, opened 2017-12-13
Form validation

trabaldini:
I added a refresh page as soon as the form is successfully submitted, else the error messages wont work again. There is probably a better solution to this?

### DIFF
--- a/contact/index.html
+++ b/contact/index.html
@@ -119,6 +119,9 @@ title: Contact
                     complete: function() {
                       button.prop('disabled', false);
                       button.text('Send');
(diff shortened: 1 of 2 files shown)


marcoow (commented):
Hm, I don't think we need to do any of our own form validation really - should be much simpler to use HTML5's form data validation feature and only add a few `required` and `pattern` attributes as I had suggested here: https://github.com/simplabs/simplabs.github.io/issues/161#issuecomment-346412523

trabaldini:
closes #161

marcoow (approved):

PR #178, opened 2018-01-15
Cookie banner

### LINKED ISSUE #104: Add cookie warning
I think we're required to show a cookie warning on the website actually 😬

https://www.cookielaw.org/the-cookie-law/

There might be an external service we can use rather than implement it ourselves.

trabaldini:
works, but when page is loaded you see it for a short period of time before it's dismissed. Have to improve that

closes #104


_includes/cookie_banner.html
```diff
@@ -0,0 +1,20 @@
...
+        <p>This website uses cookies. <a href="/privacy/">Learn more</a></p>
+      </div>
+      <div class="col-md-6 col-sm-6 col-xs-12">
+        <a class="custom-button cookie-button pull-right" onclick="setCookie('cookiesAccepted', 'true', 365)">Ok!</a>
+        <script>
+          function setCookie(name, value, days) {
+            var d = new Date;
+            d.setTime(d.getTime() + 24*60*60*1000*days);
+            document.cookie = name + "=" + value + ";path=/;expires=" + d.toGMTString();
+            document.getElementById('cookie-banner').style.display = 'none';
```
marcoow: You could actually remove it from the DOM completely:

```js
let banner = document.getElementById('cookie-banner');
banner.parentNode.removeChild(banner);
```

_includes/cookie_banner.html
```diff
@@ -0,0 +1,20 @@
+<div class="cookie-banner" id="cookie-banner">
+  <div class="container">
+    <div class="row">
+      <div class="col-md-6 col-sm-6 col-xs-12">
+        <p>This website uses cookies. <a href="/privacy/">Learn more</a></p>
+      </div>
+      <div class="col-md-6 col-sm-6 col-xs-12">
+        <a class="custom-button cookie-button pull-right" onclick="setCookie('cookiesAccepted', 'true', 365)">Ok!</a>
```
marcoow: Instead of attaching the JS functionality to the element like this you could keep everything in JS actually (which I find a bit cleaner):

```js
Zepto(function($) {
  let banner = document.getElementById('cookie-banner');
  banner.on('click', function() {
    document.cookie = …
    …
  });
});
```

_includes/head.html
```diff
@@ -16,6 +16,14 @@
...
   <link rel="stylesheet" href="//fonts.googleapis.com/css?family=Ubuntu:300,400">
   <link rel="stylesheet" href="/css/app.css">
+  <script>
+    function checkCookies() {
+      if (document.cookie.includes('cookiesAccepted=true')) {
+        document.getElementById('cookie-banner').style.display = 'none';
+      }
+    }
+    window.onload = checkCookies;
+  </script>
```
marcoow: I'd move this into `_includes/cookie_banner.html` and turn it the other way round:

```js
Zepto(function($) {
  if (document.cookie.includes('cookiesAccepted=true')) {
    let banner = document.getElementById('cookie-banner');
    banner.parentNode.removeChild(banner);
  }
});
```

css/app.scss
```diff
@@ -190,6 +190,23 @@ pre.highlight {
   }
 }
 
+.cookie-banner {
```
marcoow: Can we reuse the `secondary-banner` class instead (and make the background solid `#2b98df`)?

marcoow (commented):
Nice! Just needs a bit of cleanup.

_includes/cookie_banner.html
```diff
@@ -0,0 +1,34 @@
...
+    <div class="row">
+      <div class="col-md-6 col-sm-6 col-xs-12">
+        <p>This website uses cookies. <a href="/privacy/">Learn more</a></p>
+      </div>
+      <div class="col-md-6 col-sm-6 col-xs-12">
+        <a class="custom-button cookie-button pull-right" id="cookie-button">Ok!</a>
+        <script>
+          Zepto(function($) {
+            if (document.cookie.includes('cookiesAccepted=true')) {
+              let banner = document.getElementById('cookie-banner');
```
marcoow: Using `let` is actually good but I'm not sure all the browsers that we need to support already support it so for now we'll have to stick with `var` 😞

_includes/cookie_banner.html
```diff
@@ -0,0 +1,34 @@
...
+              banner.parentNode.removeChild(banner);
+            }
+          });
+
+          Zepto(function($) {
+            let banner = document.getElementById('cookie-banner');
```
marcoow: Instead of using `document.getElementById` you can use Zepto: `$(#cookie-banner)`. That way you'd also not have to wrap `button` again before calling `on` on it (as Zepto extends it automatically).

_includes/cookie_banner.html
```diff
@@ -0,0 +1,34 @@
...
+            let button = document.getElementById('cookie-button');
+            $(button).on('click', function() {
+              var name = 'cookiesAccepted';
+              var value = 'true';
+              var days = 365;
+              var d = new Date;
+              d.setTime(d.getTime() + 24*60*60*1000*days);
+              document.cookie = name + "=" + value + ";path=/;expires=" + d.toGMTString();
```
marcoow: I think you can just use a static string here:

```js
document.cookie = "cookiesAccepted=true;path=/;expires=Mon, 24 Mar 2319 09:52:10 GMT";
```

_includes/cookie_banner.html
```diff
@@ -0,0 +1,34 @@
...
+              banner.parentNode.removeChild(banner);
+            });
+          });
```
marcoow: You can move both code blocks into the same `Zepto(function($) {` block - that way you can also share the `banner` variable and don't have to look it up twice:

```js
Zepto(function($) {
var banner = $('#cookie-banner');
if (document.cookie.includes('cookiesAccepted=true')) {
  …
}
var button = $('#cookie-button');
button.on('click', function() {
…
```

css/app.scss
```diff
@@ -188,8 +188,35 @@ pre.highlight {
...
+      &:hover {
+        color: #000;
+      }
+    }
+
+    .cookie-button {
+      text-decoration: none;
+      cursor: pointer;
+    }
+  }
```
marcoow: I think many of these styles are not needed when the banner has the `secondary-banner` class. Also, please define font-related styles and colors in the dedicated SCSS files.

css/color-scheme.scss
```diff
@@ -120,6 +120,16 @@ table {
     }
   }
 
+  &.cookie-button {
+    border-color: $secondary-highlight-color;
+    color: $secondary-highlight-color;
+
+    &:hover {
+      background-color: $secondary-highlight-color;
+      color: #252e39;
```
marcoow: Please always use variables for colors. I think these styles are likely not necessary at all though if you use the `secondary-banner` and `custom-button` classes.

trabaldini:
@marcoow the Zepto: $(#cookie-banner) didn't work, it always said, that it's undefined.
How can I solve that?

marcoow:
What specifically were you trying to do? This should work

```js
Zepto(function($) {
  var banner = $('#cookie-banner');
…
```

css/app.scss
```diff
@@ -188,8 +188,24 @@ pre.highlight {
   h4 {
     margin: 10px 0;
   }
+
+  &.cookie-banner {
+    position: fixed;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    padding-top: 20px;
```
marcoow: This is defined in `secondary-banner` already and isn't necessary here.

css/app.scss
```diff
@@ -188,8 +188,24 @@ pre.highlight {
...
+    padding-bottom: 5px;
```
marcoow: This should also not be necessary. You can set to bottom margin of `p` inside `cookie-banner` to `0` instead.

css/app.scss
```diff
@@ -188,8 +188,24 @@ pre.highlight {
...
+    z-index: 9999;
+    background-image: none;
+  }
+
+  .cookie-button {
+    cursor: pointer;
```
marcoow: Can you add this as a style for the `custom-button` instead?

css/color-scheme.scss
```diff
@@ -120,6 +120,15 @@ table {
...
   }
 
+  &.cookie-button {
+    border-color: $secondary-highlight-color;
+
+    &:hover {
+      background-color: $secondary-highlight-color;
+      color: $primary-highlight-color;
+    }
+  }
```
marcoow: I think you can just change this line https://github.com/simplabs/simplabs.github.io/blob/34f23dd3574df41103b1fd7ce49811e026ff7559/css/color-scheme.scss#L114 to `.contact-button, .cookie-button`

css/color-scheme.scss
```diff
@@ -227,6 +236,25 @@ table {
       }
     }
   }
+
+  &.cookie-banner {
+    background-color: $primary-highlight-color;
```
marcoow: This is define for the `secondary-banner` already.

css/color-scheme.scss
```diff
@@ -227,6 +236,25 @@ table {
...
+        color: $code-text-color;
+      }
+    }
+
+    .cookie-button {
+      &:hover {
+        color: $primary-highlight-color;
+      }
+    }
+  }
```
marcoow: All of these are the same as for the contact banner. Can the 2 be combined maybe?

marcoow (commented):
Looks good 👍  Just some styles that could potentially be cleaned up.

trabaldini:
Cleaned up the styling. If it's good this way I would make the cookie-banner for the other pages as well and link to the privacy-site of simplabs?

marcoow (approved):

marcoow:
@trabaldini: we need to do the same thing for ember-simple-auth.com and the workshop website. Can you create issues and prepare PRs?

PR #176, opened 2018-01-05
[WIP] Website facelift

marcoow:
### TODO

- [x] Main Menu: improve design, in particular hover menu
- [x] Services: Content
- [x] Services/Software Engineering: Content
- [x] Services/Team Augmentation: Content
- [x] Services/Ember: Content
- [x] Services/Elixir & Phoenix: Content
- [x] Work: Content and styles for cases list
- [x] About: Content
- [x] Overall: review (and improve) wording
- [x] get design for embedding links to cases etc. in content sites (e.g. embed case on http://localhost:4000/services/engineering/)
- [x] Overall: improve design
- [x] make blog text narrower again
- [x] fix title on Elixir & Phoenix page
- [x] add CTA on workshop pages somewhere at the top as well
- [x] ~~add tech page on Rails as well~~ (see follow-up #188)

### Start Page

- [x] add icons/background (incl. animation) #f03c15
- [x] add latest news teaser under _"Work"_ section (news are generally things we do, e.g. conferences we sponsor, talks we give, produce/client announcements), links to new _"News page"_ #f03c15
- [x] remove quote
- [x] add **some** description to client cases teasers
- [x] make it a bit clearer what the client logos are (e.g. _"Who we have worked with"_)
- [x] add brief description of our offerings (engineering, team aug, training - along with the respective icons) #f03c15
- [x] decide: display case teasers as banners or tiles?

### _"Services"_ Page

- [x] add icon/background #f03c15
- [ ] add icons for each service #f03c15

### _"Services/Full Stack Engineering"_ Page

- [x] add icon/background #f03c15
- [ ] add engineering process diagram (and explanation of it) #f03c15

### _"Services/Team Augmentation"_ Page

- [x] add icon/background #f03c15

### _"Services/Training"_ Page

- [x] add icon/background #f03c15
- [x] include reviews, architecture guidance etc. #f03c15
- [x] write text (benefits of trainings, jumpstart dev team etc.)
- [ ] design/implement catalog (on separate page; we can do this in a follow-up issue)

### _"Work"_ Page

- [x] add icon/background #f03c15
- [ ] add teaser for DD-WRT client case #f03c15

### _"Work/trainline"_ Page

- [x] add logo as title
- [x] replace background image (cannot use trainline's assets) #f03c15
- [ ] insert text block about mobile visits vs. app installs (we can do this in a follow-up issue when we add the tech page on pwas) #f03c15
- [x] review wording - expand on new sales channel etc.
- [ ] better screenshots
[...]

PR #179, opened 2018-01-22
WIP: Facelift styling

trabaldini:
The float of the contact-button caused the cut off of the body, so I made the button a bit smaller.
Looks better on the iPhone SE. Also some small layout changes for the contact banners.


marcoow:
thanks - added the fix to #176

PR #185, opened 2018-02-06
Blog post: handling webhooks in phoenix

niklaslong:
This is a first draft - let me know what you think!


blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
+---
+layout: article
+section: Blog
+title: Handling Webhooks in Phoenix
+author: "Niklas Long"
+github-handle: niklaslong
+twitter-handle: niklas_long
+---
+
+I recently had to implement a controller which took care of receiving and processing webhooks. The thing is, the API could send many different payloads to the webhook callback and there was only one tired and bloated controller action responsible for welcoming them. This didn't really seem to fit with the concept of using routes to identify controller actions. So I set out to find a better solution.
```
marcoow: > concept of using routes to identify controller actions

I'd say _"…my goal of keeping controller actions concise and focussed"_ or so. After all, this fits perfectly with the concept of using routes to identify controller actions as all webhooks go the the same route and that is mapped to the same controller action ;)

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
...
+
+<!--break-->
+
+#### tl;dr
+
+Use `forward` on `MyApp.router` to forward `conn` to a custom plug which transforms the `conn`; call `MyApp.WebhookRouter` from the plug. The `WebhookRouter` matches the modified `conn` to the controller actions.
```
marcoow: I'd maybe replace `conn` with _"the request"_ here to make clearer what this is about? Also I'd maybe say something like _"…to a custom plug which maps the request to a controller action based on the data in the request body"_. I'm not sure saying the plug modifies the request makes clear what the approach actually is.

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
...
+
+#### lv;e (long version; enjoy!)
+
+Let's restate the problem:
+* one controller action
```
marcoow: Maybe _"all incoming webhooks go to the same route, thus same controller action"_ to be clearer?

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
...
+* many different possible webhook payloads
+* different computation required depending on payload
+
+I actually came up with three different ways of solving this problem; only the last was to my satsifaction. I will, however, go through all three approaches, as they proved interesting learning opportunities.
+
+1. multiple function clauses for controller action
+2. plug called from endpoint
+3. plug and second router
```
marcoow: We need to fix the styles for `<ol>`s:

bildschirmfoto 2018-02-06 um 20 08 56

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
...
+def divide(params), do: #handle division
+```
+
+In this case, each controller action performs a specific function, the router maps each incoming request to these actions and the code is easily maintainable, well-structured and won't become jumbled over time. To achieve this however, we need to change a couple of things.
+
+First off, we need to interfere with the incoming `conn` before it hits the router so it will match our new routes. This is because the webhook callback url is usually always the same and doesn't depend on what event triggered it e.g., `"myapp.com/webhook"`. You would think we could create a plug for this and simply add it to a custom pipeline for the routes. The problem with this, is the router will invoke the pipeline **after** it matches a route. Therefore, we cannot modify the `conn`'s path in this pipeline and expect it to match our `add`, `subtract`, `multiply` or `divide` routes. If we want our new routes to match, we need to intercept the `conn` in a plug called in the endpoint. The endpoint handles starting the web server and transforming requests through several defined plugs before calling the router.
```
marcoow: > webhook callback url is usually always the same

I'd leave out the _"usually"_ here - after all you're talking about a very concrete example anyways.

marcoow: > e.g., `"myapp.com/webhook"`

Jekyll adds a link on that automatically. I'm not sure how to prevent that but if possible we should.

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
...
+
+```ruby
+defmodule MyApp.Endpoint do
+  # ...
+  plug(MyApp.WebhookShunt)
+  plug(MyApp.Router)
+end
+```
+
+And let's create a file called `webhook_shunt.exs` and add it to our `plugs` folder:
```
marcoow: > `webhook_shunt.exs`

Is `.exs` actually correct here? Should this not be `.ex`?

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
...
+    |> change_path_info(["webhook", "add"])
+  end
+
+  def call(conn, _opts), do: conn
+
+  def change_path_info(conn, new_path), do: put_in(conn.path_info, new_path)
+end
+```
+
+`change_path_info/2` changes the `path_info` property on the `conn`, in this case to `"webhook/add"`. You'll notice I also added a no-op function clause for `call/2`. If other routes are added and don't need to be manipulated in the same way as the ones above, we need to make sure the request gets through to the router unmodified.
```
marcoow: I'd add again here that this changes the path based on the request's payload as that's the important point really.

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
...
+
+This solution isn't great however. We are placing code in the endpoint, which will be executed no matter what the request path is. This isn't ideal, and may cause issues down the line.
+
+#### Forwarding conn to the shunt and calling another router
+
+Instead of intercepting the `conn` in the endpoint, we could forward it from the application's main router to the `WebhookShunt`, modify it and call a second router who's sole purpose would be to handle the hooks.
+
+1. The request hits router which has one path for all webhooks (`"/webhook"`)
+2. `conn` is forwarded to the WebhookShunt which modifies the path based on the request payload
+3. The WebhookShunt calls the WebhookRouter, passing it the modified `conn`
```
marcoow: I'm not sure why these have 2 be 2 different things to be honest.

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
...
+    conn
+    |> change_path_info(["webhook", "subtract"])
+    |> WebhookRouter.call(opts)
+  end
+
+  def change_path_info(conn, new_path), do: put_in(conn.path_info, new_path)
+end
+```
+
+You'll notice I've removed the no-op `call/2` function clause. This is because we no longer have to handle all potential requests like we did in the endpoint; we can focuse entirely on the hooks. Now, if we receive a webhook which doesn't match a clause of `call/2`, Phoenix will raise an error.
```
marcoow: > Phoenix will raise an error.

Maybe add something like _"…which is what we want as we don't know how to handle that webhook."_ or so?

marcoow (commented):
This is a great post! I left a few suggestions for making some things a bit clearer but overall this is really good 🎉 

Could you also have a look at the styling for `<ol>`s? The should basically be styled like `<ul>`s which have a special style when they appear in articles.

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
niklaslong: Do you think it's still important to mention the `WebhookRouter`, which sits between the plug and the controller actions? 

> _"…to a custom plug which maps the request to a controller action based on the data in the request body"_

^ might make it seem like the plug is calling the controller actions directly?

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
niklaslong: I couldn't find much after some quick googling - I changed it to _"my_app_url/webhook"_

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
niklaslong: whoops - yes!

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
niklaslong: how about:
> `change_path_info/2` changes the `path_info` property of `conn`, based on the request payload matched in `call/2`; in this case to `"webhook/add"`.

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
niklaslong: You mean the `Plug` and the `Router`? Or `2.` and `3.`?

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
niklaslong: fixed

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
...
+
+I recently had to implement a controller which took care of receiving and processing webhooks. The thing is, the API could send many different payloads to the webhook callback and there was only one tired and bloated controller action responsible for welcoming them. This didn't really seem to fit with my goal of keeping controller actions concise and focussed. So I set out to find a better solution.
+
+<!--break-->
+
+#### tl;dr
+
+Use `forward` on `MyApp.router` to forward `conn` to a custom plug which maps the request to a controller action based on the data in the request body. The `WebhookRouter` matches the modified `conn` to the controller actions.
+
+#### lv;e (long version; enjoy!)
```
geekygrappler: 😂

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
...
+Here's our controller with the multiple clauses:
+
+```ruby
+def hook(%{event: "addition"} = params), do: add(params)
+def hook(%{event: "subtract"} = params), do: subtract(params)
+def hook(%{event: "multiply"} = params), do: multiply(params)
+def hook(%{event: "divide"} = params), do: divide(params)
+```
+
+The request payload will match the clauses for the `hook/1` function and execute different functions depending on what `event` was passed in. This solution is simple and clean, but it doesn't fit well with the idea that a controller action should handle a specific request. The router serves no real purpose and our code has the potential to get very messy, quickly.
```
geekygrappler: Not sure about `clean`

geekygrappler: Also if I understand correctly this is solution what we currently have that is `tired and bloated`? If this is different to what we currently have maybe worth stating what the current controller does/is? If this is what we currently do on the controller I would say that and why it's not a good solution.

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
...
+defmodule MyAppWeb.Plug.WebhookShunt do
+  alias Plug.Conn
+
+  def init(opts), do: opts
+
+  def call(conn, _opts), do: conn
+end
+```
+
+The core components of a Phoenix appplication are plugs. This includes endpoints, routers and controllers. There are two flavors of `Plug`, function plugs and module plugs. We'll be using the latter in this example, but I highly suggest checking out the [docs](https://hexdocs.pm/plug/readme.html).
```
jayjayjpg: I think it would be 
> of a Phoenix app~~p~~lication are plugs

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
...
+
+```ruby
+scope "/", MyAppWeb do
+  forward("/webhook", Plugs.WebhookShunt)
+end
+```
+
+As long as your external APIs makes a request to this path when you do the setup for the webhook callbacks, every incoming request to this path will be forwarded to the WebhookShunt.
+
+Let's also add a second `call/2` clauses to handle the `subtraction` event in the WebhookShunt:
```
jayjayjpg: I guess this would be 

> second `call/2` clause~~s~~ to 

correct?

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
...
+
+  def call(conn, _opts), do: conn
+
+  def change_path_info(conn, new_path), do: put_in(conn.path_info, new_path)
+end
+```
+
+`change_path_info/2` changes the `path_info` property on the `conn`, based on the request payload matched in `call/2`, in this case to `"webhook/add"`. You'll notice I also added a no-op function clause for `call/2`. If other routes are added and don't need to be manipulated in the same way as the ones above, we need to make sure the request gets through to the router unmodified.
+
+This solution isn't great however. We are placing code in the endpoint, which will be executed no matter what the request path is. This isn't ideal, and may cause issues down the line.
```
geekygrappler: What issues down the line? Can you give an example please.

geekygrappler (approved):

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,202 @@
+---
+layout: article
+section: Blog
+title: Handling Webhooks in Phoenix
+author: "Niklas Long"
+github-handle: niklaslong
+twitter-handle: niklas_long
+---
+
+I recently had to implement a controller which took care of receiving and processing webhooks. The thing is, the API could send many different payloads to the webhook callback and there was only one tired and bloated controller action responsible for welcoming them. This didn't really seem to fit with my goal of keeping controller actions concise and focussed. So I set out to find a better solution.
```
jayjayjpg: That's a nice intro which already gives the reader a feeling of the scope of this blog post and also what the specific problem at hand is :thumbsup: Is there actually a term for this pattern of keeping module methods concise to handle one specific thing in the Elixir / Phoenix ecosystem?  If so, do you feel it might be good to mention it here as well?

jayjayjpg (commented):
That's a great and nicely structured blog post 👍  I haven't got too much of an insight into Elixir / Phoenix yet, so I only left a question regarding the intro and a few comments on spelling essentially. Hope that helps!

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
niklaslong: Would this make it clearer?
>   All incoming webhooks go to the the same controller action, albeit different clauses thereof; the router serves no real purpose and our code has the potential to get very messy, quickly. This is only a very slight improvement over the one clause of `hook/1` we were initially using.

geekygrappler: albeit & thereof have never made anything clearer in my experience.

It is not clear to me what the original state of the controller is. Therefore solution 1, multiple `hook` functions that will pattern match on the payload, seemed to me to be the most basic way of dealing with multiple possible payloads. Which is why I'm confused, but I assume there is an even more basic way of dealing with multiple payloads? Maybe:
```
def hook(payload) do:
  if payload.event == xxxx

  else if payload.event == yyyy
```
Might be worth writing out what the starting code for the controller is so people can understand the difference between starting point and solution 1.

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
niklaslong: 🤔 How about rephrasing like this:

> This solution isn't great however. We are placing code in the endpoint, which will be executed no matter what the request path is. Furthermore, the endpoint is only supposed to:
> - provide a wrapper for starting and stopping the endpoint as part of a supervision tree;
> - define an initial plug pipeline for requests to pass through;
> - host web specific configuration for your application.

> Interfering with the request to map it to a route at this point would be, at best, unidiomatic Phoenix. It would also make the app slower, and harder to maintain and debug.

There's one thing I'm not sure about: 
>  It would also make the app slower, and harder to maintain and debug.

This is not something I can speak of from personal experience, but rather a conjecture - can someone either confirm or deny this?

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
niklaslong: I suppose it's a pattern often found in functional programming? I've seen _function composition_ being thrown around. ~Although I don't know if it's a thing?~ It is a thing, but it refers to writing functions to do very specific things and then assembling them like Lego. 

The very specific function would apply to the controller actions; the assembling bit, not so much. Although they do act as wrappers for more specialised functions they run depending on the case. I don't know if it should be mentioned then...?

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
niklaslong: Point taken! Yes, the most primitive way of dealing with this is what you wrote above, i.e. one clause of `hook/1` with all the logic done inside the controller action.
Adding an example of that is a good idea.

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
marcoow: Yes, the plug and the router. Couldn't those be just one module?

niklaslong: I'm not sure really. I suppose the plug could call the controller actions directly by turning the plug into a "hybrid" between the plug and the router.

I think that might only be beneficial in a simple example. As the app grows in complexity it's quite nice to have clearly defined routes. It's more verbose, but it is clearer, in my opinion.

Having a router is also useful if you want to add pipelines to specific routes or namespace them.

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
niklaslong: I think one could also view it as an application of the single responsibility principle. The controller action handles a single type of event and dispatches the computation to other specialised functions; i.e., it only has one responsibility. But I might be completely wrong about this?

marcoow: SRP usually refers to classes and modules but I'd say what you say makes sense.

niklaslong: @marcoow, @jessica-jordan do you think I should just briefly mention SRP in the intro, add a small explanation (something like above) or not mention it at all?

marcoow: I think

> This didn't really seem to fit with my goal of keeping controller actions concise and focussed.

already gets the point across well.

niklaslong:
@jessica-jordan, @marcoow fixed wording and spelling.

@geekygrappler I added the explanation for why the plug shouldn't be in the endpoint. I also added the initial state example for the controller and router and rearranged the text a little. Let me know what you think!

Also, thank you all for the feedback!

Also I've just noticed some phrases which aren't as clear as I'd like, so I'll do some editing tomorrow with fresh eyes and resubmit.

marcoow:
@niklaslong: are you planning more changes or should this be good to go?

niklaslong:
@marcoow I still want to change a few things. After adding everyone's suggestions, the post had less consistency and clarity. I still have some editing to do but it should be ready in the next couple of days.

niklaslong:
@marcoow I did a couple things: 

- tried to make the post clearer in terms of steps (refactors). 
- simplified the code in the the last refactor for the `WebhookShunt` (using the event name as part of the path). 
- rewrote and rearranged quite a few parts so the post would flow better. 
- added some clarifications about `Phoenix.Endpoint`

Let me know what you think!

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,238 @@
+---
+layout: article
+section: Blog
+title: Handling Webhooks in Phoenix
+author: "Niklas Long"
+github-handle: niklaslong
+---
+
+I recently had to implement a controller which took care of receiving and processing webhooks. The thing is, the source site (API) could send requests with many different payloads to the webhook callback url and my first implementation consisted of only one bloated controller action responsible for handling them. This didn't really seem to fit with my goal of keeping controller actions concise and focussed. So I set out to find a better solution.
```
marcoow: > concise and focussed

I think I corrected 5 appearances or so of _"focussed"_ to _"focused"_ today per your comment on the website facelift branch ;)

marcoow: > The thing is, the source site (API) could send requests with many different payloads

I think this sounds a bit strange when reading it first (I wasn't sure what _"source site (API)"_ referred to initially). Maybe just say something along the lines of all webhooks going to the same endpoint with entirely different payloads and thus meanings?

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,238 @@
...
+
+`%Conn{}` -> `Router` -> `WebhookShunt` -> `WebhookRouter` -> `WebhookController`
+
+#### lv;e (long version; enjoy!)
+
+Let's restate the problem:
+
+* all requests are being sent to the same webhook callback url;
+* there are many different possible request payloads;
+* app requires different computation depending on payload.
```
marcoow: nitpick: I'd leave out the _";"_ and _"."_ here as the items are separated by being list items already.

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,238 @@
...
+end
+```
+
+All incoming webhooks go to the same route and therefore, the same controller action.
+
+It took three refactors to get to a satisfactory solution. I will, however explain each one in this post, as they are logical steps in reaching the final solution and proved interesting learning opportunities:
+
+1. multiple function clauses for controller action;
+2. plug called from endpoint;
+3. plug and second router.
```
marcoow: see above comment on _";"_ and _"."_

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,238 @@
...
+
+#### multiple function clauses
```
marcoow: lowercase looks a bit strange in a heading

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
```diff
@@ -0,0 +1,238 @@
...
+* host web specific configuration for your application.
+
+Interfering with the request to map it to a route at this point would be unidiomatic Phoenix. It would also make the app slower, and harder to maintain and debug.
+
+#### forwarding conn to the shunt and calling another router
+
+Instead of intercepting the `%Conn{}` in the endpoint, we could forward it from the application's main router to the `WebhookShunt`, modify it and call a second router who's sole purpose would be to handle the incoming webhooks.
+
+1. The request hits router which has one path for all webhooks (`"/webhook"`).
+2. `%Conn{}` is forwarded to the WebhookShunt which modifies the path based on the request payload.
```
marcoow: `WebhookShunt` is missing backticks, also some more occurences below

marcoow (commented):
Really good post. I left a few more comments on minor things but I think we should be good getting this merged tomorrow.

Also another tip: I find it helps reading to make some text bold here and there to highlight the most important things.

blog/_posts/2018-02-06-handling-webhooks-in-phoenix.md
niklaslong: > I think I corrected 5 appearances or so of "focussed" to "focused" today per your comment on the website facelift branch ;)

Yup, I can definitely see the irony in that 😅

marcoow (approved):

PR #198, opened 2018-05-28
Add "actix – an actor framework for the Rust programming language" blog post

### LINKED ISSUE #197: Blog post: actix - an actor framework for the Rust programming language
I've recently played around with https://actix.rs/ and could write a few words explaining what I did and how it can be used for certain tasks.

Turbo87:
Resolve #197


blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,264 @@
...
+
+
+#### Rust
+
+What is Rust?
+
+Aside from being "an iron oxide" according to [Wikipedia], Rust is also the
+name of a programming language that was originally invented by Graydon Hoare
+back in 2010. You can think about it as an interesting mix of high-level
+scripting languages like JavaScript or Ruby and a low-level high-performance
```
oliverbarnes: I'd remove the `scripting`, seems to imply they're not full-fledged languages. some people might get stung by it ;)

blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,264 @@
...
+language like C++.
+
+Similar to C++ it is a compiled language without a garbage collector. The fact
+that it usually uses LLVM to compile to machine code means that a lot of the
+optimizations that were developed to compile C/C++ code can also be applied
+to Rust programs.
+
+Similar to JavaScript or Python the language often feels more high-level than
+C/C++, and it has a built-in dependency manager called `cargo`. This is somewhat
+similar to `npm` in JavaScript or `pip` in Python, with the main difference
```
oliverbarnes: maybe mention javascript and ruby (`bundler`) first and mention python and pip as an aside, given we work with the first two?

blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,264 @@
...
+that `cargo` is also used to build and test your applications and libraries.
```
oliverbarnes: like `mix`, in Elixir

blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,264 @@
...
+
+The main advantage over all the other languages is Safety. The Rust compiler is
```
oliverbarnes: *safety*, lowercase

blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,264 @@
...
+
+[Wikipedia]: https://en.wikipedia.org/wiki/Rust
+
+
+#### Actors
+
+The "actor model" is the main primitive that powers the [Erlang] programming
+language and its descendant, [Elixir].
+
+An actor is a thing, that can only be interacted with using "messages". A
```
oliverbarnes: no need for the comma

blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,264 @@
...
+message can basically be anything that the actor can understand and in response
+to a message an actor is allowed to do several things, including:
+
+- send a response
+- send messages to other actors
+- change its own state
+
```
oliverbarnes: this summary about the actor model is great. but one thing I got confused about when first reading about the actor model is how it sounds like the object model, given it also involves message passing between things with behavior and state. maybe I don't understand it in full, still - the only distinction I make, based on Elixir, is that this "thing" has its own process within the VM. is this it? have no idea how it works with Rust. In any case, would be great to distinguish between the two, here

blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,264 @@
...
+[Erlang]: https://www.erlang.org/
+
+
+#### actix
+
+[actix] is the low-level actor framework that powers [actix-web], a
+[high-performance](https://www.techempower.com/benchmarks/#section=data-r15&hw=ph&test=plaintext)
+web framework for the [Rust] programming language.
+
+While [actix-web] is interesting and worth another blog post we will focus on
```
oliverbarnes: *...and worth another blog post, we will...` (adding the comma)

blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,264 @@
...
+    .map_err(|error| {
+        println!("An error occured: {}", error);
+    });
+
+Arbiter::handle().spawn(result);
+
+sys.run();
+```
+
+The first thing to do when using actix actors is to set up a [`System`], that
```
oliverbarnes: what do the square brackets mean here? is it a syntax thing in Rust?

oliverbarnes: just noticed the link on the rendered page, but don't understand how it's linking ;) Anyway, `[Arbiter]` below isn't linked (yet?)

blog/_posts/2018-05-28-actix.md
oliverbarnes: Now that I've finished the post, seems that would mean threads in Rust

oliverbarnes (changes requested):
👏 great post, really well written. Rust syntax still seems alien to me, but this was an interesting first read about it, for me. Just the small comments below

blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,264 @@
...
+defined earlier. In the `handle()` method we increment the `count` state and
+then return the new value to tell actix that this is our response to the
+message.
+
+[actix-web]: https://github.com/actix/actix-web
+
+
+##### Running our `CounterActor`
+
+While building the actor was relatively okay, running it is unfortunately still
```
niklaslong: Small suggestion: _easy_ instead of _okay_

blog/_posts/2018-05-28-actix.md
niklaslong: like hex in Elixir? (used by mix)

niklaslong (commented):
I really enjoyed this post (looking forward to the next one as well)! 🎉 
+1 on what @oliverbarnes mentioned and I made some small wording suggestions.

blog/_posts/2018-05-28-actix.md
Turbo87: yeah, with OOP you call things on the class instances directly. with the actor model you have this indirection through the actor framework that abstracts away the communication between the actors so that you can distribute them to different threads, processes or even entirely different machines. actix currently only supports different threads, but the rest is on the roadmap too.

blog/_posts/2018-05-28-actix.md
Turbo87: it was missing the following line below the section:

```
[`Arbiter`]: https://actix.rs/actix/actix/struct.Arbiter.html
```

added now, good catch 👍

oliverbarnes (approved):

blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,267 @@
...
+<!--break-->
+
+
+#### Rust
+
+What is Rust?
+
+Aside from being "an iron oxide" according to [Wikipedia], Rust is also the
+name of a programming language that was originally invented by Graydon Hoare
+back in 2010. You can think about it as an interesting mix of high-level
```
chriskrycho: Actually started all the way back in 2006! See [this FAQ entry](https://www.rust-lang.org/en-US/faq.html#is-this-project-controlled-by-mozilla).

Turbo87: whoops... don't believe everything Wikipedia tells you 😜

blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,267 @@
...
+```
+
+The `CounterActor` class in this example is initialized with an internal state
+called `count` that is set to zero and it responds to `plus-one` messages by
+increasing the `count` state and returning the new value.
+
+The complexity of this actor is relatively low, and that is because it is
+usually hidden in the actor frameworks that are used to run these types of
+primitives in the end. One example of such an actor framework is [actix], which
+we will have a closer look at now.
```
marcoow: I think this sentence is unnecessary hard to understand as it initially refers to the actor from the example but then makes a general point regarding actors which is not actually true for the actor in the example (as that's just a base JS class). So basically what I'm saying is, this gets simpler if you change it to _"The complexity of actors is relatively low…"_.

blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,267 @@
...
+    }
+}
+```
+
+Since Rust is a typed language all structures need to be declared upfront. In
+the snippet above we first import all the necessary things from the
+`actix::prelude` module, and then we define how a `PlusOne` message should look
+like. In the JavaScript implementation the `message` had a `type` property, but
+since we have a strict type system available in Rust there is no need to
+explicitly declare that. That leaves us with an empty `PlusOne` message,
```
marcoow: I think it might be worth explaining this a bit more, e.g. by saying that we need the `type` property on the message in JS so we can identify what kind of message we're receiving while in Rust we can declare the handler to handle messages of type `PlusOne` as that type exists and has meaning in Rust. I think it might not be 100% clear to everyone what you're trying to say here otherwise.

blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,267 @@
...
+handles all those actor interactions for us. We do so by calling
+`actix::System::new()` and passing it a name.
+
+Next we start an [`Arbiter`] in a new thread, that runs our `CounterActor`. If
+that sounds like a foreign language to you, don't worry, I had the same feeling
+at first. For now all you need to know is that `Syn` means that it is running
+in a separate thread, and that the `Arbiter` is the thing that controls that
+thread.
+
+The `Arbiter::start()` call returns us an `Addr`, that we can use to talk to
```
marcoow: Does _"returns us"_ work in english? I think the _"us"_ should be removed?

blog/_posts/2018-05-28-actix.md
```diff
@@ -0,0 +1,267 @@
...
+the actor. The `Addr` struct has methods like `send()` that can be used to send
+messages to the actor and receive their responses.
+
+As mentioned earlier, Rust is very strict around data ownership. Since we use
+the `counter` variable for the first `send()` call, we are (at least currently)
+not allowed to reuse it inside the callback. Instead we need to create a
+`clone()` and use that one instead.
```
marcoow: This doesn't actually explicitly get mentioned earlier. I'd add just one sentence saying explicitly that Rust checks who owns which data and thus might modify data and that the concept of _"borrowing"_ exists etc. (I'm not really familiar with the details - you'll know much better) to https://github.com/simplabs/simplabs.github.io/pull/198/files#diff-129064ba4cecc35f3701f92524e29ea2R44 so that this gets clearer.

marcoow (commented):
This is a very good post 👍 I left some comments on wording where changes could make the text easier to understand, in particular for people not already familiar with the matter.

I'd release this one next week and the 2nd part in 2 weeks then.

marcoow:
> I'd release this one next week and the 2nd part in 2 weeks then.

Not sure why I was thinking there was a second part of this already :)

PR #205, opened 2018-06-15
Add "actix – a basic TCP client" blog post

Turbo87:
related to #198


Turbo87:
a new version of `actix` was released last night, with a rather large amount of breaking changes. I still think this blog post makes sense, but we should include some form of notice that it is targeting the older v0.5 version.

blog/_posts/2018-06-15-actix-tcp-client.md
```diff
@@ -0,0 +1,405 @@
...
+and then connecting all the pieces together via:
+
+```rust
+ctx.add_stream(line_reader);
+```
+
+right below the line that constructs the `FramedRead` instance.
+
+I won't spoiler anything, but take a bit of time and see what happens now if
+you start `cargo run` again... 🚀
```
niklaslong: ❤️ 😍

blog/_posts/2018-06-15-actix-tcp-client.md
```diff
@@ -0,0 +1,405 @@
...
+```
+
+This combination of `LinesCodec` and `FramedRead` from the `tokio-codecs` crate
+can be used to work with the `Read` trait in a more comfortable way. The
+`FramedRead` instance acts as an asynchronous stream of values that we could
+call `poll()` on, or we'll attach it to the actor in a more convenient way
+by implementing the `StreamHandler` trait from `actix`:
+
+```rust
+impl StreamHandler<String, io::Error> for TcpClientActor {
```
niklaslong: I don't think you mention it, but I think we need to `use std::io;` for this to work?

niklaslong (commented):
This is a great and really fun post 🎉! I didn't notice any language mistakes when I was following along, but I'll check again if I get a chance. 

Small suggestion: I think it might be helpful to some if you could link to the full code at the end of the post (maybe a gist or something)?

blog/_posts/2018-06-15-actix-tcp-client.md
Turbo87: oh, good point, must have missed that.

blog/_posts/2018-06-15-actix-tcp-client.md
```diff
@@ -0,0 +1,405 @@
...
+Before we can start to implement our TCP client actor we need to tell `cargo`
+about the `actix` dependency. For that we open the `Cargo.toml` file in the
+project and add the following line *after* the `[dependencies]` declaration:
+
+```toml
+actix = "0.5"
+```
+
+Now, whenever we build the project again, `cargo` will make sure to download
+the necessary dependencies from crates.io and build them, before building the
```
marcoow: I **think** _"build them ~~,~~ before building"_ (no comma)

blog/_posts/2018-06-15-actix-tcp-client.md
```diff
@@ -0,0 +1,405 @@
...
+actual project. `cargo` is using a build cache though, so don't worry if it
+takes a long time if you first try to build it, the following builds should be
+much faster.
```
marcoow: I'd split this into 2 sentences or use a – instead of the second comma, e.g. _"build it – the following builds should be
 +much faster."_.

blog/_posts/2018-06-15-actix-tcp-client.md
```diff
@@ -0,0 +1,405 @@
...
+```
+
+At this point the above implementation should contain no surprises anymore. The
+`ConsoleLogger` actor implementation looks roughly like the other actors we have
+already implemented in this and the previous blog post. All it does is listen
+for `ReceivedLine` messages, and print them to the terminal once received.
+
+Next, we will need to adjust our `TcpClientActor` implementation to no longer
+print messages by itself, but instead forward them to the second actor. For that
+the `TcpClientActor` needs to now where to send them. We could save the
```
marcoow: _"…needs to **k**now where…"_

marcoow (commented):
Awesome post 🙌

marcoow (approved):

PR #211, opened 2018-07-31
replace client medikompass with cardstack

marcoow:


jayjayjpg (approved):
This works well locally :+1:

PR #213, opened 2018-08-07
Fix blog article layout

marcoow:
This removes the extra unnecessary line in the blog article layout. That was only necessary when we had comments to separate that section but since comments have been removed the line serves no purpose and is inconsistent with the layout of the rest of the page:

bildschirmfoto 2018-08-07 um 10 18 59


mansona (approved):
looks good 👍

PR #215, opened 2018-08-08
ignore .jekyll-metadata

marcoow:
This seems to be a new file that Jekyll now creates.


marcoow:
@geekygrappler: is this good to merge?

PR #216, opened 2018-08-13
fix styling of job posting

marcoow:
This fixes some styling aspects of the job posting and brings it more in line with the other pages

### DIFF
--- a/jobs/index.html
+++ b/jobs/index.html
@@ -76,7 +76,7 @@ title: Jobs
             <li><strong>collaborating</strong> well in a team environment</li>
             <li>showing attention to detail by <strong>writing tested code</strong></li>
           </ul>
-          <p>If you have experience with any of the following, we'd consider those a bonus (but not require them for a successful application)</p>
+          <p>If you have experience with any of the following, we'd consider those a bonus (but not require them for a successful application):</p>
           <ul>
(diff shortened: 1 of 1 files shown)


jobs/index.html
```diff
@@ -69,15 +69,14 @@ <h3>Responsibilities</h3>
...
         <div class="col-sm-6">
+          <p>We expect the following skills:</p>
           <ul>
             <li><strong>working experience</strong> with <strong>JavaScript and Ember.js</strong>, either in commercial projects or open source</li>
             <li><strong>collaborating</strong> well in a team environment</li>
             <li>showing attention to detail by <strong>writing tested code</strong></li>
           </ul>
-          <hr />
-          <h4>Bonus skills <span class="header-note">(but not required for a successful application)</span></h4>
+          <p>If you have experience with any of the following, we'd consider those a bonus (but not require them for a successful application)</p>
```
jayjayjpg: Since these are not header titles anymore, would it make sense to add a double colon `:` after this sentence as well?

jayjayjpg (commented):
This works great locally, I just left one more comment on how the text could possibly be even more aligned with the new styling. Apart from that this seems all :+1:

jobs/index.html
marcoow: absolutely - added 👍

jayjayjpg (approved):

PR #233, opened 2018-11-05
fix logo animation in main nav

### LINKED ISSUE #226: [Main Navigation] Possible animation error shows duplicate logo
image

wadie:
fixes #226

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -3,7 +3,7 @@
 <nav class="navbar navbar-expand-md navbar-light fixed-top">
     <div class="container">
         <a class="navbar-brand" href="/">
-            <img alt="Simplabs" class="navbar-brand-image" src="/images/simplabs-logo-ball-blue.svg" />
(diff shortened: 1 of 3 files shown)


_includes/nav.html
```diff
@@ -3,7 +3,7 @@
 <nav class="navbar navbar-expand-md navbar-light fixed-top">
     <div class="container">
         <a class="navbar-brand" href="/">
```
ghislaineguerin, line 5: @wadie I worry that without CSS we don't have any reference to our logo or brand here. Is it possible to implement this differently?

ghislaineguerin:
@wadie I've added a comment on the code. I think we need to still see our brand without CSS, just not sure how that conflicts with the animation.

_includes/nav.html
wadie: I'll make sure we use normal `` tags in order to have `alt` for the logo. 👍

ghislaineguerin:
👍 looks great

PR #237, opened 2018-11-06
Content Strategy Updates

ghislaineguerin:
In order to increase user engagement across our site, and decrease bounce rate, we need to provide clear, context-relevant navigation paths to our visitors. 
Work section client stories must provide links to services, blog posts, when available, and a call to action that is related to the service provided. This will not only drive up engagement but also help us to learn which content and needs are more relevant to our users.

trainline___simplabs

### DIFF
--- a/trainline/index.html
+++ b/trainline/index.html
@@ -51,7 +51,7 @@ title: Trainline
 
             <p class="mb-5">
               Simplabs worked closely with Trainline’s in-house engineering team to deliver an app-level experience for their mobile users.
-              A dedicated site for mobile devices, allowed Trainline's customers to benefit from the best of website performance, without the need to install an app —they could simply navigate to the site on any browser.
+              A dedicated site for mobile devices, allowed Trainline's customers to benefit from the best of website performance, without the need to install an app —they could now simply navigate to the site on any browser.
             </p>
 
             <p class="mb-5">
--- a/trainline/index.html
+++ b/trainline/index.html
@@ -4,13 +4,19 @@ title: Trainline
 ---
 
 
-<header class="page-trainline">
+<header class="page-trainline mb-5">
   <div class="jumbotron jumbotron-fluid jumbotron-fluid-trainline">
     <div class="container container-jumbotron">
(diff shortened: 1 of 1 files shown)


trainline/index.html
```diff
@@ -30,27 +36,40 @@ <h1 class="jumbotron-header">
...
+              Simplabs worked closely with Trainline’s in-house engineering team to deliver an app-level experience for their mobile users.
+              A dedicated site for mobile devices, allowed Trainline's customers to benefit from the best of website performance, without the need to install an app —they could simply navigate to the site on any browser.
             </p>
 
-            <p class="introduction-text">
-              simplabs worked closely with Trainline’s in-house engineering team and built a mobile web app to
-              accompany the existing desktop app. A dedicated site for mobile devices allowed Trainline to better
-              serve customers on the go and leverage the full market potential.
+            <p class="mb-5">
+              At simplabs, we believe that both end-users and business can benefit from the <a href="#">advantages provided by Progressive Web Applications</a>, which is why we recommend them as the better choice for our clients looking to provide seamless, engaging experiences to their mobile users.
```
marcoow, line 58: What we built for trainline was not strictly a Progressive Web App though but a normal mobile app.

marcoow (commented):
I think this is the right direction for sure but we shouldn't be saying _"Progressive Web App"_ here as it was strictly just a mobile web app but not a PWA.

trainline/index.html
ghislaineguerin: @marcoow I was about to ask that. I'm still going to revise the content, but the main goal is to link from this client story to relevant content on the site, like blog and services so that our bounce rate decreases. I will probably need to talk to devs to find out more about the technology used for this and its advantages.

marcoow: I think your original point makes sense - it's better for users to just open the mobile website than having to install the app. We should just maybe say _"mobile web app"_ or so as when we say it's a PWA and then people realize it's actually not that's not great :)

ghislaineguerin: @marcoow completely agree. I will update the content.

PR #240, opened 2018-11-08
Fix SCSS

ghislaineguerin:
This branch contains the fix for the sass task and path errors related to SCSS files.


wadie (approved):
Good to go 👍

PR #241, opened 2018-11-09
Css Refactor

ghislaineguerin:
This pull request fixes the following issues:
- Specific and/or nested CSS selectors that prevented new content from being added were removed
- Components that happened in multiple pages are now reusable includes

What's missing:
- Must create utility classes to apply background effects to any component
- Blog refactor
- Hero section refactor
- Improved circle background CSS

The main goal of these fixes is preventing the spread of technical debt onto new pages. 

@wadie I think the JS errors on the build will be fixed once we merge your changes.

PR #242, opened 2018-11-12
js dependencies

wadie:
I still need to compile into a minified main.js and add Zepto script to header.html

will be merged with #241


package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "http-server": "./node_modules/.bin/http-server",
     "sass": "sass --watch ./application/stylesheets/scss/custom.bootstrap.scss:./application/stylesheets/css/custom.bootstrap.css",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "conpile": ""
```
ghislaineguerin: @wadie is this a typo?

wadie: Fixed, it's not needed.

ghislaineguerin: 👍

ghislaineguerin:
@wadie do we have gulp watch for watching both js and CSS? this would replace sass watch I assume

gulpfile.js
```diff
@@ -0,0 +1,24 @@
...
+    watch = require('gulp-watch');
+    terser = require('gulp-terser');
+
+gulp.task('js', function(){
+    return gulp.src('./js/*.js')
+        .pipe(gp_concat('main.js'))
+        .pipe(gulp.dest('./js/'))
+        .pipe(gp_rename('main.js'))
+        .pipe(terser())
+        .pipe(gulp.dest('./js/'));
```
ghislaineguerin: @wadie we are missing this for when there are CSS changes right?

PR #247, opened 2018-11-23
Improve build

marcoow:
* update dependencies
* remove `package.json`
* move JS assets to their own folder and ignore that in the Jekyll build

This should significantly improve the build performance


ghislaineguerin:
@marcoow what about the bootstrap dependency? It is required to build the CSS files.
`@import "./../../node_modules/bootstrap/scss/functions";`
Would we remove this if we get rid of package.json?

marcoow:
@ghislaineguerin: I'm fixing that in a separate PR

marcoow:
This should be good to merge actually - we don't need the netlify deployment for this anyway.

ghislaineguerin:
@marcoow 👍

ghislaineguerin (approved):

PR #249, opened 2018-11-23
Perf improvements

marcoow:
This implements some performance improvements:

* optimized all images with ImageOptim
* removed redundant JS code
* dropped some of the font weights

Unfortunately adding a service worker in a good way will require us to change the deployment so we can have fingerprinted assets. I'll do that at a later point.


ghislaineguerin:
@marcoow this looks great. 👍

PR #219, opened 2018-09-06
Redesign

wadie:
Everything should be good, there were a lot of links to edit as well so let me know if I missed anything. We also need to set the categories for all blog posts.

Waiting for your input! 👍 

**deployed to https://angry-liskov-9cac36.netlify.com**

PR #254, opened 2018-11-26
add category colors

wadie:
Added the category colors that Marco [suggested](https://github.com/simplabs/simplabs.github.io/pull/219#issuecomment-441285403)


marcoow (changes requested):
the colors for the `js` category seem wrong:

bildschirmfoto 2018-11-26 um 16 31 42

wadie:
It works now, but I don't think the color looks good. maybe a darker yellow ?

marcoow:
Those are the official colors for the respective frameworks and languages though - I think it'll be good to use them. What do you think @ghislaineguerin?

This contains a bunch of unrelated changes now - can you rebase on `blog-refactor` and remove 48688add41bfd6a3362971622b6c46f127e7b0b0 from this PR (as well as the changes to `Gemfile`/`Gemfile.lock` - I don't remember the exact reason but we can't use Ruby 2.5 for now)?

ghislaineguerin:
@marcoow We can always darken the official colors if they are breaking the background contrast accessibility requirements. Otherwise, we should always use the official colors.

ghislaineguerin:
image
This is the same color with lower luminosity.

ghislaineguerin:
@marcoow I think we can merge this once you approve the changes.

ghislaineguerin (approved):

stylesheets/components/blog.scss
```diff
@@ -85,6 +84,53 @@
...
+          color: rgba($text, 0.5);
+          font-size: 0.75em;
+          font-weight: $font-weight-normal;
+          line-height: 1.3333em;
+
+          @include media-breakpoint-up(md) {
+            font-size: 0.875em;
+            line-height: 1.7142em;
+          }
+        }
```
marcoow, line 133: I think all of these changes are unrelated?

stylesheets/components/blog.scss
```diff
@@ -120,6 +198,10 @@
   max-width: 46em;
 }
 
+.blog-js,
```
marcoow, line 201: Will this actually be used? The category of the posts is `javascript` so I'd assume this will not actually be used (but then if might not be necessary at all - same for the other ones)?

marcoow (changes requested):
This still contains some unrelated changes and I think one error (`js`/`javascript`). Also there are lots of commits while in reality I think there should only be one - read up here on rewriting commit history to clean up commits before merging:

* https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History
* https://hackernoon.com/mastering-git-why-rebase-is-amazing-a954485b128a
* https://backlog.com/git-tutorial/rewriting-history/

ghislaineguerin (changes requested):
I think this will now conflict with what's currently on blog-refactor. I suggest starting from a new branch and adding the missing colors and classes.

PR #259, opened 2018-12-05
prevent horizontal scroll

marcoow:
closes #258

PR #261, opened 2018-12-05
remove specific Ruby version

wadie:
closes #259


marcoow (commented):
`overflow-x` gets overridden by `overflow` 4 lines below - this can't really work. Apparently it's not possible to combine `overflow-x: hidden` and `overflow-y: visible` anyway. See #259 for an approach that works. Also the Ruby version change is unrelated.

PR #283, opened 2019-03-01
Added dockerfile(s) for local Docker development

mrmrcoleman:


jekyll-serve/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+WORKDIR /site
+
+ENTRYPOINT [ "bundle", "exec", "jekyll" ]
+
+CMD [ "--help" ]
+
+##
+##
+## New stage in multi-stage image
```
Turbo87, line 17: do you know why this is using a multi-stage image?

mrmrcoleman: See the README here: https://github.com/BretFisher/jekyll-serve

Do you think people will want to use the various versions, or should I just prune the Dockerfile?

PR #286, opened 2019-03-06
Refactor Shapes to use blocks

jorgelainfiesta:
This PR refactors the styles for Shapes to `block` as suggested by `css-blocks`. In some cases the shapes were extracted to be independent components, but in others the styles were just imported into their respective routes as it seemed like the shapes there were unlikely to be reused.


src/ui/styles/blocks/shape/shape-side.block.css
```diff
@@ -0,0 +1,25 @@
+:scope {
+  block-name: shape-side;
+  transform: rotate(30deg);
+	overflow: hidden;
+	position: absolute;
```
marcoow: Can you fix the formatting bugs like this? There are several of this in other files as well…

marcoow (commented):
This looks very good to me! It's exactly the level of reuse that makes sense for now I think. We can always go back later and come up with some more reusable components e.g. for the shapes but doing that now I think would just unnecessarily limit us and make things overly complex.

src/ui/styles/blocks/shape/shape-side.block.css
jorgelainfiesta: Yeah thanks for catching that, I think it's an annoying misconfiguration in my editor that keeps automatically changing the tabs to 4 spaces on every new file.

marcoow (approved):

PR #289, opened 2019-03-08
Fix linting issues

jorgelainfiesta:
Applied format to files and changed components to follow PascalCase.


marcoow (changes requested):
Can you adapt the dir names so they match the upper-case component names?

jorgelainfiesta:
Yeah I had done that but for some reason those changes were not committed. Something's odd with my editor. Will fix it.

marcoow (approved):

PR #292, opened 2019-03-14
New header

jorgelainfiesta:
Implements the new Navigation header and applies it to all the routes.


src/ui/components/Navigation/template.hbs
```diff
@@ -0,0 +1,24 @@
+<div>
+  <div class="navbar-brand">
+    <a href="/a" data-navigo>
```
marcoow: Can we map the `Homepage` component to `/` now?

marcoow (changes requested):
The contact button looks slightly different in Invision, e.g.: https://projects.invisionapp.com/d/main#/console/16581855/343841149/preview

Bildschirmfoto 2019-03-15 um 09 29 28


Also I think we'd want hover effects?

src/ui/components/Navigation/template.hbs
marcoow: fixed it

marcoow (approved):

marcoow:
I fixed the work with us link as well. I guess for hover effects etc. we have to revisit most of the page anyway later in one go.

PR #294, opened 2019-03-14
New footer

jorgelainfiesta:
New footer component for the website.

### DIFF
--- a/src/ui/components/Footer/stylesheet.css
+++ b/src/ui/components/Footer/stylesheet.css
@@ -1,0 +1,42 @@
+@block screen from "../../styles/blocks/screen.block.css";
+
+:scope {
+  block-name: footer;
+  display: flex;
+  flex-flow: row wrap;
+  justify-content: space-between;
+  margin: 0 10em;
+}
+
+.branding {
+  min-width: 200px;
+  flex-shrink: 0;
+  font-size: 1rem;
+  margin-right: 10rem;
+}
+
+.logo-wrap {
+  display: block;
+  width: 200px;
+  margin-bottom: 2rem;
+}
+
+.menu {
+  margin-right: 3rem;
+  flex-shrink: 0;
+  font-size: 1rem;
+  display: block;
+  margin-top: 0.6rem;
+}
+
+.menu-title {
+  font-weight: bold;
+  color: #000;
+}
+
+.menu-link {
+  display: block;
+  text-decoration: none;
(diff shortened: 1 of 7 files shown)


src/ui/components/Footer/stylesheet.css
```diff
@@ -0,0 +1,42 @@
+@block screen from "../../styles/blocks/screen.block.css";
```
marcoow, line 1: If we're using a block called _"screen"_ in a component like this that is not a _"screen"_ really but only a part of a screen, I think this indicates some potential for refactorings?

src/ui/components/Footer/template.hbs
```diff
@@ -0,0 +1,61 @@
...
+  </nav>
+  <nav class="menu">
+    <strong class="menu-title">
+      Expertise
+    </strong>
+    <a href="/expertise/ember" class="menu-link" data-navigo>
+      Ember.js
+    </a>
+    <a href="/expertise/elixir-phoenix" class="menu-link" data-navigo>
+      Elixir & Phoenix
```
marcoow, line 30:
```suggestion
      Elixir &amp; Phoenix
```

src/ui/components/Footer/template.hbs
```diff
@@ -0,0 +1,61 @@
...
+  </nav>
+  <nav class="menu">
+    <strong class="menu-title">
+      Work
+    </strong>
+    <a href="/expertise/ember" class="menu-link" data-navigo>
+      Trainline
+    </a>
+    <a href="/expertise/elixir-phoenix" class="menu-link" data-navigo>
+      Elixir & Phoenix
```
marcoow, line 41: This link is wrong - check e.g. https://projects.invisionapp.com/d/main#/console/16581855/343841149/preview

marcoow (changes requested):
The code looks good but I think the content is not correct and there are some layout problems:

Bildschirmfoto 2019-03-15 um 09 23 14

src/ui/components/Footer/stylesheet.css
marcoow: I'll address that in a follow-up

marcoow (approved):

marcoow:
I fixed the wrapping and wrong link - I think for text color and size etc. we'd have to go through the code later again anyway and make sure we're using all the correct vars everywhere etc.

PR #298, opened 2019-03-16
extract fluid-image block

marcoow:
I think it's better to keep separate things in separate blocks instead of adding everything to one block that's then reused everywhere. We should probably continue this and split `screen.block.css` and other big blocks further apart.


ghislaineguerin (approved):
👍 this makes sense, thanks

PR #301, opened 2019-03-19
Add page components

ghislaineguerin:


ghislaineguerin:
@wadie I think we should merge this before there are more conflicts with the work you are doing. I will not need branches as complex anymore.

ghislaineguerin:
@marcoow @wadie I will go ahead and merge this otherwise it will create conflicts as there are many changes already.

PR #302, opened 2019-03-20
Fix routing

marcoow:
This fixes the routes and names of some components to better match what they actually contain.


ghislaineguerin:
👍 looks good. thanks

PR #306, opened 2019-03-26
Expedition

wadie:
Expedition use case.

We might still need the images for it


src/ui/components/ExpeditionCaseThumb/stylesheet.css
```diff
@@ -0,0 +1,24 @@
...
+  background-attachment: fixed;
+  background-repeat: no-repeat;
+  background-size: cover;
+  background-blend-mode: soft-light;
+}
+
+.breethe-right-background {
+  background-color: resolve("shape-right.background");
+  background-color: #6f42c1;
+}
```
marcoow: 👀

marcoow (changes requested):
`package-lock.json` should not be added - we use yarn and have `yarn.lock` already

ghislaineguerin:
@wadie Do you have an issue where the images are 404 on stage?

wadie:
@ghislaineguerin only the images that I haven't set yet

Screenshot_2019-03-26 Hello Glimmer

ghislaineguerin:
@wadie can you add placeholder images to verify the layout and I'll replace them later?

wadie:
updated

ghislaineguerin:
@wadie the bottom image should remain the same, it's supposed to be a link to another case study so the color should remain the same. Other than that, I'd replace the quote under client with the actual client description. Thanks

ghislaineguerin:
@wadie unrelated, but when clicking on a footer link why doesn't the page jump back to top?

wadie:
Yeah the route changes but the focus stays on the footer and it's happening with all the pages/links.

I think it's an unrelated issue, so we can merge this PR then I'll create a new branch to fix it.

ghislaineguerin (approved):

PR #320, opened 2019-04-04
Fix prerender

marcoow:
This should fix the prerenderer, mainly by un-limiting the number of event listeners and not pre-rendering concurrently but sequentually which doesn't work with critical apparently.

This also runs `yarn build` on CI so we can be sure we don't break the pre-rendering.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,8 @@ jobs:
       script: yarn lint:js
     - state: test
       script: yarn lint:hbs
+    - state: test
+      script: yarn build
 
 notifications:
   email: false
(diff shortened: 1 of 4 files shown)


ghislaineguerin (approved):
👍

.travis.yml
```diff
@@ -34,6 +34,8 @@ jobs:
       script: yarn lint:js
     - state: test
       script: yarn lint:hbs
+    - state: test
```
Turbo87, line 37: what is `state`?

marcoow: ha, yeah that should be `stage` I guess :)

PR #323, opened 2019-04-04
Asset path fix

ghislaineguerin:
This fixes all asset paths to be /asset instead of just 'asset'


ghislaineguerin:
@marcoow please review the fix for asset path. thanks

marcoow (approved):

PR #324, opened 2019-04-04
Generate blog posts components at build time

### LINKED ISSUE #272: Generate blog post components on the fly
For blog posts, we'd like to maintain the same mechanism we're using now for the new simplabs.com as well. That means posts are written as markdown files with some front matter that describes the category, author etc. In the end, those posts would end up as glimmer components that are routed to individual routes. These components would have to be generated (not generated as actual files in `src/ui` but generated and added to the app tree) as part of the build.

### TODO

- [x] define what a blog post component looks like (this is before we have the auto-generation so that we know what we'll be auto-generating eventually)
- [x] auto-generate blog post components on the fly during the build and add them to the app tree
- [x] [add routes](https://github.com/simplabs/simplabs.github.io/blob/new-simplabs/config/routes-map.js) for the auto-generated blog posts during build time
- [x] add a blog index page that lists all blog posts (or the last 3 or whatever); or maybe this is a curated page?
- [ ] add a [service](https://gist.github.com/mixonic/acbc9b8d57819e2f2c77b206ab7f422e) that knows which blog post components are for which tags, by which author etc. so that we can implement something like _"here are the 3 latest ember posts"_

marcoow:
This creates blog post components and the respective URLs during build time based on markdown files.

### TODO

- [x] cleanup - the code is a mess currently
- [x] apply the right classes to all content elements (paragraphs, images etc.)
- [x] extract front matter meta data and inject that into the template in the right places
- [x] build the complete blog post components including header, footer etc.
- [x] build `BlogRelatedArticle` based on the existing blog posts (`BlogRelatedArticle` would not be its own component anymore actually but be inserted into the respective blog post)
- [x] build main `Blog` list page based on the blog posts
- [x] add missing twitter handles to posts
- [x] add missing bios to posts (only Mark missing)
- [x] add missing author images

closes #272


marcoow:
some sample blog posts:

* https://deploy-preview-324--gallant-hawking-42050a.netlify.com/blog/2019-03-29-qonto-project/
* https://deploy-preview-324--gallant-hawking-42050a.netlify.com/blog/2019-03-18-emberconf-update/
* https://deploy-preview-324--gallant-hawking-42050a.netlify.com/blog/2018-07-24-from-spa-to-pwa/
* https://deploy-preview-324--gallant-hawking-42050a.netlify.com/blog/2018-11-27-open-source-maintenance/

This works relatively nice already but there are some problems with the CSS that only showed with the real content.

marcoow:
@ghislaineguerin, @pichfl: this is ready for review. It is still missing the bio for @mrmrcoleman but that should be the only thing missing. I realized we have links to author and topic pages but we don't actually have those pages yet so we can delay those until a future PR. The only drawback with this solution is when you change a post you have to restart the dev server but maybe we can improve that later…

ghislaineguerin:
@marcoow I see some styling issues but I think it's ok to merge and fix on a styling branch.

ghislaineguerin (changes requested):
I think it's better if we style hyperlinks on the baseline css.

marcoow:
Yes, inserting the real content reveals some styling issues but that was to be expected I guess.

> I think it's better if we style hyperlinks on the baseline css.

I'm not sure what you're referring to?

ghislaineguerin:
@marcoow outside of css-blocks. Otherwise we can never target hyperlinks as part of content or bold, italic text inline.

marcoow:
ah, I see - is this something that you'd expect to be done in this PR?

ghislaineguerin:
@marcoow I can take care of that on a style related PR. All else looks good.

ghislaineguerin (approved):

PR #326, opened 2019-04-05
more realistic code example

marcoow:
This makes the code example on the blog page slightly more realistic. This shows that the styles don't work if the code example gets too long: https://deploy-preview-326--gallant-hawking-42050a.netlify.com/article/

@ghislaineguerin: can you have a look?


marcoow:
this is covered by #324 actually…

PR #329, opened 2019-04-09
Content Update

ghislaineguerin:
This PR contains updates to links and contents for the existing pages except work, as we are about to change the layout.
This also introduces a new folder structure for the image assets for keeping them more organised.


pichfl (approved):
I can't comment much on the text content yet, but the structure makes sense to me.

PR #333, opened 2019-04-12
New work page layout

### LINKED ISSUE #332: New layout for Work section
As discussed with @ghislaineguerin

pichfl:
closes #332

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# http://editorconfig.org
+
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+
+[*.hbs]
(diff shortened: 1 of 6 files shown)


marcoow (commented):
LGTM but

* please let's use 2 spaces for indentation consistently - I guess our linter will catch that anyway though
* do the inline styles interfere with css-blocks somehow?
* can we have a more generic name for the component than `ShapeCase` or is it really specific to cases only?

pichfl:
- Sorry, seems the project lacks a `.editorconfig` and my defaults kicked in. Will update
- I do not use inline styles in HTML, the only inline styles are in the `<svg>` and it seems to work fine.
- Your assumption is correct, the class is specific to cases. It could be used elsewhere, but it's very unlikely. There will be others like `ShapeFullsize` (used near the footer on most pages) and alike in the future.

marcoow (approved):

PR #336, opened 2019-04-12
Link component

pichfl:
- abstracts handling of internal links
- abstracts link styling


marcoow (commented):
2 spaces indentation everywhere please 🙏 

This makes sense to me in general but I'll have to see whether I can make that work for the blog posts as well…

pichfl:
Updated an rebased. 

No need to use this in blog posts, as this is the special kind of link below blocks (with or without an arrow)

marcoow:
@pichfl this doesn't seem to build…

marcoow:
I fixed it: cf2dc97

marcoow (commented):
LGTM - only mini concern I'd have is this doesn't seem a totally generic link component but one that only works for these kind of block level links? There doesn't seem to be a single invocation of the component with `@arrow="false"` so I guess what we really want is an `<ArrowLink>` component without the extra argument?

pichfl:
Good point. We had links without the arrow at some point but since I'm touching most aspects anyway, I think it's fair to convert all of them to this format, removing the additional state. We can always expand later on.

marcoow (approved):

PR #341, opened 2019-04-15
Fix css-blocks states

marcoow:
css-blocks doesn't currently register its helpers correctly in Glimmer.js apps (it does in Ember.js apps) - see linkedin/css-blocks#195. This registers the helpers manually (based on https://github.com/linkedin/css-blocks/issues/195#issuecomment-426799214) which is pretty hacky but should be possible to remove once the underlying problem in `@css-blocks/glimmer` has been fixed.


pichfl (approved):
Seems to work as expected. Thanks for the quick fix.

PR #346, opened 2019-04-16
Split blog bundle from main app bundle

### LINKED ISSUE #273: Split blog posts in their own bundle
We should not be bundling the blog posts in the main app bundle. That would mean that we'd be loading a lot of content that is unlikely to ever be needed for everyone. Instead we should split all of the blog posts (not each one individually) into their own bundle that gets loaded when needed.

### TODO

- [x] in the package step of the build, split all of the blog bundles into their own bundle
- [x] in the blog service introduced in #272, add a method that loads that bundle in case it hasn't been loaded yet (simply by adding a `<script>` tag with the respective source)
- [x] before routing to any blog post or the blog overview (or before accessing any of the methods that return the 3 last Ember.js posts or so, the service would make sure that bundle is loaded)
- [x] in the case when the bundle is loaded on a route transition we could show one of these little line loaders on the top of the page
- [x] in the blog bundle, also add the module with the info on which components (and URLs) are related to which topic, which author etc.; this way we're only invalidating the blog bundle when writing a new post (Update: not doing yet - will do in a later PR when it's clear what the actual requirements are)

marcoow:
The blog bundle is relatively big and we shouldn't load it for everyone but only when it's needed. This PR does just that - it splits all of the blog components and templates into a separate `blog.js` and only loads that when required.

### TODO

- [x] split each blog post into its own bundle and load only that when going to the particular page
- [x] when starting the app on the `/blog` route in dev mode, there's an error
- [x] the blog bundles don't get fingerprinted correctly in the source code in production so that the requests go to the non-fingerprinted assets
- [x] show a loader while the bundle is lazily loading
- [x] the pre-rendered pages do not include the lazily inserted script tags

closes #273

### DIFF
--- a/src/ui/components/Simplabs/component.ts
+++ b/src/ui/components/Simplabs/component.ts
@@ -50,7 +50,7 @@ export default class Simplabs extends Component {
       if (bundle && !this.appState.isSSR) {
         options.before = async (done) => {
           await this._loadBundle(bundle, parentBundle);
-          this._registerContent(bundle);
+          this._registerBundle(bundle);
           done();
         };
       }
@@ -101,11 +101,8 @@ export default class Simplabs extends Component {
     });
   }
 
-  private _registerContent(bundle) {
-    let content = window[bundle.module] || {};
-    Object.keys(content).forEach((key) => {
-      this.lazyRegistration.register(key, content[key]);
-    });
+  private _registerBundle(bundle) {
+    this.lazyRegistration.registerBundle(bundle.module);
   }
 
   private _startLoader() {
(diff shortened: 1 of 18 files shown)


src/ui/components/Simplabs/component.ts
```diff
@@ -36,22 +45,83 @@ export default class Simplabs extends Component {
...
     });
     this.router.resolve(this.appState.route);
   }
 
   private _bindInternalLinks() {
     if (!this.appState.isSSR) {
-      document.addEventListener('click', (event: Event) => {
-        const target = event.target as HTMLElement;
-      
+      document.addEventListener('click', async (event: Event) => {
```
pichfl: Is there a reason this is `async`?

pichfl (commented):
I'm not a specialist on the whole bundling and Glimmer specific stuff, but I've tested locally and it worked as expected.

src/ui/components/Simplabs/component.ts
marcoow: No, I'll remove that ;)

pichfl (approved):

PR #347, opened 2019-04-17
Refactor footer

### LINKED ISSUE #343: Footer: Refactor
The footer needs better semantics, ideally a page should have only one or two navigations and we should use proper headings and lists for the links.

This should also make it easier to style the footer for mobile.

pichfl:
- drop extra containers
- add better semantics
- use CSSBlocks composition
- use CSS grid

closes #343

### DIFF
--- a/src/ui/components/Footer/stylesheet.css
+++ b/src/ui/components/Footer/stylesheet.css
@@ -1,9 +1,12 @@
 @block fluid-image from "../../styles/blocks/fluid-image.block.css";
+@block screen from "../../styles/blocks/screen.block.css";
 
 :scope {
   block-name: footer;
+  composes: 'screen.container';
+
   display: flex;
-  flex-flow: row nowrap;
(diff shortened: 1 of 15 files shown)


src/ui/components/Footer/stylesheet.css
```diff
@@ -1,9 +1,12 @@
 @block fluid-image from "../../styles/blocks/fluid-image.block.css";
+@block screen from "../../styles/blocks/screen.block.css";
 
 :scope {
   block-name: footer;
+  composes: 'screen.container';
```
marcoow, line 6: Nice - this is new in css-blocks, right?

marcoow (approved):

src/ui/components/Footer/stylesheet.css
pichfl: It is 🙂 and it's a very very useful addition.

PR #348, opened 2019-04-17
SVG based shapes

### LINKED ISSUE #309: Refactor shape blocks
Current classes for the shape elements on the site do not reflect their design intent or usage.

Instead of
shape left, shape right, shape-content, shape-text, shape-center etc

We will call the elements based on the following structure:

Featured Block
Wrapper
Background
Image
Caption

image

pichfl:
This provides SVG based versions of all shapes previously created in CSS, making 

- blog header
- feature shape (used for demoing cases)
- across shape (used for expertise)
- center shape (large case study)

closes #309 

Please note that this does not remove the old shapes right away as that would have made the PR too large. Removing the old ones and special cased components will be done in a separate PR.


src/ui/components/Homepage/template.hbs
```diff
@@ -105,8 +113,25 @@
         </div>
       </div>
     </div>
+    <Shape
+      @type="across"
+      @color="#E15D44"
+      @background="#222"
+      @src="/assets/images/marco.jpg"
+      @width="2636"
+      @height="1938"
```
marcoow: Can we keep these values in the CSS?

src/ui/components/Shape/component.ts
```diff
@@ -0,0 +1,46 @@
...
+    return this.args.type === 'feature';
+  }
+
+  get isAcross() {
+    return this.args.type === 'across';
+  }
+
+  get isCenter() {
+    return this.args.type === 'center';
+  }
```
marcoow: I'm not sure these switches are a good idea tbh - if we render different blocks for each of them it's probably better to use separate components. In the end what we have here are separate components anyway, just all merged into one that then decides based on these switches which one of the wrapped components to actually render…

marcoow (commented):
I'm not sure I understand what the benefit of this is tbh…

src/ui/components/Homepage/template.hbs
pichfl: All of these are part of the content image which differs between usages.

src/ui/components/Shape/component.ts
pichfl: I'm not very happy about that either. The reason I've collapsed all of them into a single component is their very similar usage patterns and to allow referencing a single `<svg>` definition for the base shape. I think I can send that id down to the child components, so a refactor is totally an option.

pichfl:
The point of this component is to remove components like `<TrainlineCaseThumb />`, <BreetheCaseThumb />`, `<LeadingExperts />` and others, which can be expressed in a much broader way, reducing the amount of necessary styling work.

src/ui/components/Shape/component.ts
marcoow: I think we should do that - with this solution we still effectively have the separate components we just hide them. We'll also have to add more cases here as we go. I guess we should instead try and extract smaller and more generic sub-components that we'd then be able to reuse in the similar-but-not-same components that we're currently all wrapping inside this one?

src/ui/components/ShapeBase/component.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default class ShapeBase extends Component {
+  public shapeId = null;
+
+  constructor(options) {
+    super(options);
+
+    this.shapeId = Math.random()
+      .toString(36)
+      .substr(2, 5);
```
marcoow: Is there a reason why this needs to be limited to 5 characters?
```suggestion
      .substr(2);
```

src/ui/components/ShapeBlog/component.ts
```diff
@@ -0,0 +1,3 @@
+import ShapeBase from '../ShapeBase/component';
+
+export default class Blog extends ShapeBase {}
```
marcoow:
```suggestion
export default class ShapeBlog extends ShapeBase {}
```

src/ui/components/Training/template.hbs
```diff
@@ -176,13 +176,20 @@
...
+      <p class="typography.small-text">
+        Expertise
+      </p>
+      <h3 class="typography.headline-2">
+        We are Europe’s Leading Ember Experts
+      </h3>
+      <ArrowLink @href="/expertise/ember">
+        Learn More
+      </ArrowLink>
+    </ShapeAcross>
```
marcoow, line 189: Would it not still make sense to keep the `LeadingExperts` component and use the `ShapeAcross` component inside of that? Like this, we're duplicating this snippet 5 times in the codebase, and will likely duplicate it more often in the future.

src/ui/components/ShapeBase/component.ts
pichfl: Not really, I just wanted to keep them short.

src/ui/components/Training/template.hbs
pichfl: I thought about that as well. I decided against it as I wanted to steer the components to encapsulate functionality/layout over content, keeping the content inside the "routes". But I don't have much of a preference either way, so if you prefer to keep the components there, I'll restore them.

pichfl: After pressing send on the last comment I realized why I went that direction: Keeping the content inside the routes would allow us to change the content ever so slightly which can be good for SEO. But I'm no expert by any means, so I might be wrong there.

marcoow: > Keeping the content inside the routes would allow us to change the content ever so slightly which can be good for SEO.

Yeah, we **might** want that at some point but let's do that once we need it. For now I'd not add more duplication while we're trying to reduce to duplication actually ;)

marcoow:
@ghislaineguerin this wasn't quire ready actually - there were still some open discussions…

pichfl:
I think that was a misunderstanding from our discussion yesterday. I'll create a new PR with the proposed changes (bringing back the components we discussed and so on)

PR #354, opened 2019-04-23
set page title correctly

### LINKED ISSUE #316: Document titles
I don't see document titles getting updated (checked in 54bfb86a925d6cfb70116ff6ce456658d893b88b), they are stuck at `Hello Glimmer`, which don't think would be a good thing for SEO or sharing.

marcoow:
Page titles can be configured per route in `config/routes-map.js`.

closes #316

### DIFF
--- a/config/routes-map.js
+++ b/config/routes-map.js
@@ -9,6 +9,7 @@ module.exports = function() {
   let blogPostRoutes = blogPosts.reduce((acc, post) => {
     acc[`/blog/${post.queryPath}`] = {
       component: post.componentName,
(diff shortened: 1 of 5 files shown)


pichfl:
I know this comes a bit late, but is it possible to extend this to other meta tags as well? It may be helpful to have other tags (description, open graph) as well.

marcoow:
Yes, that would be possible but we could do it in a separate issue/PR now that we have the basic mechanism in place.

pichfl (approved):

PR #369, opened 2019-04-29
navigate for clicks on elements inside links

marcoow:
This fixes navigation when the user clicks on elements **inside** of a link. Currently, we only navigate correctly when the clicked element itself is a link while we actually want to navigate for clicks on elements **inside** of links as well (e.g. the image in the main navigation).


pichfl (approved):
Works for me, I was just wondering if using the array created by `Event.composedPath()`, `Event.deepPath()` and `Event.path` could be a more modern approach. The loop based solution is fine though.

marcoow:
@pichfl those new properties don't seem to be well supported yet (e.g. not at all in IE or Edge).

PR #370, opened 2019-04-29
use arrow link component consistently

marcoow:
This replace a use of a plain link tag with the arrow link component.


pichfl (approved):
Needs a quick rebase, otherwise thanks for fixing that oversight.

PR #383, opened 2019-04-29
don't navigate on modifier clicks

### LINKED ISSUE #381: Click plus modifier should not navigate
I noticed this while working on the page, trying to open a link in a separate tab by clicking with a modifier key (command in Safari or Firefox) yields unexpected results. Both browsers navigate instead of opening a second tab/window.

Ember.js solves this issue with this helper util:
https://github.com/emberjs/ember.js/blob/8b273eb04023a876dbf968a05929d8a21a8fd27b/packages/%40ember/-internals/views/lib/system/utils.js#L9-L14

marcoow:
…but let the user open the target in a new tab instead.

closes #381


pichfl (approved):
I always try to reduce nesting and keep the happy path outside of conditionals by using early returns, but this works as well.

PR #361, opened 2019-04-25
add client descriptions

marcoow:
This adds descriptions for all of the clients that we list on the work page and a brief summary of what we did for them. I didn't add the ones we have success stories for anyway as that seemed repetitive to me.

closes #312

PR #402, opened 2019-04-30
ClientsGrid component

pichfl:
Since we use the clients grid on multiple pages, I thought it could be useful to abstract the grid into a component.

Extends #361

### DIFF
--- a/src/ui/components/ClientsGrid/component.ts
+++ b/src/ui/components/ClientsGrid/component.ts
@@ -1,0 +1,136 @@
+import Component from '@glimmer/component';
+
+const allClients = [
+  {
+    id: 'qonto',
+    src: '/assets/images/logos/qonto.svg',
+    text:
+      'Qonto is the leading neobank for SMEs and freelancers in Europe. simplabs works with their web frontend team to boost their productivity, establish Ember.js best practices and ensure long term success.',
+  },
+  {
+    id: 'cardstack',
+    src: '/assets/images/logos/cardstack.svg',
+    text:
+      'Cardstack builds the experience layer for a decentralized internet. simplabs helped them completing their Card UI system and validate its abilities in various prototype projects, identifying and fixing some issues in their core framework along the way.',
+  },
+  {
+    id: 'generali',
+    src: '/assets/images/logos/generali.svg',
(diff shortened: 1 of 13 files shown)


marcoow:
I don't think we always want to show all clients though?

marcoow:
I'm also not sure Ah sorry, just saw there's an option for including and excluding. I'm not sure we actually want to show the description on all pages though? Maybe it makes more sense extracting only each grid item so that we can reassemble those?

Also the shuffling will lead to problems with SSR as the order will be different in the SSR response than what is calculated when the app rehydrates in the browser so that the DOM changes on rehydration which looks strange to the user.

pichfl:
As of now all pages required the descriptions. We could make them optional as well. I just thought it might be sensible to have the content separated from the actual html as it’s very repetitive across pages.

Re shuffling: I feared as much and we can easily drop that option but I thought it might be useful.

src/ui/components/ClientsGrid/component.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (included.length > 0) {
+      clients = clients.filter(client => included.includes(client.id));
+    }
+
+    if (excluded.length > 0) {
+      clients = clients.filter(client => !excluded.includes(client.id));
+    }
+
+    if (this.args.shuffle) {
+      shuffle(clients);
```
marcoow: As discussed, we can't actually shuffle due to different renderings then in SSR and when the app rehydrates in the client.

src/ui/components/ClientsGrid/component.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    if (this.args.limit) {
+      return clients.slice(0, parseInt(this.args.limit, 10));
+    }
+
+    return clients;
+  }
+}
```
marcoow, line 111: I think this support too many options (which aren't used anyway). I'd rename `include` to `only` and drop support for `exclude` and `limit`. I don't think we'll need those.

marcoow (commented):
Cool, this makes sense to me but let's simplify a bit.

marcoow (approved):

PR #417, opened 2019-05-09
Add Sentry

### LINKED ISSUE #401: Add Sentry
We should have Sentry enabled for the page so we are aware if this doesn't work for anyone…

marcoow:
This adds Sentry (if an env var with the DSN is present).

closes #401

### DIFF
--- a/lib/sentry/index.js
+++ b/lib/sentry/index.js
@@ -1,0 +1,33 @@
+'use strict';
+
+module.exports = {
+  name: require('./package').name,
+
+  isDevelopingAddon() {
+    return true;
+  },
+
+  contentFor(type) {
+    if (type === 'body-pre-app-bundle') {
+      let { SENTRY_DSN, SENTRY_ENV } = process.env;
+      if (SENTRY_DSN) {
+        let options = {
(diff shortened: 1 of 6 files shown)


pichfl (approved):

lib/sentry/index.js
```diff
@@ -0,0 +1,33 @@
...
+  },
+
+  contentFor(type) {
+    if (type === 'body-pre-app-bundle') {
+      let { SENTRY_DSN, SENTRY_ENV } = process.env;
+      if (SENTRY_DSN) {
+        let options = {
+          dsn: SENTRY_DSN,
+        };
+        if (SENTRY_ENV) {
```
pichfl, line 17: Wouldn't it be enough to check for both `SENTRY_DSN` and `SENTRY_ENV` at the same time and collapse these conditionals? I may be missing information here, but given that both variables need to be present for the integration to work in a meaningful way, I'm not sure if it's necessary to make the conditional that complicated.

marcoow: the idea was to not make `SENTRY_ENV` mandatory (might well be over-engineering ;) ).

PR #423, opened 2019-05-10
New simplabs about

### LINKED ISSUE #396: About Page - finishing touches
- [x] change to match design: https://share.goabstract.com/01fb1499-0be0-4125-b38c-41b5b6d12b95

pichfl:
Does not add the "Sharing" section as I don't have content for it and I think we add it any time in the future.

closes #396 

### TODO

- [x] add real text (@marcoow)
- [x] update image section according to updated design that @ghislaineguerin will prepare
- [x] replace oktoberfest image with one that covers the complete area


marcoow:
@pichfl just put some dummy content in for sharing - I'll add real text later on.

pichfl:
Ok

pichfl:
@ghislaineguerin I think we should export the image grid as a single image.
@marcoow I added placeholders for missing text.

ghislaineguerin:
@pichfl looks good. We do need an issue opened for content and images. I will update the design since we aren't using image grids elsewhere, I'd rather keep consistent.

marcoow:
@pichfl: I renamed _"About"_ to _"Why simplabs"_

pichfl:
@marcoow I rebased and removed the todo from HTML so we could merge this without the additional images and make a separate issue for them.

marcoow:
I'd rather add the images first and merge it then - this doesn't block anything else after all. @ghislaineguerin can you add the updated design to unblock this?

marcoow:
@ghislaineguerin: can you replace the oktoberfest image with one that covers the complete visible area of the shape? Otherwise this should be good to merge.

marcoow (approved):

ghislaineguerin:
@marcoow ok. I'll add that now

ghislaineguerin:
@marcoow the combining images didn't work, but I have added blurred borders to both so that they fit nicely and the important parts are visible.

marcoow:
@ghislaineguerin ok, looks good to me - can you review and merge?

marcoow:
ah nevermind, was approved already

PR #419, opened 2019-05-09
Add Legal pages

marcoow:
This adds imprint and privacy policy pages.

closes #359 

### TODO

- [x] the content should use the full width of the page but `main-block.content` defaults to `width: var(--col-3)` - not sure what the best solution is
- [x] we should de-emphasize the links in the footer as the others are more relevant - at least we should move the legal links to a new line, maybe also make them smaller/brighter

PR #425, opened 2019-05-10
Ensure legal links are always in a new grid line

### LINKED ISSUE #359: Imprint and Data privacy
We should be able to re-use content from https://github.com/simplabs/simplabs.github.io/blob/master/imprint/index.html

pichfl:
Fixed bug introduced in #419 and closes #359

### DIFF
--- a/src/ui/components/Footer/stylesheet.css
+++ b/src/ui/components/Footer/stylesheet.css
@@ -23,7 +23,7 @@
 }
 
 .legal {
-  grid-row: 2/3;
+  width: 100%;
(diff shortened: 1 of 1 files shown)


marcoow (commented):
This works in Chrome but not Safari (which likely means it won't work in e.g. IE11 at all)…

pichfl:
Upon close inspection it shouldn't work in Chrome either, I was targeting the wrong element. Pushed a fix to the fix.

marcoow (approved):

PR #426, opened 2019-05-10
add favicon

marcoow:
…so it doesn't 404


pichfl:
Naming it "favicon.ico" and excluding it from hashing makes the link unnecessary, all browsers will try to request that path.

marcoow:
ah nice, learned something - updated (it's exclude from fingerprinting by default)

pichfl (approved):

PR #436, opened 2019-05-16
Cleanup

marcoow:
The combines several cleanups:

* organize the `public` folder a bit better by moving assets into directors and removing unused assets
* removing some unused CSS blocks
* removing unused components


pichfl:
Would it be sensible to drop the "images" folder and put everything directly into assets while you're at it? I don't see any naming collisions in the future and it would keep the paths a bit shorter.

Otherwise I'm all for this.

marcoow:
`images` is standard in Ember CLI - I'd leave it as I'm not sure dropping might have implications on the build system.

ghislaineguerin (approved):

PR #442, opened 2019-05-16
Fix Trainline screenshots

### LINKED ISSUE #437: Fix screenshots of trainline app
The screenshots we are currently showing are not from the app we wrote. That app is live on the french version though: https://trainline.fr (it's not a responsive page, request it with a (simulated) mobile device to be served the mobile version). We need to replace the screenshots on all images we show for trainline.

marcoow:
This replaces the images with trainline screenshots with ones of the app that we actually wrote (as opposed to the React app that we're currently showing).

closes #437


ghislaineguerin (approved):
👍

PR #445, opened 2019-05-16
split playbook into separate bundle

marcoow:
…so we don't grow the main bundle as the playbook receives more and more content over time and also we don't invalidate the main bundle in everyone's cache each time the playbook updates.


ghislaineguerin (approved):
makes sense 👍

PR #458, opened 2019-05-20
Crop and Replace images

ghislaineguerin:
As part of fixing the sketch file for images I have replaced them on the website to find the right position. Please review.


marcoow (commented):
Just had a quick look at `public/assets/images` and it seems some of the images still need to be optimized? There are some photos that have 1MB+ and images for the calendar etc. that are pretty big.

pichfl (approved):

PR #462, opened 2019-05-20
improve scroll behavior on navigation

marcoow:
We are currently scrolling to top when the navigation is triggered instead of when it completes. The problem here is that the navigation might potentially have to wait until it can perform if a JS bundle has to be loaded which makes this strange for the user as we scroll to the top before navigation actually occured. This changes it so that we only scroll after we have transitioned to the new page.


ghislaineguerin (approved):
👍

Issue #443, opened 2019-05-16
Optimize images


marcoow:
Some of the images we're currently using are huge and need to be scaled down as well as compressed before launch. Generally, we should run **all** images through ImageOptim.

marcoow:
done via #516

Issue #486, opened 2019-05-21
Performance issue when scrolling on some pages with images in shapes


mansona:
If you try to scroll reasonably quickly on the `/services` page you will see the page freeze for a short amount of time the first time you pass the shape with the image in the background. 

Here is a demo of the behaviour: https://cl.ly/a2fb5ebbeccc

In that video, both of the scrolls have the same "velocity" but you will see the page seem to skip the first time it is scrolled past. 

This doesn't seem to be network related i.e. I don't see anything downloading an image dynamically when we scroll to that point or anything in the network tab in chrome. It also doesn't happen on all of the pages that have this particular shape 🤔 The homepage photo of @marcoow doesn't have this issue but the `/why-simplabs` photo of me and @pichfl **does** have this issue. 

I suppose this could be related to the images not being optimised 🤔 but something to watch out for once they are if indeed this is the cause.

pichfl:
This should go away once #443 is done. The images are way larger than they need to be, both in dimension (>4k) and file size (>1mb)

mansona:
@pichfl yes I figured that was probably it, although I've never seen image rendering freeze up the UI thread like that 😱 Worth checking to make sure once #443 is done 👍

pichfl:
@mansona can you check again?

mansona:
So it is still there but it's significantly less noticeable I think we can get away with closing this issue for now. 👍

PR #496, opened 2019-05-21
Scale photos to reasonable size

pichfl:
1.5x width of a 1440pt wide display is a good compromise between file size and image quality.

Re #443, #486


pichfl:
This does not yet close #443 completely as I haven't touched anything beside the photos, but it's a start.

marcoow (approved):

PR #516, opened 2019-05-22
optimize images

### LINKED ISSUE #443: Optimize images

ghislaineguerin:
closes #443


marcoow (changes requested):
I ran https://imageoptim.com/mac with max settings over everything again but some of the images can be improved further still as they are oversized for the way we use them (e.g. the EmberFest logo we use for the talks page). Can you go over all of them again and make sure the dimensions are just big enough for the usage (and run https://imageoptim.com/mac on them after changing them)?

ghislaineguerin:
@marcoow I've checked all images and I think 2400px max is good for most of the images that go on the shapes and backgrounds. Cards seem ok as well. 
btw, I use https://tinypng.com/ for optimizing as I have a paid account, I can switch to imgoptim if you think it's better or if we all want to use the same.

marcoow:
I can recommend ImageOptim - it is an app that you install so you save the upload step and it supports SVGs as well. Also I ran it over all the assets you had compressed with TinyPNG already and it was still able to reduce the file size of some of them by an additional 60%.

I just had a look - should some of the PNGs be converted to JPGs? In particular in `/work` there are some pretty big PNGs still…

marcoow (approved):

PR #520, opened 2019-05-23
Add author badge to header

ghislaineguerin:


ghislaineguerin:
@marcoow @pichfl 
I have modified the header component to allow for showing the author info to the side, I think we can expand this in the future to include some icons maybe for the services and other visual interesting stuff on the top. However, I need the excerpt to be rendered inside the content div, so now it's looking like this:
image

Could any of you help me solve this?

lib/generate-blog-components/lib/files/author/template.hbs
```diff
@@ -1,8 +1,27 @@
...
+          @{{twitter}}
+        </a>
+      </div>
+    </div>
+  </Header>
   <div class="container">
     <div class="main-block">
-      <div class="main-block.content">
+      <div class="blog-post.sidebar-post">
+      </div>
```
marcoow: Is this still needed?

lib/generate-blog-components/lib/files/post/template.hbs
```diff
@@ -1,9 +1,29 @@
...
+        </a>
+      </div>
+    </div>
+    
   </Header>
   <div class="container">
     <div class="main-block">
+      <div class="sidebar-post">
+       
+      </div>
```
marcoow: Is this still needed?

src/ui/components/Header/template.hbs
```diff
@@ -47,17 +47,24 @@
...
+      {{/if}}
+      {{#if @title}}
+        <h1 class="typography.display">
+          {{@title}}
+        </h1>
+      {{/if}}
+      {{#if @excerpt}}
+      <p class="typography.lead">
+        {{@excerpt}}
       </p>
```
marcoow: Is there any way we could make this go into the `yield`? I think that would be much cleaner actually…

marcoow (commented):
I think if there's any way we could make the excerpt go into the `yield` of the header that would be great.

Doing `{{{excerpt}}}` (triple curlies) instead of `{{excerpt}}` would probably fix the problem you're having but if we could have it in the `yield` it would be a bit simpler and we wouldn't have to introduce the concept of a pre-htmlized lead text into the `<Header>` component.

I think we should make the main content section a bit narrower btw - I feel wide text content like this is a bit harder to read?

ghislaineguerin:
@marcoow I'll try to make this work without having to add the excerpt to the header.

ghislaineguerin:
@marcoow please review. The idea I had of doing absolute positioning didn't work as it won't fit the contents in a way that aligns with all else. We must use flex, so I had to move the title and label to the page level in order to have them on the same div as the excerpt.

src/ui/styles/blocks/blog-post.block.css
```diff
@@ -18,14 +18,38 @@
...
+
+.badge-content {
+  font-size: 1.75rem;
+  width: 15rem;
+  margin-top: 1rem;
+  text-align: right;
+}
+
+.header-excerpt {
+}
```
marcoow: This doesn't seem to be used anywhere?

lib/generate-blog-components/lib/files/author/template.hbs
```diff
@@ -1,42 +1,56 @@
 <div>
-  <Header @label="Blog" @title="Posts by {{name}}" />
+  <Header @is-blog="true">
```
marcoow: Can we rename `@is-blog` to something else, maybe something like `@flex-content` or so? We don't want the `<Header>` component to know about a specific page it is used on after all. The property name should make sense in the context of the component itself where it makes the content use `display: flex;` vs. regular display - the fact that this is used is a detail that's not relevant inside of the `<Header>` component.

marcoow (commented):
Seems good to me - I can't really say much about the CSS actually though. @pichfl can you have a look at that?

pichfl (approved):

PR #526, opened 2019-05-24
move loader UI to bottom on mobile

### LINKED ISSUE #521: Loader is not visible on mobile with open menu
Because the loader has the same background color as the menu, it is not visible on mobile when the menu is open. We could move it to the bottom or use a different color…

marcoow:
This moves the loader UI to the bottom of the screen on mobile devices since it could be hidden by the menu at the top. I'm not sure this is a great way to fix this actually though…

close #521


pichfl (approved):
Yeah, I agree with this being not an ideal solution, but it should work for now.

PR #530, opened 2019-05-24
mixed Improvements

marcoow:
* add `alt` attributes to **all** images
* add `lang` to `<html>`
* add `rel="noopener"` to all external links
* use `font-display: swap` instead of `fallback`

### DIFF
--- a/lib/generate-talks-archive/lib/files/page/template.hbs
+++ b/lib/generate-talks-archive/lib/files/page/template.hbs
@@ -18,7 +18,7 @@
             <div class="main-block.content-section">
               <div class="main-block.content-section-item">
                 <div class="main-block.content-section-image">
-                  <img class="main-block.image-small" src="{{meta.image}}" />
+                  <img class="main-block.image-small" src="{{meta.image}}" alt="photo of the talk" />
                 </div>
               </div>
               <div class="main-block.content-section-item">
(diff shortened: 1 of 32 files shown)


lib/generate-talks-archive/lib/files/page/template.hbs
```diff
@@ -36,7 +36,7 @@
           {{/each}}
         </div>
         <div class="main-block.sidebar offset-vertical">
-          <Card @src="{{meta.image}}">
+          <Card @src="{{meta.image}}" @alt="{{meta.title}} logo">
```
pichfl:
```suggestion
          <Card @src="{{meta.image}}" @alt="{{meta.title}}">
```

Skip the "logo", it doesn't help text only users as that information can be gathered from context.

src/ui/components/CardPlaybook/template.hbs
```diff
@@ -1,4 +1,4 @@
-<Card @label="Playbook" @src="/assets/images/card/how-we-work.jpg" @title="How We Work">
+<Card @label="Playbook" @src="/assets/images/card/how-we-work.jpg" @title="How We Work" @alt="photo of a work meeting">
```
pichfl, line 1: A screenreader will read: "Image. Photo of a work meeting". I'd argue that nearly all images we use are pure decoration and are fine with `alt=""`

pichfl (dismissed):
`alt="photo of the author"` is not a good idea. Use `alt=""` if a text based user can't gain additional information or context from the alternative description.

PR #532, opened 2019-05-25
Stabilize crawler on CI

marcoow:
The crawler is flaky on CI - trying to stabilize it…


marcoow:
seems to be stable now

Issue #357, opened 2019-04-24
Responsive layouts


pichfl:
While I'm still not quite happy with the state of layout abstractions, I think it's time to ensure the site works well on mobile devices.

Once we know what the relevant breakpoints are we need to update the pre-renderer so it injects the critical CSS for the respective breakpoints - see #457.

### TODO

- [ ] the shapes are displaced on some mobiles (seems to be the case on iOS but also when simulating a Galaxy S5 in Chrome): 

<img width="513" alt="Bildschirmfoto 2019-05-25 um 13 02 24" src="https://user-images.githubusercontent.com/1510/58368437-66e2cf80-7eed-11e9-9c38-62034d61932e.png">

- [ ] the author badge on mobile is currently hidden and needs to be shown again and fixed so it renders correctly/reasonably on mobile
- [ ] intermediate resolutions between desktop and mobile (e.g. iPad) are currently broken and need to be addressed as well, e.g.:

![Bildschirmfoto 2019-05-25 um 13 08 04](https://user-images.githubusercontent.com/1510/58368502-423b2780-7eee-11e9-8056-7060a7c880c6.png)

- [ ] some of the shapes have excessive amounts of whitespace in some resolutions:

<img width="830" alt="Bildschirmfoto 2019-05-27 um 14 40 33" src="https://user-images.githubusercontent.com/1510/58420424-8284dc00-808d-11e9-9706-223d0ec0998c.png">

- [ ] we should restore the parallax scrolling effects for the shapes on the desktop

marcoow:
@ghislaineguerin for context: we have 90% desktop users so while we want a mobile version that works, it's not worth (for now) to spend a lot of work to make the mobile version super awesome.

ghislaineguerin:
Homepage Responsive Layout:
https://share.goabstract.com/e0d0094d-e983-4c8e-a59d-d7d4c36cdb57?mode=design&sha=90cf65596cb8369829a4e33f3bc6ecc12d12dca6

I will add more pages as I complete them.

marcoow:
I added a list of TODOs at the top and am moving this as the top prio to the post-launch fixes.

marcoow:
setting this as blocked as it depends on #569, #570, #571, #573

PR #554, opened 2019-05-27
Change breakpoint to 888px

pichfl:
Should provide a quick fix for 

> intermediate resolutions between desktop and mobile (e.g. iPad) are currently broken and need to be addressed as well

Re #357

### DIFF
--- a/src/ui/components/CookieBanner/stylesheet.css
+++ b/src/ui/components/CookieBanner/stylesheet.css
@@ -14,7 +14,7 @@
   overflow: hidden;
 }
 
-@media (max-width: 699px) {
+@media (max-width: 887px) {
   :scope {
     --offset-horizontal: 2.5rem;
   }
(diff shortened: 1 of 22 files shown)


marcoow:
I was thinking we use a breakpoint like this for the shapes only and leave the content displaying the way it does for desktop until we reach the mobile breakpoint? Wouldn't that be a quick but proper solution or do you think we need to do more than that?

pichfl:
The content gets very narrow and loses it's proportion when going below that width, so this makes for a better option design wise and I think it's reasonable well readable even on the iPad.

marcoow (approved):
This doesn't fix all of the issues but should be an improvement at least for now: 
Bildschirmfoto 2019-05-27 um 14 26 44

pichfl:
I'm working on those in a separate branch.

PR #557, opened 2019-05-27
redirect old training page to turoring

### LINKED ISSUE #555: Setup redirect for /training to go to the tutoring page instead
The /training page has a lot of SEO value and it is now going to a 404 page

marcoow:
…so we don't lose all of the traffic which would be pretty bad of course 🤦‍♂ 

closes #555


pichfl (commented):
👍

PR #558, opened 2019-05-27
replace marcos photo

### LINKED ISSUE #538: Fix Marco's EmberFest photo
The photo of myself giving a talk at EmberFest has a black right border so while it actually covers the complete background of the respective shape it ends well before the right edge of the shape because that black border is in the image:

Bildschirmfoto 2019-05-26 um 12 53 52

ghislaineguerin:
closes #538 

I replaced the photo with another screenshot that eliminates the black border situation


marcoow (approved):
Sure, that works.

Do we have any good photos of other people giving talks as well though? I don't think it should necessarily be me on the front page.

ghislaineguerin:
@marcoow I could screenshot other simplabs members talks. Should we do that as part of this card though?

marcoow:
let's stick with this if we don't have any other pictures yet - we should just change it from time to time if we get any new pictures. Maybe there will be a nice one of Jessica at JSConf.

PR #560, opened 2019-05-27
Display author badge for blog posts on narrow screens

pichfl:
As this is not in the screens by @ghislaineguerin I tried to make it work anyway. 

Re #357

### DIFF
--- a/lib/generate-blog/lib/files/post/template.hbs
+++ b/lib/generate-blog/lib/files/post/template.hbs
@@ -1,5 +1,14 @@
 <div>
   <Header @flex-content="true">
+    <div class="blog-post.excerpt">
+      <p class="typography.small-text">
+        Blog
+      </p>
+      <h1 class="typography.display">
+        {{title}}
+      </h1>
+      {{{excerpt}}}
+    </div>
     <div class="blog-post.badge">
       <div class="blog-post.badge-image">
         <img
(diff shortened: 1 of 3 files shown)


marcoow (commented):
I think this is strange as now the author is between the first and second paragraph of the content. I guess on mobile we should ideally have the author below the article.

pichfl:
Putting it below the main text requires a larger refactoring of the current DOM.

But: 
It was a deliberate choice. The lead paragraph provides information about what's to expect and putting the author between that and the main content highlights their voice which shows the strength of our team.

marcoow:
Yeah, I see what you mean but I don't think it works well really:

Bildschirmfoto 2019-05-27 um 15 53 23

ghislaineguerin:
@pichfl I will create a mockup for this if that's easier. I do agree with your point of showing author before content.

ghislaineguerin (approved):

ghislaineguerin:
@pichfl the change looks good. there's a conflict that needs to be solved.

marcoow (changes requested):
Let's not merge it like this. We need to rethink it properly and come up with a design first. The author showing between the intro and the rest of the article seems like a bug.

pichfl:
I'm closing this in favor of a future PR which will come in once @ghislaineguerin has time to create a proper design for this.

PR #565, opened 2019-05-27
Never cache service worker

marcoow:
#562 didn't actually work - here's another try…


marcoow:
This works as expected now

ghislaineguerin (approved):

PR #568, opened 2019-05-28
add service teaser and refactor content

### LINKED ISSUE #552: Teaser our services on the Ember page
The purpose of the Ember Expertise page is to explain to visitors why we are the best choice if they need help with their Ember.js projects and why we have the experience and expertise that we have. As potential customers reach that page, the next step we want them to make is to either contact us or at least be interested in how we can help them. The first step for that would be to go to the _"Team Augmentation"_ or _"Tutoring"_ pages to learn more about these offerings. We should have some teasers for these pages on the Ember page so it's easy for the visitors to go there. I think we can have these teasers towards the bottom of the page as I'd expect people to look at the content of the Ember page first and then when they reach the bottom we should give them an easy way to make the next step.

### TODO

- [ ] design the teasers
- [ ] add the teasers to the Ember page (HTML/CSS)

ghislaineguerin:
closes #552 

In order to implement the teaser I had to refactor the content class names and the way we style them to prevent adding non reusable classes.


marcoow (commented):
Are all the changes that seem unrelated to the services teaser on the Ember page actually necessary for it? If not, can you move them to a separate PR?

Including all these changes makes it much harder to review. I just saw that this broke on the Ember page:

Bildschirmfoto 2019-05-28 um 13 05 35

and I'd have to check the rest of the page still to know whether more breaks there…

marcoow:
Ah sorry - I realize this refactors to new content classes to support that new layout. Can we come up with a more contained change for now though @ghislaineguerin? Maybe a class that's specific to that one page? I think @pichfl wanted to refactor the way the classes that define the general content structure anyway?

ghislaineguerin:
@marcoow I understand, I will fix that particular issue. But trying to solve this without working out the content classes issue would require me to create new classes that introduce more problems... not sure how I could isolate that from the task. I will push a new commit to fix the image margin issue.

marcoow:
Yeah, but the new classes would only be used on that one page so they could only lead to new problems on that one page - if @pichfl works on a refactoring anyway the new classes introduced by this might be redundant anyway. Can you chime in @pichfl ?

ghislaineguerin:
@marcoow I wouldn't be aware of a future refactoring, which I think is good, I think a lot of the layouts we create could be further simplified. I think we need to make the refactoring plans more visible then.

pichfl:
I think this is a good start for the refactoring and won't stand in the way of what we've planned.

pichfl (approved):

PR #574, opened 2019-05-29
handle global CSS in in-repo-addon

marcoow:
This improves the way we handle global styles that we keep outside of the component-specific css-blocks styles. Instead of statically prefixing `app.css` with the content of some files that are then not observed and changes to them would not trigger a rebuild, we move this into an in-repo-addon that allows:

* having one central `global.css` file that `@import`s everything it needs - its contents will be prepended to the `app.css` generated by css-blocks
* changing, adding and removing files will trigger a rebuild
* in development mode, we leave the `@import`s intact for easier debugging
* in production mode, we merge all CSS files into one for fewer files and fewer server roundtrips


pichfl (approved):
Neat.

PR #579, opened 2019-05-31
baseline.css

### LINKED ISSUE #569: Improve baseline CSS styles
We currently only have very rudimentary baseline and reset styles so that we have to style everything with css blocks which can be tedious and also requires assigning classes to all elements, e.g. blog post that are generated from markdown. It also makes things in css-blocks more complex as we have to handle all of the baseline stuff there.

### TODO

- [x] move custom properties from css-blocks into the baseline so we don't have to extend `screen.block.css` everywhere
- [x] extract default styles for text, links etc. from the blocks and keep them in the baseline css
- [x] remove class assignments from the markdown handling code where the respective styles have been replaced with classes in the baseline

pichfl:
Moving basic typographic styles out of CSS Blocks removes a lot of noise from the HTML of components and pages like the blog posts. 

Closes #569


lib/generate-blog/lib/templates/author/stylesheet.hbs
```diff
@@ -3,6 +3,8 @@
 @block blog-post from "../../styles/blocks/blog-post.block.css";
 @block blog-author from "../../styles/blocks/blog-post.block.css";
 @block fluid-image from "../../styles/blocks/fluid-image.block.css";
+@block container from "../../styles/blocks/container.block.css";
+@block o from "../../styles/blocks/offset.block.css";
```
marcoow: Let's name things properly please - seeing something like `o.class` in the template basically makes it impossible to understand what that is about.

src/ui/components/FormContact/component.ts
```diff
@@ -38,7 +38,6 @@ export default class FormContact extends Component {
       form.reset();
 
       this.formState = FormState.Waiting;
-      this.errors = {};
```
marcoow: Thy is this needed? Also seems unrelated?

marcoow (commented):
This leads to a few visual changes which I'm not sure we actually want, e.g.

* imprint and privacy links now are blue on hover
* the buttons now show an underline under the text on hover
* the links and button in the cookie teaser are now blue
* the header links in the footer now show an underline on hover
* email, phone number etc. on the contact page are now indented
* header links on services page now show underlines on hover

…and potentially others - have only looked on the home page for now

src/ui/components/FormContact/component.ts
pichfl: That was not intended for this PR, but it is no longer required and generates a warning in the build process. This was most likely an artifact from my attempt to do the validations manually.

lib/generate-blog/lib/templates/author/stylesheet.hbs
pichfl: I took inspiration for that from the CSS Blocks fixtures, but I agree with preferring speaking names in general, will update.

src/ui/components/FormContact/component.ts
marcoow: let's keep separate things separate though 🙏

pichfl:
> * imprint and privacy links now are blue on hover

Intended change

> * the buttons now show an underline under the text on hover

Bug, will update

> * the links and button in the cookie teaser are now blue

Unintended change, will update.

> * the header links in the footer now show an underline on hover

Intended change.

> * email, phone number etc. on the contact page are now indented

Oversight, will update.

> * header links on services page now show underlines on hover

Intended change.
 
> …and potentially others - have only looked on the home page for now

Thanks for all of these, it's hard to see that stuff when you're reiterating that often.

src/ui/components/FormContact/component.ts
pichfl: Done. See #584

src/ui/styles/blocks/list.block.css
```diff
@@ -6,8 +6,6 @@
 }
 
 .item {
-  list-style-type: disc;
-  margin-bottom: 0.5rem;
 }
```
marcoow: Do we still need the class if it is now empty anyway?

src/ui/components/ShapeAcross/stylesheet.css
```diff
@@ -1,4 +1,5 @@
-@block screen from "../../styles/blocks/screen.block.css";
+@block containerBlock from "../../styles/blocks/container.block.css";
```
marcoow, line 1: I'd name this just `container` everywhere for consistency.

marcoow (changes requested):
I found some more things that now show the wrong way (all Safari):

* text color is wrong in case studies:

Bildschirmfoto 2019-05-31 um 14 59 45
Bildschirmfoto 2019-05-31 um 14 59 36

* missing space after paragraph in form submission confirmation text:

Bildschirmfoto 2019-05-31 um 15 00 17

* mobile menu items have wrong color and are displaced when hovering:

Bildschirmfoto 2019-05-31 um 15 00 35
Bildschirmfoto 2019-05-31 um 15 00 39

* text underline is strangely broken on mobile blog index when hovering:

Bildschirmfoto 2019-05-31 um 15 01 33

src/ui/styles/blocks/list.block.css
pichfl: I'm hunting down empty ones, good catch.

src/ui/components/ShapeAcross/stylesheet.css
pichfl: Sadly no, there is a class "container" in the local scope, so I had to rename. This will most likely go away when the responsive shape refactoring is done.

ghislaineguerin (approved):
👍 excellent, thanks!

Issue #571, opened 2019-05-29
Separate CSS Block for page layout


marcoow:
We currently have various classes for defining the page's layout structure that also require a number of extra wrappers that would not normally be necessary. That makes everything pretty complex and also makes building the responsive version of the page correctly. Instead, we'll reduce these classes to one layout block with a number of helper classes that can be used throughout the project.

### TODO

- [x] identify and define layut classes in the new layout block
- [x] apply the new classes to the existing page (which will allow us to remove various wrapper elements as well then)
- [x] remove the layout classes in `main-block.block.scss` and potentially the block altogether

PR #586, opened 2019-06-05
Grid layout

### LINKED ISSUE #570: Separate CSS Block for spacing
We currently have various classes for defining spacing and offsets between elements that also require a number of extra wrappers that would not normally be necessary. That makes everything pretty complex and also makes building the responsive version of the page correctly. Instead, we'll reduce these classes to one spacing block with a number of helper classes that can be used throughout the project.

### TODO

- [x] identify and define spacing classes in the new spacing block
- [x] apply the new classes to the existing page
- [x] remove the spacing classes in `main-block.block.scss`

pichfl:
Removes deeply nested DOM in favor of a much slimmer CSS grid based layout. 

Closes #570, #571


ghislaineguerin (approved):
👍

PR #590, opened 2019-06-06
Disable serviceworker in development mode

pichfl:
I got stumped by the service worker on iOS more than once so I propose we disable it for good while in development. This change ensures it's only added to the DOM if the environment is set to production.


ghislaineguerin:
👍 this is helpful, thanks

ghislaineguerin (approved):

PR #594, opened 2019-06-11
Cleanup button block

pichfl:
Use the `:scope` directly. Reduces clutter in templates and makes composition easier, as we no longer compose a subclass of a scope.


ghislaineguerin (approved):
👍

PR #596, opened 2019-06-17
Improved shapes

ghislaineguerin:


ghislaineguerin:
Update:

A white frame and border are now used to enhance mobile view. This will simplify the responsiveness of most shapes.

As you can see I still have to figure out some issues like rotation of the placeholder. but the solution seems to work fine for all viewport sizes.

localhost_4200_ (1)
localhost_4200_

marcoow:
> A white frame and border are now used to enhance mobile view. This will simplify the responsiveness of most shapes.

Hm, is this what you agreed on? I'm not a huge fan tbh - I think it should be possible to implement the design we have on abstract.com?

ghislaineguerin:
@marcoow I've added the white frame to simplify the implementation. Florian was away. I have now fixed most issues so I think I can focus on mobile. I will fix it to look like the abstract version.

marcoow:
Hm, I'd say let's try and simplify the implementation without changing the design (unless that turns out to be impossible for some reason of course). Do you think that makes sense?

ghislaineguerin:
@marcoow Sure, I can do that.

ghislaineguerin:
@pichfl looks great, mobile it's also working fine for me.

marcoow (changes requested):
I found some issues on some pages:

* Services -> Full Stack Engineering:

Bildschirmfoto 2019-07-04 um 14 41 20

* Elixir & Phoenix:

Bildschirmfoto 2019-07-04 um 14 44 26

* pages are too wide on mobile now - don't know whether that's because of the shapges:

Bildschirmfoto 2019-07-04 um 14 47 26

* trainline shape on work page doesn't work on mobile:

Bildschirmfoto 2019-07-04 um 14 48 53

* not sure Timify shape is broken on mobile or as intended?

Bildschirmfoto 2019-07-04 um 14 49 42

* Timify shape in the success story is broken:

Bildschirmfoto 2019-07-04 um 14 50 12

* same in the trainline case:

Bildschirmfoto 2019-07-04 um 14 50 44

src/ui/components/PageHomepage/stylesheet.css
```diff
@@ -4,6 +4,7 @@
 @block offset from "../../styles/blocks/offset.block.css";
 @block typography from "../../styles/blocks/typography.block.css";
 
+
```
marcoow: 👀

src/ui/components/ShapeAcross/component.ts
```diff
@@ -1,7 +1,23 @@
 import ShapeBase from '../ShapeBase/component';
 
 export default class ShapeAcross extends ShapeBase {
+  public maskId = null;
```
marcoow, line 4: Should this not be private or protected?

src/ui/components/ShapeAcross/component.ts
```diff
@@ -1,7 +1,23 @@
...
+
+  constructor(options) {
+    super(options);
+
+    this.maskId = Math.random()
+      .toString(36)
+      .substr(2);
```
marcoow, line 11: This is problematic as it will be different in the SSR pre-render and the actual render in the browser. Can we have something that is stable?

src/ui/components/ShapeBold/component.ts
```diff
@@ -0,0 +1,21 @@
+import ShapeBase from '../ShapeBase/component';
+
+export default class ShapeBold extends ShapeBase {
+  public maskId = Math.random()
+    .toString(36)
+    .substr(2);
```
marcoow, line 6: see above - we shouldn't have anything random in the DOM

src/ui/components/ShapeBold/component.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    if (this.args.backgroundA) {
+      style.set('--shape-color-a', this.args.backgroundA);
+    }
+
+    if (this.args.backgroundB) {
+      style.set('--shape-color-b', this.args.backgroundB);
+    }
+
+    return [...style.entries()].map(([key, value]) => `${key}:${value}`).join(';');
```
marcoow: I'm not sure why we need `[...style.entries()]` here as we **know** `style.entries()` is going to be unique (`'--shape-color-a'`, `'--shape-color-b'`)?

src/ui/components/ShapeFeature/component.ts
```diff
@@ -18,4 +18,14 @@ export default class ShapeFeature extends ShapeBase {
...
+
+  get bottomImageStyle() {
+    let style = '';
+
+    if (this.args.image) {
+      style += `background-image: url(${this.args.image})`;
+    }
+
+    return style;
+  }
```
marcoow, line 26: What about

```typescript
get bottomImageStyle() {
  if (this.args.image) {
    return `background-image: url(${this.args.image})`;
  }
}
```

?

src/ui/components/ShapeFeature/stylesheet.css
```diff
@@ -4,19 +4,33 @@
...
-
   --shape-color: var(--color-accent);
+  transform: rotate(30deg);
+}
+
+:scope[state|hasImage] {
+}
+
+:scope[state|hasVideo] .video {
+}
```
marcoow: are these empty blocks needed?

src/ui/styles/blocks/offset.block.css
```diff
@@ -55,3 +55,5 @@
 .after-21 {
   margin-bottom: 21rem;
 }
+
+
```
marcoow: 👀

marcoow (commented):
Also, the changed images are all bigger than before and some new ones are huge - I think they all need compression.

src/ui/components/PageHomepage/stylesheet.css
pichfl: Took me until now to realize that prettier isn't run for CSS files, added PR #606 for that. Will rebase once merged.

src/ui/components/ShapeAcross/component.ts
pichfl: No reason not to do it.

src/ui/components/ShapeAcross/component.ts
pichfl: This is only the latest implement of that code, it's used all across the components. 
Sadly I can not use a static string for those ids as they would clash if a component is reused twice on a page, which can and does happen.

It's used in `ShapeBase` and any Shape* class that has inline SVG references. 

I was hoping that a changing id wouldn't actually have any impact on the pre-render as they are used only locally in that component, which in turn makes a strong argument for changing those fields to private.

src/ui/components/ShapeBold/component.ts
pichfl: Overengineering at it's finest. Will replace with a less computationally offensive solution. There is absolutely no need for the `Map` in the first place.

Sidennote: The `[...style.entries()].map` call is not used for uniqueness, but as the only way to map the entries inside a Map, since the Map object does not support `.map()` itself.

src/ui/components/ShapeFeature/component.ts
pichfl: Tried to keep the method signature similar to the code above, I'm not attached to that bit, will refactor.

src/ui/components/ShapeAcross/component.ts
marcoow: not sure what you mean

src/ui/components/ShapeAcross/component.ts
marcoow: Yeah, only recognized that now - we should have a separate issue to handle this. Can you create one?

src/ui/components/ShapeAcross/component.ts
pichfl: Sorry. Meant to say: Will change to private and update similar props in other components as well.

src/ui/components/ShapeAcross/component.ts
pichfl: Done, see #608

src/ui/components/ShapeAcross/component.ts
pichfl: Actually, if I make these private, I get warnings in TSLint about props that are never read, as the linter doesn't know about the templates. Should I still do that and add a comment to disable that warning?

marcoow: I'd say let's just not specify `private`/`public` at all.

pichfl: It gets inserted by the formatter task automatically

src/ui/components/ShapeAcross/component.ts
```diff
@@ -1,7 +1,23 @@
...
+
   get backgroundStyle() {
-    return this.args.background && `--shape-color: ${this.args.background}`;
+    let style = '';
+
+    if (this.args.background) {
+      style += `--shape-color: ${this.args.background}`;
+    }
+
+    return style;
```
marcoow: I'd simplify the method to

```es6
get backgroundStyle() {
  if (this.args.background) {
    return `--shape-color: ${this.args.background}`;
  }
}
```

marcoow (commented):
This looks very good in general but the content still seems a bit too wide (tested in iPhone 6 simulator but same on my iPhone XS): 

Bildschirmfoto 2019-07-08 um 18 05 46

Also the images are all growing in size significantly which is probably wrong. Have a look at https://imageoptim.com/mac for a good compression tool.

pichfl:
Everything is compressed with ImageOption and where possible with ImageAlpha as well. For the overlapping images we have to rely on pngs (at least for stable solution) so while unfortunate, there is no other option here 😞

marcoow:
@pichfl can we check we can shrink the image dimensions a bit maybe? 350KB is pretty huge…

marcoow (approved):
🎉

PR #606, opened 2019-07-04
Format CSS files using Prettier

pichfl:
Removes inconsistencies across CSS files.


marcoow (commented):
Makes sense 👍 

Can we have a lint job on CI for the CSS as well please?

marcoow (approved):

Issue #607, opened 2019-07-04
Page header margins are too large on narrow screens


pichfl:
- [x] define smaller size in design
- [x] implement

Example:
![](https://user-images.githubusercontent.com/845767/60675206-165f7a00-9e7c-11e9-8f72-70757cd5ec54.jpeg)

PR #611, opened 2019-07-09
Navigation active state

### LINKED ISSUE #572: Navigation active states
We do not currently have an active state for the navigation so while we have a hover state, users don't know which page they're currently at.

### TODO

- [x] design an active state for the top navigation items
- [x] define a property for setting the active item on the `<Header>` component
- [x] set that new property in every page component where the `<Header>` component is invoked (also in generated page components like the blog)
- [x] add active state to mobile menu (white underline)

pichfl:
This enables the active state in the main navigation on larger screens. 

Closes #572, #607

### DIFF
--- a/lib/generate-calendar/lib/templates/page/template.hbs
+++ b/lib/generate-calendar/lib/templates/page/template.hbs
@@ -1,5 +1,5 @@
 <div>
-  <Header @title="Calendar">
+  <Header @active="about" @title="Calendar">
     <p class="typography.lead">
       Find out about upcoming events, conferences and meetups we will be attending or organizing.
     </p>
--- a/lib/generate-calendar/lib/templates/page/template.hbs
+++ b/lib/generate-calendar/lib/templates/page/template.hbs
@@ -1,5 +1,5 @@
 <div>
-  <Header @active="about" @title="Calendar">
+  <Header @title="Calendar">
     <p class="typography.lead">
       Find out about upcoming events, conferences and meetups we will be attending or organizing.
     </p>
--- a/lib/generate-talks-archive/lib/templates/page/template.hbs
+++ b/lib/generate-talks-archive/lib/templates/page/template.hbs
@@ -1,5 +1,5 @@
 <div>
-  <Header @active="about" @title="Talks">
+  <Header @title="Talks">
     <p class="typography.lead">
       We strongly believe in the value of sharing our expertise and experience with others. Here are a collection of talks that members of our team have given at various conferences all over the world in the past years.
     </p>
--- a/lib/generate-talks-archive/lib/templates/page/template.hbs
+++ b/lib/generate-talks-archive/lib/templates/page/template.hbs
@@ -1,5 +1,5 @@
 <div>
-  <Header @title="Talks">
+  <Header @active="about" @title="Talks">
     <p class="typography.lead">
       We strongly believe in the value of sharing our expertise and experience with others. Here are a collection of talks that members of our team have given at various conferences all over the world in the past years.
     </p>
--- a/src/ui/components/Header/template.hbs
+++ b/src/ui/components/Header/template.hbs
@@ -48,13 +48,13 @@
     </div>
   </div>
   <div class="content" state:flex-content={{@flex-content}}>
-    {{#if @label}}
-      <p class="typography.small">
-        {{@label}}
-      </p>
-    {{/if}}
     {{#if @title}}
       <h1 class="typography.display">
+        {{#if @label}}
+          <small class="typography.small">
+            {{@label}}<span class="typography.hidden">:</span>
+          </small>
+        {{/if}}
         {{@title}}
       </h1>
     {{/if}}
--- a/src/ui/components/Header/template.hbs
+++ b/src/ui/components/Header/template.hbs
@@ -29,19 +29,19 @@
         </label>
       </div>
       <nav class="nav">
-        <a href="/services" class="nav-item" state:active={{eq @active 'services' 'active'}} data-internal>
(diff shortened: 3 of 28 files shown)


marcoow (commented):
Can we not compare the links against the current URL to find which one should be active?

pichfl:
While that works for most, it won't work for all pages as cases are under `/cases` but the overview page is under `/work`. Same goes for the `/expertise` links. 

I don't think it's worth to find a mapping scheme, adding the active state to the each usage of the header component was only 5 minutes of work and is a few characters per new page.

src/ui/components/PageCaseDdWrt/template.hbs
```diff
@@ -1,5 +1,5 @@
 <div>
-  <Header @title="A modern UI for an open-source router firmware" />
+  <Header @active="work" @label="Case study for embeDD" @title="A modern UI for an open-source router firmware" />
```
marcoow: I think this should be one of

* _"Case Study"_
* _"Expeditoin Case Study"_
* _"Case Study: Expedition"_

– I'd prefer the last. The same applies for the other case studies of course.

lib/generate-talks-archive/lib/templates/page/template.hbs
```diff
@@ -1,5 +1,5 @@
 <div>
-  <Header @title="Talks">
+  <Header @active="about" @title="Talks">
```
marcoow: Is this correct for the talks page? I guess none of the top menu items should be active in this case?

lib/generate-calendar/lib/templates/page/template.hbs
```diff
@@ -1,5 +1,5 @@
 <div>
-  <Header @title="Calendar">
+  <Header @active="about" @title="Calendar">
```
marcoow: Is this correct for the calendar page? I guess none of the top menu items should be active in this case?

src/ui/components/Header/template.hbs
```diff
@@ -29,32 +29,32 @@
...
-        <a href="/why-simplabs" class="nav-item" data-internal>
+        <a href="/why-simplabs" class="nav-item" state:active={{eq @active 'about' 'active'}} data-internal>
           Why simplabs
         </a>
-        <a href="/blog" class="nav-item" data-internal>
+        <a href="/blog" class="nav-item" state:active={{eq @active 'blog' 'active'}} data-internal>
           Blog
         </a>
-        <a href="/contact" class="work-with-us" data-internal>
+        <a href="/contact" class="work-with-us" state:active={{eq @active 'contact' 'active'}} data-internal>
```
marcoow: I think `eq` is a bit misleading as it combines an equality check and an if/else - I think if `eq` does only what its name implies, we could still combine that with the predefined `if` helper:

```hbs
state:active={{if (eq @active 'contact') 'active'}}
```

src/ui/components/PageElixirExpertise/template.hbs
```diff
@@ -1,5 +1,5 @@
 <div>
-  <Header @title="The best of all worlds">
+  <Header @active="services" @title="The best of all worlds">
```
marcoow: Is this correct? I think the expertise pages don't really belong to any of the top level navigation items?

src/ui/components/PageEmberExpertise/template.hbs
```diff
@@ -1,5 +1,5 @@
 <div>
-  <Header @title="Europe's leading Ember experts">
+  <Header @active="services" @title="Europe's leading Ember experts">
```
marcoow: Is this correct? I think the expertise pages don't really belong to any of the top level navigation items?

src/ui/components/PagePlaybook/template.hbs
```diff
@@ -1,5 +1,5 @@
 <div>
-  <Header @title="Playbook">
+  <Header @active="about" @title="Playbook">
```
marcoow: Is this misleading? I'm not sure one can actually navigate to the playbook from the _"Why simplabs"_ page?

src/ui/components/PageCaseDdWrt/template.hbs
pichfl: Doing the last one gives this to a screen reader:

- "Case Study: embeDD: A modern UI for an open-source router firmware"

The labels are actually prefixes to the main headline of the page. If you don't like the "for" in the labels, I'd go with just "Case Study".

lib/generate-talks-archive/lib/templates/page/template.hbs
pichfl: The talks page was linked from there so I think it might qualify as being nested. I do know it's linked elsewhere as well, but that's what I picked, so removing that is fine, too.

lib/generate-calendar/lib/templates/page/template.hbs
pichfl: See above, will remove.

src/ui/components/PageElixirExpertise/template.hbs
pichfl: I saw both expertise pages as an extension of our services, will change.

src/ui/components/PageCaseDdWrt/template.hbs
marcoow: Should we use _"Expedition Case Study"_ then?

pichfl: Works as well 👍 Updating…

ghislaineguerin (approved):

PR #639, opened 2019-07-22
simplabs monthly update - July 2019

mrmrcoleman:

Issue #561, opened 2019-05-27
Improve layout for blog posts


marcoow:
The current layout for blog posts still lacks refinement. 

### TODO

- [x] Responsiveness isn't used everywhere
- [x] Autor card is missing on mobile
- [x] New design
- [x] Update layout code
- [x] Improve table typography
- [x] Images
- [x] embedded contents are not handled properly
- [x] ~Title images should be used more prominently~ → see #438

### Steps

- Create new layout for both the narrow and wide screen layout with samples for possible content types
- Implement that
- Prepare and configure images for most popular existing posts (we can just have a look at GA for identifying those)

marcoow:
This might be obsolete once we have #573, #574

marcoow:
I realized also that tables don't render nicely, see e.g. https://simplabs.com/blog/2017/12/04/enginification

PR #641, opened 2019-07-23
Wrap blog post images in figures and limit them to the width of the content

pichfl:
This is not a final solution for the blog layout issues described in #561, but it allows us to release #639

### DIFF
--- a/lib/generate-blog/lib/components-builder.js
+++ b/lib/generate-blog/lib/components-builder.js
@@ -165,27 +165,35 @@ function encodeCurlies(content) {
 
 function htmlizeBody(content) {
   let html = htmlizeMarkdown(content, { renderer: Renderer }, dom => {
+    let replaceWithFigure = (element, content) => {
+      let figure = dom.createElement('figure');
+
+      figure.classList.add('figure');
+      figure.appendChild(content);
+
+      element.parentElement.insertBefore(figure, element);
+      element.parentElement.removeChild(element);
+    };
+
     dom.querySelectorAll('p > img').forEach(img => {
       let paragraph = img.parentElement;
+
       img.classList.add('figure.content-centered');
-      let figure = dom.createElement('figure');
-      figure.classList.add('figure');
-      figure.appendChild(img);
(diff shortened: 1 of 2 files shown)


lib/generate-blog/lib/components-builder.js
```diff
@@ -168,10 +168,20 @@ function htmlizeBody(content) {
...
+
+    dom.querySelectorAll('p > a > img').forEach(img => {
+      let paragraph = img.parentElement.parentElement;
+      img.classList.add('figure.content-centered');
+      let figure = dom.createElement('figure');
+      figure.classList.add('figure');
+      figure.appendChild(img.parentElement);
+      paragraph.parentElement.insertBefore(figure, paragraph);
       paragraph.parentElement.removeChild(paragraph);
     });
```
marcoow, line 192: Can we remove duplication here with something like this?

```js
let replaceWithFigure = (element, content) => {
  let figure = dom.createElement('figure');
  figure.classList.add('figure');
  figure.appendChild(content);
  element.parentElement.insertBefore(figure, element);
  element.parentElement.removeChild(element);
}

dom.querySelectorAll('p > img').forEach(img => {
  let paragraph = img.parentElement;
  img.classList.add('figure.content-centered');
  replaceWithFigure(paragraph, img);
});

dom.querySelectorAll('p > a > img').forEach(img => {
  let paragraph = img.parentElement.parentElement;
  img.classList.add('figure.content-centered');
  replaceWithFigure(paragraph, img.parentElement);
});
```

marcoow (changes requested):

lib/generate-blog/lib/components-builder.js
pichfl: Sure, I just put it as a quick fix to be improved upon when working on #561, but I can update.

marcoow: These quick fixes generally have a tendency to never get cleaned up afterwards ;)

marcoow (approved):

PR #647, opened 2019-07-25
prevent replacement of `app.js` and `app.css` with fingerprinted filenames

### LINKED ISSUE #630: Using `app/app.js` in a blog post gets replaced by broccoli-asset-rev
Turns out `broccoli-asset-rev` is really eager to replace `app/app.js`, which leads to a false positive inside our blog, more precisely inside the blog post about Sentry.

There is the `ignore` array which we could add here, where we could ignore the blog posts https://github.com/simplabs/simplabs.github.io/blob/a09082456ba63652ca9aafcf507e87b267bb4b92/ember-cli-build.js#L229-L233

I'm just not entirely sure if that might have side effects. Another option though much sneakier could be to add an invisible whitespace to the path inside the blog post.

marcoow:
The fingerprint replacement also replaces all occurences of `app.js` and `app.css` in the blog posts. That's not what we want of course as it

* makes the content harder to read
* means that **every** deployment will also change the blog bundle even if no new blog post released - that will invalidate the bundle if everyone's caches as well

As I'm not aware of a better solution, I'm simplabs inserting a non-visible whitespace into occurences of the respective file names so that e.g. `app.js` becomes `app<non-visible whitespace>.js` which should prevent the replacing. As nobody is likely to copy/paste these file names anyway that should be a good-enough solution.

closes #630

### DIFF
--- a/lib/markdown-content/htmlize-markdown.js
+++ b/lib/markdown-content/htmlize-markdown.js
@@ -5,6 +5,7 @@ const marked = require('marked');
 const jsdom = require('jsdom');
 
 module.exports = function htmlizeMarkdown(source, options = {}, callback = null) {
+  source = preventBundleFingerprintReplacement(source);
   let html = marked(source, options);
   html = manipulateDom(html, dom => {
     dom.querySelectorAll('a').forEach(a => {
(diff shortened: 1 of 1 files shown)


lib/markdown-content/htmlize-markdown.js
```diff
@@ -27,3 +28,14 @@ function manipulateDom(html, callback) {
...
+
+// During a production build, references to app.js and app.css will be replaced
+// with their fingerprinted versions in the entire website. Of course we do not
+// actually want mentions of these filenames in blog posts to be replaced so
+// we're replacing any mentions of these files with a version that includes a
+// non-visible whitespace so they won't match the regex that
+// broccoli-asset-rewrite uses but will still look correct to the reader. See
+// https://github.com/simplabs/simplabs.github.io/issues/630 for reference.
+function preventBundleFingerprintReplacement(source) {
+  return source.replace(/app\.js/g, 'app​.js').replace(/app\.css/g, 'app​.css');
```
pichfl: I would prefer using `\u200b` in the replacement string instead of the actual invisible character. If it doesn't work you can just ignore that comment.

pichfl (changes requested):

lib/markdown-content/htmlize-markdown.js
marcoow: Yes, that makes sense 👍 - updated

pichfl (approved):

PR #652, opened 2019-07-26
Improve blog layout

### LINKED ISSUE #561: Improve layout for blog posts

pichfl:
Closes #561


marcoow (changes requested):
There's now a lot of whitespace between the excerpt and author info on each post on the list page:

Bildschirmfoto 2019-07-26 um 12 23 15

There's also still too much whitespace between the tag and the title of the featured post on the list page:

Bildschirmfoto 2019-07-26 um 12 24 10

Big pictures don't work on mobile:

Bildschirmfoto 2019-07-26 um 12 25 24

Should there be a bit more space between the author info and the first paragraph on mobile?

Bildschirmfoto 2019-07-26 um 12 25 53

_posts/2018-07-24-from-spa-to-pwa.md
```diff
@@ -200,7 +200,7 @@ of the network is that it **must be able to start up while there is no network
 at all** and the device is offline. For a browser-based application designed to
 load all required assets remotely, that is no easy task. One of the first
 approaches for achieving this was
-[Google Gears](https://en.wikipedia.org/wiki/Gears_(software)). Gears was
+[Google Gears](<https://en.wikipedia.org/wiki/Gears_(software)>). Gears was
```
marcoow, line 203: what does this change do?

lib/generate-blog/lib/templates/author/template.hbs
```diff
@@ -1,34 +1,35 @@
...
-        </a>
-        <p>
-          {{bio}}
-        </p>
-        <a href="https://twitter.com/{{author.twitter}}" target="_blank" rel="noopener">
-          @
-          {{twitter}}
-        </a>
-      </div>
+  <Header @active="blog" @noContent="true" />
```
marcoow: Is there no [`hasBlock`](https://api.emberjs.com/ember/3.11/classes/Ember.Templates.helpers/methods/hasBlock?anchor=hasBlock) in Glimmer.js?

lib/generate-blog/lib/templates/author/template.hbs
```diff
@@ -1,34 +1,35 @@
...
+  <header class="blog-post.header">
+    <h1 class="typography.display">
+      <small class="typography.small">
+        Blog<span class="typography.hidden">:</span>
```
marcoow: Why do we need this hidden `:`?

_posts/2018-07-24-from-spa-to-pwa.md
pichfl: Seems to be the way Prettier wants to escape that link as it contains parenthesis. I had run Prettier on these files by accident and must have missed that one when reverting.

lib/generate-blog/lib/templates/author/template.hbs
pichfl: None that I know of. Trying to use `hasBlock` inside a template halts rendering.

lib/generate-blog/lib/templates/author/template.hbs
pichfl: It's not part of the design (ending a label in punctuation doesn't look quite as good) but part of the written semantics for screen readers.

_posts/2018-07-24-from-spa-to-pwa.md
marcoow: ah ok 👍

lib/generate-blog/lib/templates/author/template.hbs
marcoow: ah ok

lib/generate-blog/lib/templates/author/template.hbs
marcoow: Hm, this seems to add it though: https://github.com/glimmerjs/glimmer-vm/pull/112

pichfl: 😯 thanks for digging there. Turns out you have to use the helper like this: `{{#if (has-block)}}`.

pichfl: While has-block exists, it doesn't seem to work 😞 it's always true, even for self closing invocations.

marcoow: Isn't it a problem though that with this aproach we always have 2 `<header>` elements on all blog pages now?

Bildschirmfoto 2019-07-26 um 16 41 33

I guess if we have different scenarios for the `<Header>` component now it would make sense to break up the header and header content up like so

```js
<Header @active="services" />
  <HeaderContent @label="Services" @title="Full-Stack Engineering" />
</Header>
```

Then in the blog we could do something like this:

```js
<Header @active="blog" />
  <h1 class="typography.display">
    <small class="typography.small">
      Blog<span class="typography.hidden">:</span>
    </small>
    Posts by
    {{name}}
  </h1>
</Header>
```

I think the problem we have here (besides `has-block` not working correctly of course) is that the `<Header>` component is too big and not flexible enough for what we want.

pichfl: The additional header doesn't hurt us semantically, we could even add an additional `<section>` to the blog to make the separation more distinct.

I agree on the Header being rather too big though, but I'm not sure ho to split it as it has to have some part of it's markup for the navigation to work as expected.

marcoow: So what I'm suggesting is to move

```hbs
<div class="content" state:flex-content={{@flex-content}}>
  {{#if @title}}
    <h1 class="typography.display">
      {{#if @label}}
        <small class="typography.small">
          {{@label}}<span class="typography.hidden">:</span>
        </small>
      {{/if}}
      {{@title}}
    </h1>
  {{/if}}
  {{yield}}
</div>
```

into a new `<HeaderContent>` component so we'd call the header as

```hbs
<Header @active="services" />
  <HeaderContent @label="Services" @title="Full-Stack Engineering" />
</Header>
```

in most cases while in the blog we do:

```hbs
<Header @active="blog" />
  <h1 class="typography.display">
    <small class="typography.small">
      Blog<span class="typography.hidden">:</span>
    </small>
    Posts by
    {{name}}
  </h1>
</Header>
```

The `<navbar>` can remain in the header as we **always** want that - we're only moving the part out that we only want in most but not all cases.

ghislaineguerin (approved):

lib/generate-blog/lib/templates/author/template.hbs
pichfl: The problem here is that this will still nest the contents that need to be after the header into the header. Doing that will make it much harder to use the existing CSS grid to layout the author badge. I think it's time for me to make a small codepen to test this out.

pichfl:
@marcoow can we consider merging this right now and make a separate issue for the header refactoring?

marcoow:
We could do that but I'd say we should work on that right away then and leave #561 open as well. I'd say during the course of working on this we discovered the current structure doesn't really support this scenario and thus is the refactoring now part of this issue as it is required to resolve it properly.

pichfl:
Good point. Went ahead and refactored as suggested.

marcoow (changes requested):
Looks good - there's only 1 thing left: there's still too much vertical spacing between the excerpt and the author and date on the blog index:

Bildschirmfoto 2019-07-31 um 11 16 35

marcoow (approved):

Issue #654, opened 2019-07-26
Adding dynamic meta tags to head


pichfl:
We need to add several additional HTML meta tags to the `<head>` depending on the page, which means we have to add them to the routing map or find some other way to add these individually.

This issue is only for implementing that mechanism. Necessary tags are determined in #412, #627 and others.

Necessary meta tags:
```html
<meta name="description"  property="og:description" content="...add something with < 165 chars here...">

<meta property="og:site_name" content="simplabs" /> <!-- For all pages, this is the *site*, not the page -->
<meta name="twitter:title" property="og:title" content="Plain page title (ie. just 'Services')" />
<meta property="og:type" content="website" /> <!-- Static for everything but blog pages where it should be `article` -->
<meta property="og:url" content="Canonic URL" />
<meta name="twitter:image:src" property="og:image" content="http://ia.media-imdb.com/images/rock.jpg" /> <!-- image must be at least 280x150px -->
<meta property="fb:admins" content="Facebook numeric ID" /> <!-- Static for everything -->
<meta name="twitter:card" content="summary> <!-- for all pages -->
<meta name="twitter:site" content="@simplabs"> <!-- same for all pages -->
<meta name="twitter:description" content="Duplicates meta name=description"> <!-- optional, twitter may fall back to the other description. Or not. -->

<!-- If we can have those, they would be nice -->
<meta property="article:published_time" content="2013-09-17T05:59:00+01:00" />
<meta property="article:modified_time" content="2013-09-16T19:08:47+01:00" />
```

- [x] figure out where to put these meta tags
- [x] implement the solution
- [ ] Add all necessary meta tags (see list above)

marcoow:
@pichfl I think there's something missing here?

```html
<meta  content="http://www.example.com/image.jpg">
```

pichfl:
@marcoow yes, I just removed it. That one was collapsed into 
```html
<meta name="twitter:image:src" property="og:image" content="path/to/image.jpg" />
```

marcoow:
I'm  closing  this for now as we have as all of  this  is done except for a few tags that are handled via #655, #672.

PR #659, opened 2019-07-30
Add mechanism for arbitrary head tags

marcoow:
This adds a mechanism for defining arbitrary head tags. It introduces a `HeadTags` class that is injected as `headTags` into all components and that has 2 methods:

* `write` writes a head tag or updates its attributes or content if it exists already; it takes a `tagName`, `keys` (attributes that are used to check if the tag exists already), `values` (attributes that are not used to check if the tag exists already) and `textContent`
* `remove` removes a tag from the `<head>` again; it takes it takes a `tagName` and `keys` (attributes that are used to check if the tag exists already)

This is the foundation for #654. This MR also adds the following tags:

* `meta[property="article:published_time"]` for blog posts
* `meta[property="fb:admins"]`
* `meta[property="og:site_name"]`
* `meta[name="twitter:site"]`


src/utils/head-tags.ts
```diff
@@ -0,0 +1,49 @@
+export default class HeadTags {
+
+
+  public static create(): HeadTags {
+    return new HeadTags();
+  }
+  private document: HTMLDocument = window.document;
+
+  public write(tagName, keyAttrs = {}, contentAttrs = {}, textContent = null) {
```
pichfl, line 9: I'm not sure why `keyAttrs` and `contentAttrs` are split here if they end up getting merged. Even for the component usage, I think having a single object makes for a better dev experience.

pichfl (changes requested):
Very elegant

src/utils/head-tags.ts
marcoow: The key (😄) here is that `keyAttrs` are

> attributes that are used to check if the tag exists already

while `contentAttrs` are not. If a tag `meta[name="twitter:site"]` exists with a different `content` attribute we want to update it vs. adding another one. `HeadTags` needs to be able to understand which attributes should be used to find an element that should potentially be updated vs. which attributes to ignore for looking up the element.

pichfl: I think documenting this to some extend might be helpful. I see how it works in the code after your explanation.

pichfl (approved):

Issue #627, opened 2019-07-17
Create meta description, keywords and OG tags for all pages


ghislaineguerin:
@pichfl: This *won't* affect our page rank, but it makes the search results and social media embeds better, which why I would prioritize this.

- [x] Description: 160 characters or less per page. Not used for SEO directly, but improves the text below the title in search results and is used as fallback for Open Graph content (Facebook, Twitter, Slack, etc)

### Pages that need content

- [x] Home
- [x] Services
- [x] Services - Engineering
- [x] Services - Augmentation
- [x] Services - Tutoring
- [x] Expertise - Ember
- [x] Expertise - Elixir
- [x] Work
- [x] Work - Trainline
- [x] Work - Timify
- [x] Work - DDWRT
- [x] Work - Expedition
- [x] Why simplabs
- [x] Blog
- [x] Calendar
- [x] Talks
- [x] Contact us

This issue is for creating the content, implementing should be done in a separate issue.

marcoow:
* Home

> We deliver ambitious digital products on the web and mobile devices for our clients to rely on. Our team of experts manages projects from idea to release.

* Services

> We turns ideas into products, managing projects from start to end and helps tech teams move faster with confidence via Team Augmentation and Tuturing.

* Services > Full Stack Engineering

> Our expert team guides clients through the complete project from idea to release, managing everything from concept and design to engineering and operations.

* Services > Team Augmentation

> By augmenting our clients’ engineering teams, we grow them both in numbers and capability. We will share our know-how and expertise for sustainable value.

* Services > Tutoring

> We tutor teams that approach new technologies, overcome long standing issues or struggle with delivering large scale projects on time and budget.

* Work

> simplabs has a proven track record of successful projects for international clients. We have worked with companies from small startups to big enterprises.

* Work > trainline

> Trainline is Europe’s leading rail and coach platform. We helper them deliver a high-performance mobile web app, along with an improved engineering process.

* Work > Timify

> We rebuilt Timify's MVP into a well-architected system that would be the foundation for future evolution, setting them up for sustainable long-term success.

* Work > Expedition

> simplabs helped Expedition to develop a stable and sustainable architecture, getting the most out of their technology stack based on Ember.js and Elixir.

* Work > DD-WRT

> DD-WRT is a firmware for wireless routers running on millions of devices worldwide. simplabs developed an Ember.js based foundation for a new configuration UI.

* Why simplabs

[...]

PR #662, opened 2019-07-30
Better handling of title tag and related meta tags

marcoow:
This is a bit of cleanup related to the work on #627. Previously we were setting the title tag and related tags in the `Simplabs` component's source which meant we had to have the title etc. in the route map. As we're aiming to have more page-component specific meta tags like `meta[name="description"]` with that solution we would have had all these tags and their content in the routes map as well which would have gotten messy quickly and would have grown the JSON we're injecting in every page to get the route map into the running app quite a bit. With this approach we're simply using the  new `<HeadTag>` component to render tags in the `<head>` which we can invoke from any template - that way we get the same  functionality but can keep all of the tags attributes and content in the templates where it belongs.

### DIFF
--- a/src/ui/components/Header/template.hbs
+++ b/src/ui/components/Header/template.hbs
@@ -48,14 +48,14 @@
     </div>
   </div>
   <div class="content" state:flex-content={{@flex-content}}>
-    {{#if @title}}
+    {{#if @headline}}
       <h1 class="typography.display">
         {{#if @label}}
           <small class="typography.small">
             {{@label}}<span class="typography.hidden">:</span>
           </small>
         {{/if}}
-        {{@title}}
+        {{@headline}}
       </h1>
     {{/if}}
     {{yield}}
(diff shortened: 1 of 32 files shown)


src/ui/components/Header/template.hbs
```diff
@@ -1,4 +1,6 @@
 <header>
+  <HeadTag @name="title" @content={{this.pageTitle}} />
```
pichfl: `documentTitle` would be more idiomatic considering the browser and web api in general

pichfl (changes requested):
Small nitpick, but I hope it would be worth the change.

src/ui/components/Header/template.hbs
marcoow: yep, good point - changed 👍

pichfl:
The blog index somehow lost its headline.

marcoow:
@pichfl yes, fixed that already

pichfl (approved):

PR #668, opened 2019-07-31
Replace random shape id's by counter based ids for rehydration

### LINKED ISSUE #608: Shape* components should not rely on random ids to reference SVG nodes
@marcoow says in review for #596 

> This is problematic as it will be different in the SSR pre-render and the actual render in the browser. Can we have something that is stable?

https://github.com/simplabs/simplabs.github.io/blob/f10dc5c59b40c4a3cf8456605df3c086223317de/src/ui/components/ShapeBase/component.ts#L4-L12

For more details see discussion there.

- [x] find a more stable way so SVG in components can reference their `<defs>` in some way
- [x] use variable local to the module to count

pichfl:
Closes #608

### DIFF
--- a/src/ui/components/ShapeAcross/component.ts
+++ b/src/ui/components/ShapeAcross/component.ts
@@ -1,14 +1,14 @@
 import ShapeBase from '../ShapeBase/component';
 
+let maskId = 0;
+
 export default class ShapeAcross extends ShapeBase {
-  public maskId = null;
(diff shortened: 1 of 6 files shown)


marcoow (approved):
I just realise the old code probably didn't do what was intended anyway as assigning the random ID in the class body (vs. the constructor) should result in the same random ID being used for **every** instance.

PR #670, opened 2019-07-31
add descriptions for pages

### LINKED ISSUE #654: Adding dynamic meta tags to head

marcoow:
This adds `meta[name="description"][property="og:description"]` and `meta[name="twitter:description"]` tags for all pages except the blog posts. It's one more step in order to close #654 but we need to add descriptions for the blog posts still.

### DIFF
--- a/lib/generate-blog/lib/templates/author/template.hbs
+++ b/lib/generate-blog/lib/templates/author/template.hbs
@@ -1,5 +1,9 @@
 <div>
-  <Header @active="blog" @title="Posts by {{name}}" @documentTitle="Posts by {{name}} | Blog">
+  <Header
+    @active="blog"
+    @title="Posts by {{name}}" @documentTitle="Posts by {{name}} | Blog"
+    @description="Our teams shares what they learn in our blog, covering topics around tech, design/UX and the engineering process. Find all posts by {{name}} here."
+  >
     <div class="blog-post.header">
       <div class="blog-post.headline">
         <h1 class="typography.display">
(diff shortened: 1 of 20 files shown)


pichfl (approved):
I think those should work, but I'm no native speaker obviously.

PR #671, opened 2019-07-31
add og:type, twitter:card meta tags

marcoow:
This addresses parts of #654 by adding the `meta[property="og:type"]` and `meta[name="twitter:card]"` tags to the document head.

### DIFF
--- a/src/ui/components/Header/component.ts
+++ b/src/ui/components/Header/component.ts
@@ -7,8 +7,9 @@ export default class Header extends Component {
     super(options);
 
     let documentTitle = this.args.documentTitle === undefined ? this.args.title : this.args.documentTitle;
-    
     this.documentTitle = formatformatDocumentTitle(documentTitle);
+
+    this.pageType = this.args.pageType || 'website';
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/ui/components/Header/component.ts
```diff
@@ -7,8 +7,9 @@ export default class Header extends Component {
     super(options);
 
     let documentTitle = this.args.documentTitle === undefined ? this.args.title : this.args.documentTitle;
-    
     this.documentTitle = formatformatDocumentTitle(documentTitle);
+
+    this.pageType = this.args.pageType || 'website';
```
pichfl: This results in a TypeScript warning that `this.pageType` wasn't defined. Should be added as `public pageType = 'website'` above

pichfl (changes requested):

src/ui/components/Header/component.ts
marcoow: done - `pageType` is actually a private field on the componen

pichfl (approved):

PR #683, opened 2019-08-07
Add missing trailing slash to paths in sitemap.xml

### LINKED ISSUE #681: Urls in sitemap.xml don't have trailing slashes
Found via Google Search Console.

- [x] generate sitemap with trailing slashes

pichfl:
Closes #681

### DIFF
--- a/lib/routes/index.js
+++ b/lib/routes/index.js
@@ -47,6 +47,10 @@ function generateSitemap(routes) {
 
   let paths = Object.keys(routes).filter(path => path !== '/404');
   for (let path of paths) {
+    if (path.slice(-1) !== '/') {
+      path = `${path}/`;
+    }
+
     sitemap += `
       <url>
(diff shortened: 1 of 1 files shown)


ghislaineguerin (approved):

lib/routes/index.js
```diff
@@ -47,6 +47,10 @@ function generateSitemap(routes) {
 
   let paths = Object.keys(routes).filter(path => path !== '/404');
   for (let path of paths) {
+    if (path.slice(-1) !== '/') {
```
marcoow, line 50: I'd suggest using `String.prototype.endsWith`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith.

pichfl: Oh right, this file is server side only. Will prepare an update.

PR #685, opened 2019-08-07
Add canonical link tag to head

### LINKED ISSUE #682: Canonical urls
Google Search Console seems to be unhappy about missing "Canonical URLs"

- [x] find out how to change this
- [x] implement necessary meta tags

pichfl:
Google Search Console seems to be unhappy about missing "Canonical URLs", this adds them.

Closes #682

### DIFF
--- a/src/ui/components/Simplabs/component.ts
+++ b/src/ui/components/Simplabs/component.ts
@@ -198,11 +198,32 @@ export default class Simplabs extends Component {
   }
 
   private _setCanonicalUrl(path) {
-    this.headTags.write('meta', {
-      property: 'og:url',
-    }, {
-      content: `https://simplabs.com${path}`,
-    });
+    // Ensure trailing slash
+    if (path.slice(-1) !== '/') {
+      path = `${path}/`;
+    }
+
+    this.headTags.write(
+      'meta',
+      {
+        property: 'og:url',
+      },
+      {
+        content: `https://simplabs.com${path}`,
+      },
+    );
+
+    // Canonical urls are required by search engines to define what represents
+    // the one true URL of a page, which should exclude query params, anchors, …
+    this.headTags.write(
+      'link',
+      {
+        rel: 'canonical',
+      },
+      {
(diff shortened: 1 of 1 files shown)


ghislaineguerin (approved):

src/ui/components/Simplabs/component.ts
```diff
@@ -198,11 +198,32 @@ export default class Simplabs extends Component {
...
+      path = `${path}/`;
+    }
+
+    this.headTags.write(
+      'meta',
+      {
+        property: 'og:url',
+      },
+      {
+        content: `https://simplabs.com${path}`,
```
marcoow, line 212: We should move ``https://simplabs.com${path}`` as it's repeated twice.

src/ui/components/Simplabs/component.ts
```diff
@@ -198,11 +198,32 @@ export default class Simplabs extends Component {
   }
 
   private _setCanonicalUrl(path) {
-    this.headTags.write('meta', {
-      property: 'og:url',
-    }, {
-      content: `https://simplabs.com${path}`,
-    });
+    // Ensure trailing slash
+    if (path.slice(-1) !== '/') {
```
marcoow, line 202: I'd suggest using `String.prototype.endsWith`.

pichfl: While I agree with changing this on the server side, Browser support for `endsWith` isn't that broad. While our page won't look that good in older browsers because of CSS grid, it still works without failing on a JavaScript error (at least to my knowledge)

src/ui/components/Simplabs/component.ts
pichfl: Good point.Fix up PR incoming.

src/ui/components/Simplabs/component.ts
marcoow: The full page works even when js fails as it just falls back to being a regular static Page then - all links are just plain links and result in server requests. I think by using css grid we already decided we don’t care about old browsers.

PR #687, opened 2019-08-07
Fixup changes in PR #683 and #685

pichfl:
- remove repetition in head meta tag configuration
- use `String.endsWith()` when possible*

*: browser-support is lagging behind. We could polyfill but it's only used in one place and the polyfill is larger than using the more old fashioned `String.slice()` option.


marcoow (commented):
I think we can ignore browser support and the polyfill. Everything will always work as we just fall back to server side (pre-)rendered static pages. And by using css grid we already decided we don’t care about old browsers.

marcoow (approved):

Issue #684, opened 2019-08-07
Additional redirects for pages in Googles index


pichfl:
```
https://simplabs.com/jobs/
https://simplabs.com/full-stack-engineering/index.html
https://simplabs.com/dd-wrt/index.html
https://simplabs.com/team-augmentation/index.html
https://simplabs.com/about/index.html
https://simplabs.com/cases/dd-wrt
https://simplabs.com/about
https://simplabs.com/about/
https://simplabs.com/trainline/index.html
https://simplabs.com/expedition/index.html
https://simplabs.com/elixir/index.html
https://simplabs.com/training/2017-02-17-modern-seo.html
https://simplabs.com/sitemap
https://simplabs.com/blog/2015/12/03/ember-cli-deploy-notifications.html
https://simplabs.com/blog/ember/2018/05/30/a-little-encouragement-goes-a-long-way-in-2018.html
https://simplabs.com/timify/index.html
https://simplabs.com/clients/
https://simplabs.com/ember/index.html
https://simplabs.com/team-augmentation/
```

- [x] Add urls to Netlify configuration
- [x] Update Search Console

PR #688, opened 2019-08-07
URL maintainance: Add redirects and mark pages as gone 

### LINKED ISSUE #677: Remove old training pages from google index
A lot of our old training pages are still in Google's index and I think we should remove them, e.g.

* simplabs.com/training/2017-06-11-web-security
* simplabs.com/training/2017-02-16-sass-fundamentals

pichfl:
Closes #677 #684

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -243,6 +243,14 @@
   from = "/blog/2019/05/10/may-monthly-update.html"
   to = "/blog/2019/05/10/may-monthly-update/"
 
+[[redirects]]
+  from = "/blog/2015/12/03/ember-cli-deploy-notifications.html"
+  to = "/blog/2015/12/3/ember-cli-deploy-notifications/"
+
+[[redirects]]
+  from = "/blog/ember/2018/05/30/a-little-encouragement-goes-a-long-way-in-2018.html"
+  to = "/blog/2018/05/30/a-little-encouragement-goes-a-long-way-in-2018"
+
 # redirect old workshop pages
 
 [[redirects]]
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -12,318 +12,452 @@
...
+  from = "/ember/index.html"
+  to = "/expertise/ember/"
+
+[[redirects]]
+  from = "/blog/2015/12/03/ember-cli-deploy-notifications.html"
+  to = "/blog/2015/12/3/ember-cli-deploy-notifications/"
+
+[[redirects]]
+  from = "/blog/ember/2018/05/30/a-little-encouragement-goes-a-long-way-in-2018.html"
+  to = "/blog/2018/05/30/a-little-encouragement-goes-a-long-way-in-2018"
```
marcoow: Can we move these next to the existing redirects for blog posts for better organization?

netlify.toml
```diff
@@ -12,318 +12,452 @@
...
+  from = "/jobs"
+  to = "/"
+  status = 410
+  force = true
+
+[[redirects]]
+  from = "/jobs/*"
+  to = "/"
+  status = 410
+  force = true
```
marcoow: Do we need the `force` here? If so we might need it for the other ones as well?

netlify.toml
```diff
@@ -12,318 +12,452 @@
...
-  from="https://phoenix-workshops.simplabs.com"
-  to="https://simplabs.com/services/tutoring/"
+  from = "https://phoenix-workshops.simplabs.com"
+  to = "https://simplabs.com/services/tutoring/"
 
 [[redirects]]
-  from="https://phoenix-workshops.simplabs.com/*"
-  to="https://simplabs.com/services/tutoring/"
+  from = "https://phoenix-workshops.simplabs.com/*"
+  to = "https://simplabs.com/services/tutoring/"
```
marcoow, line 334: I guess we should add status 410 for these redirects as well?

netlify.toml
```diff
@@ -12,318 +12,452 @@
...
+  from = "training/2017-02-17-modern-seo"
+  to = "/"
+  status = 410
+  force = true
+
+[[redirects]]
+  from = "training/2017-02-17-modern-seo/*"
+  to = "/"
+  status = 410
+  force = true
```
marcoow: Can we redirect these URLs to the tutoring service page (https://simplabs.com/services/tutoring/) as we do for the old workshop domains?

netlify.toml
marcoow: Hm, just read the docs for 410 and they say:

> The resource is no longer available at the requested URI and no redirection will be given.

so I assume a redirect with status 410 is wrong anyway?

netlify.toml
pichfl: You're right, after re-evaluating the docs from Netlify, we should be able to omit `force`.

netlify.toml
pichfl: Netlify offers this as the only way to generate 410 responses, which indeed is a bit weird. 
We should only use 410 for pages that no longer have matching content and I tried to map as much as possible.

netlify.toml
```diff
@@ -12,318 +12,432 @@
...
+  from = "/blog/ember/2018/05/30/a-little-encouragement-goes-a-long-way-in-2018.html"
+  to = "/blog/2018/05/30/a-little-encouragement-goes-a-long-way-in-2018"
 
 # redirect old workshop pages
 
 [[redirects]]
-  from="https://elixir-phoenix-workshop.simplabs.com"
-  to="https://simplabs.com/services/tutoring/"
+  from = "https://elixir-phoenix-workshop.simplabs.com"
+  to = "/services/tutoring/"
```
marcoow: Are you sure this is correct? I'd assume this does not change the domain and redirect to `https://elixir-phoenix-workshop.simplabs.com/services/tutoring/`.

netlify.toml
```diff
@@ -12,318 +12,432 @@
...
+  from = "/training/2017-02-16-sass-fundamentals"
+  to = "/services/tutoring/"
+
+[[redirects]]
+  from = "/training/2017-02-16-sass-fundamentals/*"
+  to = "/services/tutoring/"
+
+[[redirects]]
+  from = "/training/2017-02-17-modern-seo.html"
+  to = "/services/tutoring/"
```
marcoow, line 362: Should these not have status 410 as well then?

pichfl: 410 should only be used if a content is gone without an appropriate replacement. We can consider the tutoring page a good replacement for the training pages as far as I'm concerned.

netlify.toml
pichfl: The documentation is rather inconclusive on this topic. I reverted just in case.

netlify.toml
marcoow: ok 👍

marcoow (approved):

PR #689, opened 2019-08-09
Exclude certain pages from sitemap.xml

### LINKED ISSUE #686: Remove blocked-for-robots pages from sitemap
We are blocking a few pages for robots (see #666):

* `/imprint`
* `/privacy`

However, those pages are still in the `sitemap.xml`. This is not only inconsistent but also Google complains about it so we should remove the pages from `sitemap.xml`.

pichfl:
404, Imprint, and Privacy should not be part of the `sitemap.xml`. This adds a basic filter mechanism that can be controlled from the routes configuration.

Closes #686

### DIFF
--- a/config/routes-map.js
+++ b/config/routes-map.js
@@ -37,7 +37,7 @@ module.exports = function() {
     ...blogPostRoutes,
     ...blogAuthorsRoutes,
     '/': { component: 'PageHomepage' },
-    '/404': { component: 'Page404', sitemap: false },
+    '/404': { component: 'Page404' },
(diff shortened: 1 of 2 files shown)


marcoow (commented):
Everything in `config/routes-map.js` ends up serialized in the document. We're only adding a few properties in this PR of course but I think in general we should avoid doing that at all and keep the routes config as minimal as possible. I'd suggest we instead add a blacklist to the sitemap script.

pichfl:
Okay, not entirely happy with this change, but I see how it's necessary. Updated.

marcoow (approved):

PR #690, opened 2019-08-16
Blog index pagination

### LINKED ISSUE #440: Add pagination to blog index page
The blog index page is very long and also relatively slow as **all** of the blog posts have to be loaded in order to be able to render the page. We need to add pagination to the page to make the page less long and also lighter/faster.

Navigation should be two links, one for next, one for previous page.

### TODO

- [x] Create Markup
- [x] while keeping the main `blog.js` bundle, **also** split that up into bundles for pages (10 posts each); the main `__blog__` bundle is the parent bundle of the per-page bundles
- [x] in the service worker, continue loading the `blog.js` bundle
- [x] implement pagination for the big blog page - each page has its associated bundle then (e.g. `__blog_page_1__`)

### LINKED ISSUE #492: Large amount of wasted horizontal space on the blog index
This issue shows up on quite a few pages but really it's only a "problem" on `/blog` because of the length of the page and the size of some of the affected content. 

The general motif for each of the pages is a "main content" column with a "read more" aside to the right with links that we want to highlight. 

Screenshot 2019-05-21 at 09 55 01

This works **very** well on most pages, and it works well at the _top_ of the `/blog` section, but once it moves past the point where the aside is visible then there is a lot of wasted space for the large majority of the page: 

Image 2019-05-21 at 9 58 28 a m

This feels off, and it is definitely hampered by the lack of pagination (covered in #440). It feels particularly problematic  for some of the content items that have really long intros as they would benefit from the extra space: 

Screenshot 2019-05-21 at 11 03 22

### TODO

- [ ] identify some content that we could put in the sidebar in addition to the current content (don't think we'll have to prepare new content for that)
- [ ] put it there

pichfl:
Since we only use two links, this adds a reverse option to the ArrowLink class which should be well suited for the task. 

The pagination should only show the relevant links, if you're at the start there is only the link to go to older entries, if your at the end, there is only a link to go to newer ones. 

#### TODO

- [x] Implement bundling
- [x] bundle the posts with the pages they appear on so that the page bundles are the parent bundles of the respective posts
- [x] drop the `blog.js` bundle
- [x] preload blog page bundles and author bundles in service worker

closes #440
closes #492


marcoow:
This should be working now and is ready for review. @pichfl: I recognized 2 things:

* when I'm on the last page, the _"Newer"_ link appears on the left vs. the right where it would normally be:

Bildschirmfoto 2019-08-21 um 11 46 37

* while the arrow of the _"Older"_ link is **outside** of the surrounding container, the arrow of the _"Newer"_ link is **inside** - not sure that's intentional?

Bildschirmfoto 2019-08-21 um 11 46 58

marcoow (approved):

PR #694, opened 2019-08-21
Run Prettier on Markdown content

### LINKED ISSUE #644: Use prettier for markdown content
Since Prettier has support for Markdown and we use it for most things in this repo, I think it might be a good idea to run it here as well. This should prevent at least some common mistakes when handling Markdown.

- [x] Add a build task for this step
- [x] Add a test task for this
- [x] Run the build task on existing content

pichfl:
Adds `<!-- prettier-ignore -->` statements to most code blocks to keep them as they were before.

Closes #644

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,9 @@ jobs:
       name: "Lint - CSS"
       script: yarn lint:css
     - stage: test
+      name: "Lint - Markdown"
+      script: yarn lint:md
+    - stage: test
       name: "Crawler"
       script: yarn build && yarn crawl
     - stage: test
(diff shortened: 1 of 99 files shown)


locks:
I noticed a lot of trailing commas in code blocks, did you make sure that they are all supposed to support [ECMAScript 5](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas#Objects)?

[edit] D'oh, ECMAScript 5 is quite old so I think we are safe xD

pichfl:
I did work through the blog posts and I found ECMA5 to be an acceptable bottom line.

locks (approved):

PR #699, opened 2019-08-26
Add open graph images

### LINKED ISSUE #421: Create Open:Graph images for Facebook and Twitter
> Add featured images to blog posts and create thumbnails of our talks with the highest impressions on image search (ember, pwa, glimmer)

- [x] Create templates for blog post types
- [x] Check for relevant posts and talks on Google Analytics / Search console
- [x] Create images for those
- [x] Add them to the OG image tags (`meta[name="twitter:image:src"][property="og:image"]`)

ghislaineguerin:
Closes #421


ghislaineguerin:
@pichfl The og:image property is still showing blank content. Do you have any idea why?

marcoow (changes requested):
* public/assets/images/posts/2018-12-10-assert-your-style/og-image.png 

The post isn't really about Percy - I think the best fit for the logo might be the CSS logo https://www.google.com/search?q=css+logo&client=safari&rls=en&source=lnms&tbm=isch&sa=X&ved=0ahUKEwjwhuTLuaLkAhUEPFAKHWdAAgsQ_AUIESgB&biw=1250&bih=706 - @jessica-jordan what do you think?

Overall the images look pretty big both in terms of dimensions and in terms of file size. If we need these dimensions (I guess there's some kind of spec or best practice), can we run them through ImageOptim?

ghislaineguerin:
@marcoow I am using the default size for open graph but saving as 2x for retina displays, not sure if this is necessary so I will look into that. I agree that we must optimize them.
As for the logo, I can switch that to a CSS one.

marcoow (changes requested):
public/assets/images/posts/2018-07-24-from-spa-to-pwa/og-image.png 

I think that images needs to be run through ImageOptim.

Also 1 more thing that needs to be changed: we don't want the og images to end up in the service worker and should exclude them here: https://github.com/simplabs/simplabs.github.io/blob/master/lib/service-workers/lib/worker-builder.js#L36 - I can do that

marcoow:
@ghislaineguerin I made the fixes for the service worker

lib/generate-blog/lib/templates/post/template.hbs
```diff
@@ -2,6 +2,7 @@
   <HeadTag @name='meta' @keys=\{{hash property="article:published_time"}} @values=\{{hash content="{{isoDate}}"}} />
   <HeadTag @name='meta' @keys=\{{hash property="article:section"}} @values=\{{hash content="{{topic}}"}} />
   <HeadTag @name='meta' @keys=\{{hash name="twitter:creator"}} @values=\{{hash content="{{author.twitter}}"}} />
+  <HeadTag @name='meta' @keys=\{{hash property="og:image"}} @values=\{{hash content="{{og.image}}"}} />
```
marcoow: I assume this needs to be an absolute URL

marcoow (changes requested):

ghislaineguerin:
@marcoow I've run the images through img optim. I will try again and see if it makes a difference.

lib/generate-blog/lib/templates/post/template.hbs
marcoow: I fixed it

marcoow (approved):

PR #700, opened 2019-08-27
Fix Rehydration of SVG shapes

### LINKED ISSUE #698: Rehydration of SVG elements with IDs fails
When the app rehydrates SVG elements with that use IDs (e.g. for `xlink:href`) it throws away the DOM and renders new as the IDs will be different. That leads to a flash where the content briefly goes away before it then comes back when the app re-rendered it. This is caused by the IDs being different in the server-rendered document and the code in the client app. Although we addressed this in #668, that fix does not actually work. When the app starts up in the browser, all counters are reset to 0 while when pre-rendering, the same app instance is used to render multiple pages which means the counters will increment across pages. So what happens is we have an ID like `#base55` in the pre-rendered document but the client starts the counter at 1 again so we get an ID like `#base1` there which breaks rehydration.

marcoow:
This fixes the rehydration of SVG elements by using fixed shape IDs instead of statically incremented ones that will be different on the server and the client. We are simply using a `@key` argument now that we assign a meaningful default value to so that you only have to set the value explicitly if you're using the same shape more than once in the same parent component (which only happens twice in the entire app). This should be a good compromise between ease of use and the amount of code/work we have to do to get that.

This also removes critical as that failed to include **all** of the critical CSS which lead to a flash of unstyled content and I think caused more harm than good in our case. The site should be pretty fast anyway.

closes #698

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,7 @@ The source code for [https://simplabs.com](https://simplabs.com).
 We use [Glimmer.js](https://glimmerjs.com) for rendering,
 [navigo](https://github.com/krasimir/navigo) for client-side routing and
 [CSS Blocks](https://css-blocks.com) for CSS. Pages are statically pre-rendered
-using [Puppeteer](https://pptr.dev) and we inject critical CSS with
-[critical](https://github.com/addyosmani/critical).
+using [Puppeteer](https://pptr.dev).
 
 ## Installation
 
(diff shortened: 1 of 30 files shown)


marcoow:
This should be more stable now but loading still doesn't look quite right - the shape briefly goes too far to the right before it's correct again:

load

any idea @pichfl?

marcoow:
I got it: critical is broken somehow - when you block the CSS file you get this: 


deploy-preview-700--objective-northcutt-37494c netlify com_

marcoow:
I removed critical

pichfl (approved):

Issue #664, opened 2019-07-31
TypeScript warnings should break CI


pichfl:
We believe the following TypeScript warnings could break CI.

These are new: 
```
src/index.ts(51,5): error TS1068: Unexpected token. A constructor, method, accessor, or property was expected.
src/index.ts(88,21): error TS2339: Property 'appName' does not exist on type 'App'.
src/index.ts(94,21): error TS2339: Property 'appName' does not exist on type 'App'.
src/index.ts(98,25): error TS2339: Property 'appName' does not exist on type 'App'.
src/index.ts(104,25): error TS2339: Property 'appName' does not exist on type 'App'.
src/ui/components/HeadTag/component.ts(20,10): error TS2339: Property 'headTags' does not exist on type 'HeadTag'.
src/ui/components/HeadTag/component.ts(24,10): error TS2339: Property 'headTags' does not exist on type 'HeadTag'.
src/ui/components/Simplabs/component.ts(201,10): error TS2339: Property 'headTags' does not exist on type 'Simplabs'.
```

marcoow:
I think what we really want to do is treat warnings as errors so they break CI.

pichfl:
Good point. Updated the title.

pichfl:
I tried some things today which went nowhere. While this would be a good thing, we also have warnings coming from glimmer itself which we can't fix so we would have to parse them individually.

PR #706, opened 2019-09-02
Fixes for TypeScript warnings

pichfl:
Does not quite fix with #664 but helps a bit.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -4,6 +4,8 @@ import moduleMap from '../config/module-map';
 import resolverConfiguration from '../config/resolver-configuration';
 
 export default class App extends Application {
+  public appName: string;
+
   constructor(options = { hasSSRBody: false, element: null }) {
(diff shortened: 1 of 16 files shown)


src/ui/components/Simplabs/component.ts
```diff
@@ -49,6 +48,8 @@ export default class Simplabs extends Component {
   @tracked
   private isCookieBannerVisible: boolean = false;
 
+  private headTags: any;
```
marcoow: This should be `HeadTags` instead of `any`.

src/main.ts
```diff
@@ -4,6 +4,8 @@ import moduleMap from '../config/module-map';
 import resolverConfiguration from '../config/resolver-configuration';
 
 export default class App extends Application {
+  public appName: any;
```
marcoow: I think this should be `string`.

marcoow (approved):

PR #715, opened 2019-09-27
Add `qunit-console-grouper` blog post

Turbo87:
see https://github.com/simplabs/qunit-console-grouper/


_posts/2019-09-27-qunit-console-grouper.md
```diff
@@ -0,0 +1,48 @@
...
+Previously, all those warnings would have been in one long list, with no
+reference to the test that caused them. With qunit-console-grouper it is now
+much easier to figure out that the `linkAttachmentTask` test suite in this
+example is causing these deprecation warnings.
+
+If you want to give this a try, the install instructions are in our
+[README](https://github.com/simplabs/qunit-console-grouper/#install) and if you
+have any feedback we'd love to hear from you in the
+[Issues](https://github.com/simplabs/qunit-console-grouper/issues/new) section
+of the project! ✨
```
jayjayjpg, line 45: :sparkles:

_posts/2019-09-27-qunit-console-grouper.md
```diff
@@ -0,0 +1,48 @@
...
+[QUnit]: https://qunitjs.com/
+[Ember.js]: https://emberjs.com/
+
+<!--break-->
+
+## Introducing ...
+
+🎉 [qunit-console-grouper] 🎉
+
+This new project is a regular QUnit plugin that can be used via a `<script>`
```
jayjayjpg, line 27: If it is a QUnit plugin, this means you could install it any kind of JavaScript app, correct? Would it be useful to highlight this with something along the lines of

```
This new project is a regular QUnit plugin that can be used via a `script` tag to be installed into any kind of JavaScript test suite, but as it is also an Ember addon, installing qunit-console-grouper into your Ember app is even easier than that and only one `npm install` (or `yarn add`) away. This will automatically hook....
```

jayjayjpg (approved):
This reads great! I left one more small question / suggestion but otherwise this seems good to go - thank you for another :sparkles: blog post!

_posts/2019-09-27-qunit-console-grouper.md
Turbo87: > any kind of JavaScript app

would imply that it also works for Mocha or Jest, but it's specifically for QUnit. I thought mentioning it should probably be enough. I assume that the majority of people that use QUnit mostly do it because it is the Ember default.

marcoow (commented):
I'm not sure the announcement along is worth a blog post tbh.

It might be more valuable to the reader if the post described the problem a bit first (which it does already), but then introduces the `console.group` API (which I for examples wasn't even aware of) and only in the end mentioned we also have an addon that automatically does this. That could mean mostly regrouping the existing content, e.g. leaving the intro but then move the screenshot before the announcement and add some general intro to `console.group` before that.

I think that's a general thing we also want to do in talks etc. which is not to announce something that we like to talk about directly but to focus the content more on the *general* problem and solution and then make the announcement a nice surprise in the end kind of.

Turbo87:
that seems like an implementation detail to me that is not necessarily relevant to the *user* of the addon 🤔 

I mean, we can write a blog post reiterating all of https://developer.mozilla.org/en-US/docs/Web/API/console but I'm not sure how valuable that would be.

if we don't want this annoucement then I can close the PR and just put out a tweet about it, but since there are no other blog posts in the pipeline I thought that this would be better than nothing.

marcoow:
Yeah so what I'm suggesting is make the post about the implementation detail and make the announcement a side note. I think posts that are announcements *only* are not great for the same reasons talks like that are not great.

While we should definitely have new blog posts, I don't think *anything* is better than nothing. Don't get me wrong - I appreciate the general idea and effort, I just think this can be a better post that can deliver value to the reader besides the pure announcement.

Turbo87:
I'll announce it on Twitter then. I don't currently have enough free time to write a full blown blog post about all of `console`.

Issue #707, opened 2019-09-03
Improve layout for smaller screens


pichfl:
We should improve how our website looks smaller screen such as the iPad and the MacBook Pro 13". To do so, we should reconsider and optimize our vertical rhythm and page offsets. 

Since we have all of these converted into ratios and parametric changes, we can do so by creating a single mockup for a page and everything else will fall into place automatically. There might be minor glitches, but those will be easily identified by Percy. 

To do this

- [x] Create visual mockup for a smaller base width
- [x] Consider adding another breakpoint for larger screens (based on the current measurements)
- [x] Implement the changed size values
- [x] Verify on devices

#### Shape cleanup

- [x] ShapeBlog
- [x] ShapeFeature
- [x] ShapeAcross
- [x] ShapeBold
- [x] ShapeDecorative
- [ ] ShapeWork

This is related to, but not a full successor to #439

marcoow:
I'm closing this as the only thing that's still open is work on the shapes but those are going to change anyway.

Issue #613, opened 2019-07-10
Design improved work page shape


ghislaineguerin:
The current shape being used for case studies page has the following issues:

- Responsive design breaks at lower resolutions
- Contains blocks of paragraph text that create layout issues
- Is not up to date with other shape components in the site

We need to:

- [ ] Design a new shape that solves responsive and layout issues
- [ ] Implement and replace the new shape across the site

PR #722, opened 2019-10-14
Use CSS variables for font sizing and offsets

pichfl:
Re #707


pichfl:
I force pushed a quick update to make the markdown linter happy, should be ready to be merged.

Turbo87 (approved):
still to big for my taste, but better than before... 😅

marcoow (changes requested):
I reviewed the Percy screenshots and there's a bunch of visual errors - I left comments on Percy. Maybe this can be split into several separate PRs so changes have effects on fewer visual elements and are easier to visually review.

pichfl:
I can try, but changing the font sizes will mark everything everywhere as changed in all cases.

marcoow:
Sure, but this also removes some of the shapes etc. which I don’t think necessarily comes with changed font sizes (and I’m not sure is good in general)?

pichfl:
It removes the `ShapeWork` component, as it was the one component that fell apart in several ways with the new sizings and was scheduled for a rework in any case. Keeping it as is looked worse than just removing it and adding it back in when #613 is worked on.

Will work through the Percy comments.

marcoow:
> Keeping it as is looked worse than just removing it and adding it back in when #613 is worked on.

sounds like we might be doing things in the wrong order then?

pichfl:
Most likely so, but this was not expected at the time. I think it might make sense to discuss this offline.

pichfl:
I'm closing this down. We are moving to different shapes and better colors and sizes and can always pick the changes made here if necessary.

PR #724, opened 2019-10-16
Update ShapeFeatureEmberDocumentary

### LINKED ISSUE #714: Documentary Teaser wider than screen on mobile
The teaser to the Documentary on the Ember Expertise page is wider than the viewport on mobile devices (or on my iPhone X at least):

IMG_2392

### LINKED ISSUE #718: Ember Expertise page: Ember.js documentary video not playing in Safari
Needs further investigation. At first glance these things become apparent:
 
- the videos url doesn't seem to be properly fingerprinted
- we should not cache the video in the service worker because of its size

Possible helpful: Safari has known issues with videos in service workers (see [here](https://adactio.com/journal/14452) for details)

pichfl:
Closes #714 
Closes #718


ghislaineguerin (approved):

marcoow (commented):
The auto-playing doesn't seem to work all of the time and when it doesn't work you see nothing as we're not using a background image anymore apparently (or that doesn't load for some reason)?

Bildschirmfoto 2019-10-16 um 12 40 30

In general, we should split PRs like this one in 2 btw - one for fixing the autoplay, one for fixing the layout.

pichfl:
Re splitting the PR: The two issues are entangled, as I had to rearrange the video binary data for the autoplay to be better and I needed to redo the video to fix the shape (which is now part of the video). Will look into the missing cover image, there should be one included in the project, not sure where it went.

marcoow (commented):
Hm, still doesn't seem correct:

Bildschirmfoto 2019-10-16 um 14 32 56

Maybe it's slow to load and we don't show the background image **while** the video is loading?

pichfl:
Summoning @mansona for help with Netlify.

pichfl:
Yeah, don't think Netlify does support video hosting. I corrected the issues with the poster, which should be good enough for now.

marcoow (changes requested):
I think this works well but can we use a different poster that's not showing me? I'm already on the teaser on the home page and there are other team members featured in the documentary as well of course that we could  be showing here.

pichfl:
Sure, will update 👍

marcoow (approved):

PR #725, opened 2019-10-19
Check if navigator.serviceWorker is available before accessing methods

### LINKED ISSUE #723: Blog loading issues in Safari
- Trying to navigate to the blog from any other page results in an empty white page (cookie banner shows up, no errors in console)
- Trying to navigate to any other page after a refresh on the blog (which serves the prerendered version) fails

Occurs in Safari Version 13.0.2 (14608.2.40.1.2) using both regular and private browsing modes

pichfl:
Closes #723

### DIFF
--- a/src/ui/components/LazyPreloadTrigger/component.ts
+++ b/src/ui/components/LazyPreloadTrigger/component.ts
@@ -9,13 +9,13 @@ export default class LazyPreloadTrigger extends Component {
     this.preload();
   }
 
-  private preload() {
+  private async preload() {
     if (!this.appState.isSSR && 'serviceWorker' in navigator) {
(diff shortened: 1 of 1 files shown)


marcoow:
Hm, this makes sense but did you confirm that #723 was indeed caused by this?

marcoow (commented):
Besides this, I think the real problem is we're not waiting until the [service worker is ready](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/ready).

pichfl:
It does seem to provide a fix, but you're right, we should use the promise over a one shot test for the controller

marcoow (approved):

PR #736, opened 2019-10-29
Ensure ArrowLink animation doesn't bounce in Safari

### LINKED ISSUE #711: Arrow link animation is slightly off in Safari
The arrow link animation seems slightly off in Safari with very subtle vertical movement of the arrow which makes it look like the arrow _"jumps"_ forwards and backwards:

bug

I think what we want is **horizontal movement only**.

pichfl:
Closes #711

### DIFF
--- a/src/ui/components/ArrowLink/stylesheet.css
+++ b/src/ui/components/ArrowLink/stylesheet.css
@@ -31,12 +31,13 @@
   width: 5rem;
   height: 5rem;
   transition: transform 0.2s;
-  transform: translate(0);
+  transform-origin: bottom left;
+  transform: translate(0, 0);
   pointer-events: none;
 }
 
 :scope:hover .svg {
-  transform: translate(0.5rem);
+  transform: translate(0.5rem, 0);
 }
 
 .arrow {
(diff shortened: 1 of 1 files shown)


marcoow (commented):
What's strange is that it is slightly lagging and it seems to affect the rendering of the above text:

Bildschirmvideo aufnehmen 2019-10-29 um 14 16 58

That only happens on the arrow link in the teaser to the trainline success story though - it seems to interfere with the shape somehow?

pichfl:
Okay, that's odd. Doesn't happen for me any more. Your observation is correct, the shapes interfere with rendering. I'm addressing some of those in #707 as I have to touch them anyway to fit them into the new scale.

marcoow:
@pichfl so this fixes the vertical bounce and the rest will be addressed in another PR? The Percy Diff seems unintentional btw and should probably be addressed.

marcoow (changes requested):
The vertical bounce is still present on e.g. the Talks page 😞 https://deploy-preview-736--objective-northcutt-37494c.netlify.com/talks/

pichfl:
ezgif-4-a933722c87f0
This is Safari Version 13.0.2 (15608.2.30.1.1). 

I think it might be best to fix the shape rendering performance issues first.

marcoow:
I have the same version 😞 So yes, let's fix the other issue first.

PR #737, opened 2019-10-29
Blog Post: Why Companies Invest In OSS

jayjayjpg:
screencapture-localhost-4200-blog-2019-11-11-why-companies-invest-in-oss-2019-11-11-10_44_15

PR #743, opened 2019-11-08
#full and #plain images in blog posts for art direction

pichfl:
Makes it possible to have full width images and figures without a border. Necessary for #737

### DIFF
--- a/lib/generate-blog/lib/components-builder.js
+++ b/lib/generate-blog/lib/components-builder.js
@@ -203,8 +203,28 @@ function wrapNode(dom, node, className, element = 'div') {
 
 function replaceWithFigure(dom, element, content) {
   let figure = dom.createElement('figure');
+  let img = content;
+  let className = 'blog-post.figure';
 
-  figure.classList.add('blog-post.figure');
+  if (img.tagName !== 'IMG') {
+    img = content.querySelector('img');
+  }
+
+  if (img.src.endsWith('#full')) {
+    img.src = img.src.replace(/#full$/, '');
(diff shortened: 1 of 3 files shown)


lib/generate-blog/lib/components-builder.js
```diff
@@ -203,8 +203,28 @@ function wrapNode(dom, node, className, element = 'div') {
 
 function replaceWithFigure(dom, element, content) {
   let figure = dom.createElement('figure');
+  let img = content;
+  let className = 'blog-post.figure';
 
-  figure.classList.add('blog-post.figure');
+  if (img.tagName !== 'IMG') {
+    img = content.querySelector('img');
+  }
```
marcoow, line 211: Would this not be unnecessary if we changed https://github.com/simplabs/simplabs.github.io/blob/3de31b3fa9e0e7a7cf56d7f608a67442ce456416/lib/generate-blog/lib/components-builder.js#L247 to

```js
replaceWithFigure(dom, img.parentElement.parentElement, img);
```

?

pichfl:
```md
[![foo.jpg]()](#bar)
```
should be rendered as 
```html
<figure class="blog-post.figure">
    <a href="#bar">
        <img class="figure.content-centered" src="foo.jpg" />
    </a>
</figure>
```

`<a>` would get lost without that query.

marcoow: ah, nevermind - makes sense :+1:

marcoow (approved):

PR #748, opened 2019-11-13
Update og:image

arthurp1:
The current og:image link is broken. I suggest changing it to the following image: https://simplabs.com/assets/images/posts/2019-10-30-why-companies-invest-in-oss/oos-illustration-9c6ddc274018c1157eca980a3abe34c5.png


marcoow (changes requested):
The URL is actually correct but the image is in the wrong folder - [this file](https://github.com/simplabs/simplabs.github.io/blob/master/public/assets/images/posts/2019-10-30-why-companies-invest-in-oss/og-image.png) (and the other images in this post) actually need to be in  `public/assets/images/posts/2019-11-11-why-companies-invest-in-oss/`

arthurp1:
I've updated the og:image with the image and path that you suggested. It took some time to notice your comment because I didn't receive any github notifications.

pichfl (approved):

pichfl:
I force pushed a quick update to make the markdown linter happy, should be ready to be merged.

marcoow (changes requested):
See my previous review - the file name in the markdown is correct but the image is in the wrong folder. The `og:image` (and the other images used in the post) need to be moved into the correct folder (and the urls need to be updated in the markdown then for the image used in the article text).

marcoow (approved):
I fixed it

PR #749, opened 2019-11-15
Dark mode

### LINKED ISSUE #729: Support Dark Mode
Our website is very light and mostly configured by CSS variables, which makes it a perfect candidate for support dark mode.

I did a quick test to check if this is something we can do with little effort and it seems to be the case.

To enable support, we must declare the supported color schemes in the `:root` like so:
```css
:root {
  color-scheme: light dark;
}
```

and then use a media query to adjust variables and other things.
```css
@media (prefers-color-scheme: dark) {
  :root {
    --color-primary: #fff;
    --color-secondary: #a5a5a5;
    --color-accent: #007df6;
    --color-background: #222;
    --color-error: #ec2a2a;
  }
}
```

This will most likely reveal quite a few places where we used shortcuts or hardcoded color values, which should be replaced by variables anyway.

- [ ] Add color scheme
- [ ] Replace variables
- [ ] Check for hardcoded values
- [ ] Add Dark mode tests to Percy

The nice thing about this is the fact that we can use almost all modern CSS features (like clip-path) to fix issues, as the support for dark mode is limited to browsers with very good CSS support. This makes it rather straight forward to work on this, especially after #722 is completed.

pichfl:
To be merged after #722!

Closes #729


pichfl:
Postponed until the new shapes, colors and text sizes are settled. The actually necessary changes for this feature are only a few lines so we should be able to do this again when everything else is done.

PR #763, opened 2019-12-12
Move CI to github actions

marcoow:
…which might speed things up significantly. The baseline we're starting form is around 10min: https://travis-ci.org/simplabs/simplabs.github.io/builds/623700874 - let's see whether this is faster.

The Netlify deployment should eventually be handled with GitHub Actions as well maybe but that doesn't seem to be ready yet, see https://community.netlify.com/t/guide-for-deploy-using-github-actions/2981/5


marcoow:
Yep, seems like it's faster

mansona (approved):
Wow this looks great 🎉 It might even be worth a blog post in itself 😂

PR #777, opened 2019-12-16
Update shape component

### LINKED ISSUE #752: [Component] Shape Component

## What needs to be done:
We need a shape component that can be applied on all website pages. This shape component replaces all of the existing shapes.

Website__organic-shape__sketch

### Layout
The shape component spans all columns and consists of 3 parts, a top, middle and bottom part. The middle part (b) can grow vertically to fit any content.
The shape top (a) and bottom (c) parts should scale-down proportionally on smaller viewport sizes.

- [ ] Add responsive full-width shape component 

#### Layout Variations
Variations of the top and bottom shapes are possible. A variation for the footer shape will have the same layout, and replace both the top and bottom shapes.
Website__organic-shape__sketch

- [ ] Add footer variation

### Style
The shape component should be easy to customize with different colors. 

- [ ] Add blue color variation as default

ghislaineguerin:
closes #752


ghislaineguerin:
@marcoow @pichfl 
I'd like to get your suggestion on how to fix 2 things for this PR

image
I get some pixels outside of the mask as seen on the above image.

image
image
I'd like to have a slightly darker background around the edges without having to add a white background to every component and horizontal padding, unless that's the only solution.

marcoow:
@ghislaineguerin the lines on the left and right are the background of the `div` shining through - if you make the `svg` have `width: 101%;` for example they go away on the right. That wouldn't be a proper fix anyway of course - just as an explanation. Would it not be possible to have the `clip-path` on the `div` directly btw so we wouldn't need the SVG at all?

ghislaineguerin:
@marcoow I'll try with clip path, thanks.

ghislaineguerin:
localhost_4200_ (1)
localhost_4200_ (2)

@marcoow I've tried clip-path and path is not supported yet, so I used polygon, problem is we cannot maintain aspect ratio. We'll have to go with the svg mask.

marcoow:
@ghislaineguerin what do you mean `clip-path` is not supported? It should be relatively widely supported: https://caniuse.com/#search=clip-path

ghislaineguerin:
@marcoow I mean clip-path: path(), I've used polygon, but I think that's not what causes the aspect ratio issues, because the div would still need a set height, which is not the case for using svg.

marcoow:
Hm, the SVG has a fixed height as well though: https://github.com/simplabs/simplabs.github.io/pull/777/files#diff-9d15c7be363a433981215e4e283ae6faR6

ghislaineguerin:
@marcoow I've found a solution, will push in a moment.

ghislaineguerin:
@marcoow I've pushed the clip-path solution, basically I've used the width to calculate the proportional height so it works now. However, because of anti-aliasing I do have to add a 1px negative margin, otherwise we see a gap at some resolutions.

src/ui/components/ShapeFeature/template.hbs
```diff
@@ -1,52 +1,18 @@
...
-        ></path>
-        <mask id={{this.maskId}}>
-          <rect x="0" y="0" width="100%" height="100%" fill="white"></rect>
-          <use x="10" y="10" xlink:href="#{{this.shapeId}}" fill="black"></use>
-        </mask>
-      </defs>
-      <rect fill="#fff" x="0" y="0" width="100%" height="100%" mask="url(#{{this.maskId}})"></rect>
-    </svg>
+  <div class="top"></div>
+  <!--bottom-->
```
marcoow: Let's not add these comments - they don't add value really I think?

marcoow (changes requested):
The solution using `clip-path` makes things a lot simpler I think so that's 👍 Also I agree the new shape works better. A lot of the changes in the CSS seem unrelated though as they are only introducing newlines, reorder things etc.? Can those be reverted as they add a lot of noise to this PR and make it really hard to understand what is an actual change and what is not.

Also, I think the image on the trainline shape looks much better in Abstract:

Bildschirmfoto 2019-12-17 um 15 45 21

than it does on the page with this branch:

Bildschirmfoto 2019-12-17 um 15 45 44

Also I think the other client logos should be part of the shape now?

ghislaineguerin:
@marcoow 
I've fixed the image, but unfortunately the shape I had in abstract doesn't work very well when trying to keep the proportions balanced, seems we need to have straight edges. 
As for the row of logos, I've tried and it doesn't look too good, I will modify the abstract design to match.

src/ui/components/ShapeFeature/template.hbs
```diff
@@ -1,52 +1,18 @@
...
-      <div class="img-wrapper">
-        <img class="img" src={{@src}} alt="{{@srcAlt}}" width={{@width}} height={{@height}} />
-      </div>
-    {{/if}}
-    <figcaption class="content">
-      {{yield}}
-    </figcaption>
-  </figure>
+  <div class="bottom"></div>
+  <!--top-->
```
marcoow: Can we remove these comments please?

src/ui/components/ShapeFeature/template.hbs
```diff
@@ -1,52 +1,18 @@
 <div style={{this.shapeStyle}} state:hasImage={{@src}}>
```
marcoow: Is the `shapeStyle` style still necessary?

src/ui/components/ShapeFeature/template.hbs
```diff
@@ -1,52 +1,18 @@
...
-        <img class="img" src={{@src}} alt="{{@srcAlt}}" width={{@width}} height={{@height}} />
-      </div>
-    {{/if}}
-    <figcaption class="content">
-      {{yield}}
-    </figcaption>
-  </figure>
+  <div class="bottom"></div>
+  <!--top-->
 </div>
```
marcoow, line 16: I think the `component.ts` for this component can probably be deleted.

marcoow (changes requested):
I think this looks good in general but I found a few glitches:

Bildschirmfoto 2019-12-20 um 15 19 04
Bildschirmfoto 2019-12-20 um 15 19 10
Bildschirmfoto 2019-12-20 um 15 19 23

Also I'm not sure we want the margin around the shape on mobile?

localhost_4200_(iPhone 4) (1)

ghislaineguerin:
@marcoow I've removed the negative margins that caused the glitches but there's an antialias issue causing a white line between shapes, I am investigating a fix for that.

 Left: Safari / Right: Chrome
image

ghislaineguerin:
@marcoow the only way I've found to fix this is by scaling up the shapes enough to cover the gap, however I don't think this is a sustainable approach. Only option left I think is too go back to svg.

marcoow:
@ghislaineguerin I think the problem with the white lines is that between the `.top`, `.middle` and `.bottom` `div`s is that we're using a relative `padding-bottom` (of `15%` in this example) which can result in fractions (as opposed to full numbers), e.g. `137.578`. At least Chrome supports sub-pixel rendering and will correctly render that as not a full pixel so that the white line appears.

ghislaineguerin:
@marcoow thanks, will try that.

ghislaineguerin:
@marcoow We'll have to use svg then, it seems there's no way to maintain the div's aspect ratio without percentages :(

marcoow:
Hm, let's look at that together tomorrow - there should be some sort of solution.

ghislaineguerin:
@marcoow 
I've added 90 deg angles to the shapes so that I can overlap without breaking the edges.

image

marcoow:
@ghislaineguerin now we have the same problem that we had with the relative height before with the relative width around the left and right edges:

Bildschirmfoto 2020-01-15 um 13 56 02

ghislaineguerin:
@marcoow last change looks great. Thanks!

ghislaineguerin:
@marcoow I've removed the typography changes. Please review.

lib/global-css/css/baseline.css
```diff
@@ -221,7 +217,6 @@ h5 {
 @media (max-width: 887px) {
   h1 small {
     font-size: 2.5rem;
-    color: var(--color-text);
   }
 
   h2 {
```
marcoow: Are the changes in this file actually intentional? I'd assume they change everything on the entire site?

marcoow (dismissed):
I think this looks good in general but I found 2 small glitches:

* text has wrong color:

Bildschirmfoto 2020-01-17 um 13 47 43

* edges are sort of parallel but not exactly:

Bildschirmfoto 2020-01-17 um 13 47 36

marcoow:
@ghislaineguerin I rebased this on the latest `website-revamp` but there need to be 2 responsive versions of the new JPG now

lib/global-css/css/baseline.css
ghislaineguerin: @marcoow I have fixed this now.

marcoow:
@ghislaineguerin I rebased on `website-revamp` again and this is now also running the Percy check.

ghislaineguerin:
@marcoow looking at percy it seems this affects some other instances of this same shape used in other pages. Do you think those should be fixed as part of this PR?

marcoow:
Yes, we should not merge this unless it either does not affect these other shapes or correctly changes the other shapes.

ghislaineguerin:
@marcoow I have fixed the other shape instances. Please review.

locks (approved):
Checked locally and everything seemed fine 👍

PR #796, opened 2019-12-20
add missing closing }

marcoow:
…that was missing in the last blog post. While that's not really a problem as it's only in the code example that is never run anyway it should still be fixed of course.


locks (approved):
thank you for spotting it!

locks:
I don't have permissions for percy, so I cannot approve it there.

PR #814, opened 2020-01-15
generalize talks section

marcoow:
This generalizes the talks section of the website so it can cover more event and media types than just conferences and talk videos but for example also appearances in podcasts.

* renames the `conference.md` files in the `_talks` directory to `channel.md`
* renames references to `conferences` in the code to `channel`
* adds `isVideo` and `isPodcast` for each talk so we can render different link texts
* rename `_talks` folder to `_appearances` and each `talks` folder for each channel to `appearances`
* renames references to `talks` in the code to `appearances`


lib/generate-appearances-archive/lib/templates/page/template.hbs
```diff
@@ -1,23 +1,23 @@
 <div>
   <Header
     @title="Talks"
-    @description="We share our expertise and experience with others. Here are a collection of talks that our team has given at international conferences in the past years."
+    @description="We share our expertise and experience with others. Here are a collection of appeances that our team made at international events in the past years."
```
ghislaineguerin: @marcoow Typo here: appeances should be appearances

ghislaineguerin (changes requested):

lib/generate-appearances-archive/lib/templates/page/template.hbs
marcoow: fixed 👍

ghislaineguerin (approved):

PR #817, opened 2020-01-15
Make the page a proper PWA

### LINKED ISSUE #701: Make this a proper PWA
Although the page uses a service worker and fully works offline, it is not a proper PWA as it e.g. lacks a manifest. We should fix that, even if it's only for the sake of reaching a better Lighthouse score (nobody will likely actually install our site on their home screens).

### TODO

- [x] add a proper manifest file (see e.g. https://github.com/simplabs/breethe-client/blob/master/public/manifest.webmanifest for reference)
- [x] define icons and browser colors (see https://developers.google.com/web/fundamentals/design-and-ux/browser-customization/?utm_source=lighthouse&utm_medium=devtools for reference)

marcoow:
This converts the page into a proper PWA, mostly in order to get a better/perfect Lighthouse score. 

This

* adds a web manifest
* adds app icons
* adds a meta tag for the theme color

All of the above is done in an in-repo-addon.

closes #701


lib/pwa/index.js
```diff
@@ -0,0 +1,40 @@
...
+    });
+    let imagesTree = Funnel(filesDir, {
+      srcDir: 'app-images',
+      destDir: 'assets/images/app-images'
+    });
+
+    return MergeTrees([manifestTree, imagesTree]);
+  },
+};
+
```
locks:
```suggestion
```

locks (approved):

PR #828, opened 2020-01-21
update footer component

### LINKED ISSUE #820: [Components] Update footer component

## What needs to be done:
We need a new footer design to match the new shape design. The new footer will also include content and styling updates.

image

ghislaineguerin:
Closes #820


ghislaineguerin:
image

@marcoow I'm struggling to find a good solution to this gap between bottom shape and the image. I cannot make the entire footer container a clip path as the proportions won't be kept but then overlapping the image seems to not work for some reason. Ideally I wouldn't want to use absolute positioning.

marcoow:
@ghislaineguerin I'd guess you'd have to have the image as a background image on the wrapper div that wraps the `.top` etc. diffs and have the `clip-path` on that as well - I cannot really think of an alternative.

ghislaineguerin:
We can work on #839 as part of post-launch improvements

src/ui/components/Footer/template.hbs
```diff
@@ -1,128 +1,132 @@
...
-      </span>
-    </p>
-  </div>
-  <nav class="nav">
-    <div class="block">
-      <h3 class="title">
-        <a href="/services/" data-internal>
-          Services
+<div>
+  <div class="container layout">
```
marcoow: Is it possible to not have these wrapper `div`s and make the `footer` the `.container.layout`?

src/ui/components/WorkWithUs/stylesheet.css
```diff
@@ -1,18 +1,71 @@
...
+  grid-column: 1/-1;
+
   position: relative;
   z-index: 0;
-  block-name: WorkWithUs;
-  composes: 'offset.after-21';
+}
+
+.top {
+  background: #f8f8fa;
```
marcoow: Should we use CSS variables for the colors used in this file?

marcoow (changes requested):
This looks good in general - just left some comments on smaller things.

Also there are 2 visual things we should fix:

* let's not use my image in all of the footers but images of different people
* the image is cut off on the sides for medium viewport sizes:

Bildschirmfoto 2020-01-27 um 10 40 02

src/ui/components/Footer/template.hbs
ghislaineguerin: @marcoow I'll try that.

ghislaineguerin: @marcoow I get this error: Cannot put Block classes on the Block's root element.

ghislaineguerin:
@marcoow is there a way we can randomize the team member that shows on the footer? Or set a default when nothing is defined.
I've added a variation in the services page.

src/ui/components/Footer/template.hbs
marcoow: I see  - can we somehow apply the classes to the root element using any of the css-blocks mechanisms like `@composes` or so? I'm just wondering whether we can somehow get the same result without having to wrap the `<footer>`

marcoow:
We cannot randomize the team member (that has to do with the static pre-rendering where we must make sure the pre-render results in the same DOM that we get when the app boots up in the browser). I'd envision an API like

```hbs
<WorkWithUs @photo="ghislaine" />
```

where we can just pick who we want to show when invoking the component.

locks (approved):

src/ui/components/PageCaseDdWrt/template.hbs
```diff
@@ -63,6 +63,12 @@
     </p>
   </div>
   <ShapeFeatureTrainline @label="Continue Reading" />
-  <WorkWithUs />
+  <WorkWithUs
+    @color="#007DF6"
+    @srcset="/assets/images/photos/marco@604.png 604w, /assets/images/photos/marco@1208.png 1208w"
+    @sizes="(max-width: 887px) 604px, 1208px"
+    @srcAlt="Marco"
+   />
```
marcoow: Can we move these values into the component so we don't have to repeat so many details over and over? If we can invoke the component just with

```hbs
<WorkWithUs @teamMember="ghislaine"/>
```

I think that would be much easier.

src/ui/components/PageCaseDdWrt/template.hbs
```diff
@@ -63,6 +63,12 @@
     </p>
   </div>
   <ShapeFeatureTrainline @label="Continue Reading" />
-  <WorkWithUs />
+  <WorkWithUs
+    @color="#007DF6"
```
marcoow: Is this not static and thus wouldn't have to be passed like this?

marcoow (changes requested):
The images are now cut off on large viewports as well which I think looks weird:

Bildschirmfoto 2020-01-29 um 14 27 59
Bildschirmfoto 2020-01-29 um 14 28 09

Also, do we have photos of more people than you, Chris and me so we could have more variety?

src/ui/components/PageCaseDdWrt/template.hbs
ghislaineguerin: @marcoow so that means I can do something like  photos/@teamMember@604.png  ? how would this work?

src/ui/components/PageCaseDdWrt/template.hbs
ghislaineguerin: @marcoow I don't know why color is being passed there. I will remove it.

src/ui/components/PageCaseDdWrt/template.hbs
marcoow: You would have a property in the component's `.ts` file that maps short names to the actual values for `srcset` etc. - something like

```typescript
srcSets: {
  ghislaine: '/assets/images/photos/ghislaine@604.png 604w, /assets/images/photos/ghislaine@1208.png 1208w'
},

srcSizes: {
  ghislaine: '(max-width: 887px) 604px, 1208px'
},

imageAlts: {
  …
}
```

@locks will be able to help set that up if you're having problems with that.

ghislaineguerin:
@marcoow most didn't do the individual portraits. I only have Jessica, but the picture with the right aspect ratio are w/ eyes closed. We need to plan to take more during winter event.

marcoow:
@ghislaineguerin ok, we should be able to take pictures when we all meet in the hotel at 19:00 - can you bring/prepare everything for that?

src/ui/components/WorkWithUs/stylesheet.css
marcoow: @ghislaineguerin this is still open

src/ui/components/WorkWithUs/template.hbs
```diff
@@ -1,11 +1,20 @@
...
+      </h3>
+      <p class="typography.lead offset.after-12">
+        Talk to one of our experts to find out how we can help you.
+      </p>
+      <a class="button" href="/contact/" data-internal>
+        Let's discuss your project 
+      </a>
+    </div>
+    <div class="img-wrapper">
+      <img class="img" srcset="/assets/images/photos/{{@teamMember}}@604.png 604w, /assets/images/photos/{{@teamMember}}@1208.png 1208w" sizes="(max-width: 887px) 604px, 1208px" alt="{{@teamMember}}" width={{@width}} height={{@height}} />
```
marcoow: I think we should capitalize the `alt` text which means we need a property on the component context (the `.ts` file).

marcoow: Also, this only works when all of the team pictures actually have same sizes (604px width and 1208px width) - is that certainly going to be the case?

marcoow (changes requested):
Hm, I think some things still need to be done:

* there are 4 different images for Chris and Ghislaine while the component will only ever use 2
* the images are huge, some almost 2MB
* none of the images actually render, e.g.:

Bildschirmfoto 2020-01-30 um 12 16 50

src/ui/components/WorkWithUs/template.hbs
ghislaineguerin: @marcoow I'd need help with this part. I'm not sure how to make this component reusable. I'll fix all the other issues.

marcoow: @locks can you help?

locks: working on it!

ghislaineguerin:
@locks I have fixed some styling issues and removed some redundant images. However some tests are failing, please review.

marcoow (changes requested):
- Some of the mobile views are broken - I left comments on Percy.
- Also can we get the team photos smaller? They are still pretty massive with ca. 800KB for some of them - I'd say none of them should be above 150KB or so.
- there's still and unaddressed comment from a previous review

src/ui/components/WorkWithUs/template.hbs
locks: `alt` is now capitalized.

ghislaineguerin:
@marcoow It seems we don't really need the large pngs as the image doesn't grow beyond 604px. Please review. I also can't reproduce the mobile bugs from Percy. We have changed the service pages so I'm not sure it's worth fixing now?

src/ui/components/WorkWithUs/template.hbs
```diff
@@ -1,11 +1,28 @@
...
+        Talk to one of our experts to find out how we can help you.
+      </p>
+      <a class="button" href="/contact/" data-internal>
+        Let's discuss your project
+      </a>
+    </div>
+    <div class="img-wrapper">
+      <img
+        class="img"
+        srcset="{{this.teamMemberPhoto}}"
```
marcoow: We can just use a regular `src` without `srcset` and `sizes`.

marcoow (commented):
@ghislaineguerin I cannot reproduce the broken mobile pages on Percy neither - approved. There's one thing we can optimize in the code though (see comment). Also, regarding the images, I suppose we can use JPGs with a grey background now that the photos never extend beyond the grey container anyway which should shrink the images quite significantly?

src/ui/components/WorkWithUs/template.hbs
locks: Link directly to `@604px` then? Updating.

ghislaineguerin:
@marcoow I'd like to keep the transparency of the images as I still might adjust colors/shade before release.

marcoow (approved):

marcoow:
@ghislaineguerin ok, sure let's keep the PNGs then. Only the failing test preventing this from merging then

PR #860, opened 2020-02-04
Update page header

### LINKED ISSUE #812: [Component] Add header variation for service pages

## What needs to be done:
Create a variation for the header component that allows the new design to be implemented
image

- Should allow for a full-width wrapper with different background color 
- Should include polygon shape at the bottom

ghislaineguerin:
This adds the new header styles as introduced with the updated services pages, e.g.:

localhost_4200_services_

localhost_4200_services_ (1)

It also changes the mobile menu so it blends in with the new header shapes better:

localhost_4200_services_(Pixel 2 XL)

### TODO

- [x] add mobile versions
- [x] we are currently passing `@image` but we'll want to use `@srcset` and `@sizes` probably (unless we decide we don't need the images at all for mobile)
- [x] add in the real graphics we want to use (currently using the placeholder on all of the pages)

closes #812


ghislaineguerin:
@marcoow I need help with the following:
I need to create a variation for header content when it doesn't have the image w/ background, where content would take the entire width. See example below. I also need the same image attribute like we had for the footer, where the image will change on each page.

updated abstract for reference: https://share.goabstract.com/b69a3eca-2239-4585-b2f4-d7c81e52c76d?collectionLayerId=17ec6de9-653e-4e52-9426-c012b1e6b6a2&mode=design

image
BROKEN HOMEPAGE:
image

marcoow:
@ghislaineguerin I'm removing the clipped images from this branch and put them into their own branch as they are unrelated to the updated header.

ghislaineguerin:
@marcoow I will take this PR and add the mobile styles

ghislaineguerin:
image
@marcoow Is this the variation? I see no gray background yet. If we haven't done so yet, I'd prefer to drop it and have shapes just bounded to the grid.

ghislaineguerin:
@marcoow where did you get the images from? Last version for digital products should be the same that is here https://github.com/simplabs/simplabs.github.io/issues/886

ghislaineguerin:
Can we try reducing the width of the blue shape
and extending the image wrapper and using a negative left margin w the offset value

UPDATED
image

ACTUAL
image

marcoow:
@ghislaineguerin the image for digital products is the one we're already using: https://github.com/simplabs/simplabs.github.io/blob/master/public/assets/images/photos/coding%401800.jpg – we should try and avoid having to add another version of that same image.

marcoow:
I'll see what I can do with respect to the width of the sidebar etc. – we're kind of bound by our grid layout I think and we should probably not add too much variation there and break out of that layout?

locks (approved):
Manually checked the headers in the Netlify preview for desktop and mobile viewports and they look good!

marcoow:
@ghislaineguerin I changed the width of the blue ribbon version – please check again

locks (approved):

PR #873, opened 2020-02-11
Update css-blocks

marcoow:
This updates `@css-blocks/ember-cli` to the latest version and adapts the styles accordingly

* classes from external blocks are now applied like `typography:class="lead"` instead of `class="typography.lead"`
* classes from the component's block are now applied `block:class="some"` instead of `class="some"`
* a block's `:scope` styles is now applied like `media:scope` vs. `class="media"`
* external blocks need to be imported and exported from the component's own block so they can be used in the template
* the inability to use the `class` attribute directly required changes in the DOM manipulation code that we run on blog posts as we now have to `element.setAttribute('block:class', 'some')` instead of `element.classList.add('some')`

closes #836


locks (approved):

locks:
Where did you find the `media:scope` syntax? I searched the documentation and the source code and couldn't find a mention.

marcoow:
@locks the test fixtures are a pretty good source of examples: https://github.com/linkedin/css-blocks/blob/master/packages/%40css-blocks/glimmer/test/fixtures/styled-app/src/ui/components/with-dynamic-classes/template.hbs

locks:
Should the PR target the website revamp instead?

marcoow:
No this should go to master - why should it target the other branch?

locks:
I was thinking that with all the redesign and new pages we might be doubling the effort, but I'll make sure the redesign branch is updated to latest cssblocks as well.

marcoow:
The `website-revamp` branch is already big enough - let's not complicate things further by adding more weight to it.

PR #895, opened 2020-02-17
Cleanup shape feature

### LINKED ISSUE #884: Cleanup ShapeFeature component
The `ShapeFeature` component has some unused states and HTML that's not ideal and we should clean that up.

- [x] remove unused `hasImage` state (see https://github.com/simplabs/simplabs.github.io/blob/website-revamp/src/ui/components/ShapeFeature/template.hbs#L1)
- [x] if the component is used without a `@srcset` we should **not** be using `figure` and `figcaption` elements but `div`/`p`

marcoow:
This cleans up the `ShapeFeature` component a bit by:

* removing the unused `hasImage` state
* for the case that no `@srcset` is provided (thus no image is shown), using `div`/`p` tags instead of `figure`/`figcaption`

closes #884

### DIFF
--- a/src/ui/components/ShapeFeature/component-test.ts
+++ b/src/ui/components/ShapeFeature/component-test.ts
@@ -12,4 +12,22 @@ module('Component: ShapeFeature', function(hooks) {
 
     assert.ok(this.containerElement.querySelector('div'));
   });
+
+  module('when invoked with a @srcset', function() {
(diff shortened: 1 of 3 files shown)


marcoow:
This was previously reviewed but I had accidentally merged to the wrong branch. I'm merging this without repeated review.

PR #922, opened 2020-03-04
Implement better crawler

marcoow:
Our current crawler job fails randomly and regularly which obviously negatively affects productivity. This switches to a different package (https://www.npmjs.com/package/broken-link-checker) which results in a certainly much faster implementation and hopefully in one with fewer random errors as well.


locks (approved):

locks:
This is a great quality of life improvement! The random crawl failures were annoying.

PR #937, opened 2020-03-06
Fix uneven layout

### LINKED ISSUE #936: Website-revamp has horizontally uneven layout
The `website-revamp` branch has an uneven layout with different whitespace on the left and right of the page:

Bildschirmfoto 2020-03-06 um 16 40 34

We need to fix that while making sure the fix for it does not break anything else that might be depending on the bug already.

marcoow:
This fixes the horizontally uneven layout of the website revamp:

Bildschirmfoto 2020-03-06 um 17 00 32

Currently there is more whitespace to the left than to the right. This PR fixes that by reverting the [layout changes made in the `website-revamp` branch](https://github.com/simplabs/simplabs.github.io/commit/b44fe791c67247a991b78ebd19e830ff6e15ba6e). It also fixes the alignment of the client logos which were bleeding out of the main content area – this is likely broken on the current production page already but not as visible since there is a different kind of shape above the logos which bleeds until the sides of the viewport anyway.

closes #936

### DIFF
--- a/lib/global-css/css/baseline.css
+++ b/lib/global-css/css/baseline.css
@@ -2,8 +2,8 @@
   --base-scale: calc(6 / 8);
 
   /* Sizing */
-  --col: calc(128 / 1440 * 100vw);
-  --col-gap: 8rem;
+  --col: calc(156 / 1440 * 100vw);
+  --col-gap: 3rem;
   --max-width: calc(var(--col) * 6 + var(--col-gap) * 5);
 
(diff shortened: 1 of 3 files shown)


locks:
Did you get the same behaviour on different browsers? I just checked the website-revamp preview deploy and I cannot reproduce the unevenness.

marcoow:
You might have to check with different viewports

locks (approved):

PR #950, opened 2020-03-10
Revert "Remove open source block from product development page"

locks:
Reverts simplabs/simplabs.github.io#935.

It seems like the designs were not up to date with the latest marketing content and this block should not have been removed.

### DIFF
--- a/src/ui/components/PageProductDevelopment/template.hbs
+++ b/src/ui/components/PageProductDevelopment/template.hbs
@@ -91,7 +91,7 @@
       Learn how we work
     </ArrowLink>
   </div>
-  <div layout:class="full" offset:class="after-12">
+  <div layout:class="full" offset:class="after-21">
     <h2>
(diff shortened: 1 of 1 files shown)


src/ui/components/PageProductDevelopment/template.hbs
```diff
@@ -91,6 +91,14 @@
       Learn how we work
     </ArrowLink>
   </div>
+  <div layout:class="full" offset:class="after-12">
```
marcoow: Should this not be `offset:class="after-21"`?

locks: The design doc does not have this content so I can't confirm. cc @ghislaineguerin

marcoow: I **think** we always use `after-21` between any major sections of the page.

marcoow (approved):

PR #954, opened 2020-03-10
Update product design page header

### LINKED ISSUE #951: Update Digital Product Design header image
Should be using the following image to match the abstract design

design-system@1600

locks:
Updates product design page header with correct image.
Since the image is much taller than it is wider, I introduced a CSSBlocks state to switch the width and height settings in a way that didn't impact the remaining headers.

Current look:
image

Closes https://github.com/simplabs/simplabs.github.io/issues/951.


marcoow (changes requested):
If the image has the wrong proportions can we please just resize it so it fits better and keep things simple?

locks:
@marcoow comment addressed

marcoow (changes requested):
Yep, looks good to me but please don't name the file `screenshot.jpg` 🙏 The name of the previous file `ember-design-system.jpg` had a bit more meaning so you could just replace that file. Also if you're not replacing that file, that means it's now just going to be in the repo unused…

locks:
Image replaced as mentioned.

marcoow (changes requested):
the image is now broken since you changed the name of the file back but left the updated invocation in:

Bildschirmfoto 2020-03-10 um 20 17 19

marcoow (approved):

PR #957, opened 2020-03-10
Tweak offsets for digital products page sections

### LINKED ISSUE #958: Tweak offsets in digital products page
The clipped images work was not fully completed as some of the offsets are not correct and need updating.
Blocks that separate two sections should use the offset `after-21`, these include `bleed-right` blocks since on mobile viewport they will be at the end of a section.

## Tasks
- [ ] Fix offsets in digital products page
  - [ ] https://github.com/simplabs/simplabs.github.io/pull/898/files#r390551275
  - [ ] https://github.com/simplabs/simplabs.github.io/pull/898/files#r390551466
  - [ ] https://github.com/simplabs/simplabs.github.io/pull/898/files#diff-e361308ad4d1619c58e247efca1bfaeeR137
- [ ] Use `after-16` for `bleed-left` and `split-leading` blocks
- [ ] Use `after-21` for `bleed-right` and `split-trailing` blocks

## Resources
- Abstract: https://app.abstract.com/projects/f6401950-547a-11e9-a66d-7f6f3414ea33/branches/38b9d501-75be-4d7d-b595-73b0b6d24f09/collections/e312bf18-d9c5-4f10-853b-2f2acbe2e4a1

locks:
Fixes offsets for clipped images blocks.

- Made `bleed-left` and `split-leading` blocks have an offset of `after-16`
- Made `bleed-right` and `split-trailing` blocks have an offset of `after-21`

Closes https://github.com/simplabs/simplabs.github.io/issues/958.

### DIFF
--- a/src/ui/components/PageDigitalProducts/template.hbs
+++ b/src/ui/components/PageDigitalProducts/template.hbs
@@ -39,7 +39,7 @@
       @align="left"
     />
   </div>
-  <div layout:class="split-trailing" offset:class="after-16">
+  <div layout:class="split-trailing" offset:class="after-21">
     <ul list:class="vertical">
       <li list:class="vertical-item">
         Market validation
@@ -75,7 +75,7 @@
       Great product design combined with outspoken aesthetic elements will make your new digital product remarkable and unique, as well as drive activation and engagement. Great user experiences can position your product uniquely in the market, and win both raving fans and loyal audiences.
     </p>
   </div>
-  <div layout:class="split-leading" offset:class="after-21">
+  <div layout:class="split-leading" offset:class="after-16">
     <ul list:class="vertical">
       <li list:class="vertical-item">
         Wireframing & prototyping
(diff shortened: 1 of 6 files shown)


marcoow (commented):
This seems good, I'm just not sure I understand the reasoning behind this text from the issue:

>  Use after-16 for bleed-left and split-leading blocks
>  Use after-21 for bleed-right and split-trailing blocks

Can you explain why both are treated differently?

locks:
@marcoow the reasoning is that `bleed-left` and `split-leading` are on the left side when on desktop viewport, and on top when on mobile viewport. This means that the `bleed-right` and

When the elements are side by side on the desktop viewport, both of them separate the section from the following one:
image

But on the mobile viewport, only the right-hand side element separates the section from the following one, since the left-hand side is followed by the right-hand side of the same section:
image

Does this make sense or am I approaching it wrong?

marcoow:
Yes, makes sense – thanks for clarifying!

marcoow (approved):
Can you rebase and merge?

PR #959, opened 2020-03-10
Extract clipped image shapes into ShapeImage component

locks:
As a cleanup, I extracted the clipped images left and right-aligned shapes into a ShapeImage component. This implementation is using the cssblocks state feature to define which side the image should align to.
This aligns with what we do with other shapes, help prevent copy-paste mistakes, and allow changes to one location to spread to all usages of the component.

There should be no visual changes.

### DIFF
--- a/src/ui/components/ShapeImage/stylesheet.css
+++ b/src/ui/components/ShapeImage/stylesheet.css
@@ -1,0 +1,23 @@
+@block fluid-image from "../../styles/blocks/fluid-image.block.css";
+
+:scope {
+  block-name: ShapeImage;
+
+  position: relative;
+  background-color: #ccc;
+  width: 100%;
+  height: auto;
+  padding-bottom: 100%;
+}
+
+:scope[align='left'] {
+  -webkit-clip-path: polygon(100% 20%, 100% 80%, 30% 100%, 0 95%, 0 0);
+  clip-path: polygon(100% 20%, 100% 80%, 30% 100%, 0 95%, 0 0);
+}
+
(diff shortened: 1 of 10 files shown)


src/ui/components/ShapeImage/stylesheet.css
```diff
@@ -0,0 +1,23 @@
...
+  -webkit-clip-path: polygon(100% 20%, 100% 80%, 30% 100%, 0 95%, 0 0);
+  clip-path: polygon(100% 20%, 100% 80%, 30% 100%, 0 95%, 0 0);
+}
+
+:scope[align='right'] {
+  -webkit-clip-path: polygon(80% 0, 100% 5%, 100% 100%, 0 80%, 0 20%);
+  clip-path: polygon(80% 0, 100% 5%, 100% 100%, 0 80%, 0 20%);
+}
+
+@export fluid-image;
```
marcoow, line 23: I assume the styles added here should be removed elsewhere (specifically `fluid-image.clipped-right` and `fluid-image.clipped-left`)?

marcoow (commented):
Makes sense 👍 I think there's a cleanup step missing though.

src/ui/components/ShapeImage/stylesheet.css
locks: Removed `clipped-right` and `clipped-left`, but renamed `clipped-image` to `image-clipped` for consistency with the existing `image` and `image-cover` classes.

marcoow (approved):

PR #968, opened 2020-03-11
Replace Trainline quote with Qonto quote

### LINKED ISSUE #963: Replace Trainline quote with Qonto quote
We just added the Trainline quote but I realized we have a pretty good one from Qonto that we could use instead since we're referring to Trainline in quite a few other places already. We can simply replace the `<ShapeQuoteTrainline>` component with a `<ShapeQuoteQonto>` component and update all usages of the component respectively.

### TODO

- [ ] create image for the quote component
- [ ] rename `<ShapeQuoteTrainline>` to `<ShapeQuoteQonto>` and update usages
- [ ] replace quote from Carl with the quote from Marc-Antoine below

### Resources

* Quote: _"simplabs are well known as the Ember.js experts and they absolutely live up to the expectations. They had an immediate as well as significant positive impact on both our velocity and quality of output. - Marc-Antoine Lacroix, Qonto CTO"_
* Image: we could use something from their [media kit](https://projects.invisionapp.com/boards/9T3UDRXEHSN) maybe?

locks:
This PR replaces the Trainline quote with a Qonto one, updating the copy, the author, and the image, as well as the component name in the code.

Closes https://github.com/simplabs/simplabs.github.io/issues/963.


locks:
As mentioned on Slack, the resized and optimized PNGs for the quote are 85KB and 356KB which is on the heavy side. I tried converting to JPGs but they looked much darker and somewhat blurry.

locks:
Image updated to provided JPGs

marcoow (changes requested):
As said on Slack, we should not actually be showing the mobile app as that is a native app and we've not actually been involved with that. I just realized that now, sorry 😞

marcoow (approved):

PR #975, opened 2020-03-11
center left/right bleeding content vertically

### LINKED ISSUE #974: Center ShapeImage vertically next to content
We should center the images that are bleeding left/right next to the content vertically so they do not appear next to the content at the top, leaving a bunch of whitespace below them:

Bildschirmfoto 2020-03-11 um 12 01 27

@marcoow can we center them with the content? It will look much more balanced. I'm not sure we can replicate abstract completely as the clips are single polygons and not top/middle/bottom components like in sketch.

_Originally posted by @ghislaineguerin in https://github.com/simplabs/simplabs.github.io/issues/960#issuecomment-597564268_

marcoow:
This centers the content of `bleed-left` and `bleed-right` columns vertically.

### Before 

Bildschirmfoto 2020-03-11 um 12 20 34

### After

Bildschirmfoto 2020-03-11 um 12 19 59

closes #974

### DIFF
--- a/src/ui/styles/blocks/layout.block.css
+++ b/src/ui/styles/blocks/layout.block.css
@@ -58,6 +58,12 @@
     grid-column: 1/-6;
   }
 
+  .bleed-left,
+  .bleed-right {
+    display: flex;
+    align-items: center;
+  }
+
   .content {
     grid-column: 3/-3;
   }


locks (approved):
Looks really nice!

PR #1026, opened 2020-03-27
Update Page titles and Meta Descriptions

arthurp1:


arthurp1:
I expect that I'm going to make a bunch of mistakes, like forgetting to run prettier, having commit messages that are not clear or commits that are too big. I don't want to waste your and my time.

PR #1027, opened 2020-03-27
Make `yarn start` render limit content archive by default

locks:
Since most times when we are working on the website we do not need
the full content archive to be rendered, either because we are working
on a specific page or on a blog post that should be within the range
of the limited content archive, I think making `yarn start` is a quality
of life improvement.
I replaced `start-dev` with a `start:production` script for when
the full content archive is required.

Let me know if I am missing something in our pre-rendering pipeline
that might be affected by this change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,8 @@
   "scripts": {
     "build": "PRERENDER=1 ember build -prod && node scripts/prerender.js",
     "crawl": "node scripts/crawl.js",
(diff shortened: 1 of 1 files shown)


marcoow:
Hm, I don't think this is great atm because there are some problems with `start-dev`, e.g. it doesn't correctly build the component with the most recent posts for the Elixir page so that that page fails to render.

PR #1043, opened 2020-04-07
Extract ShapeFeatureContent component

marcoow:
This extracts a `ShapeFeatureContent` component out of `ShapeFeature` so that we can use that parent component in a more flexible way, in particular with different kinds of content than what is in `ShapeFeatureContent` now.

### DIFF
--- a/src/ui/components/ShapeFeature/component-test.ts
+++ b/src/ui/components/ShapeFeature/component-test.ts
@@ -12,4 +12,10 @@ module('Component: ShapeFeature', function(hooks) {
 
     assert.ok(this.containerElement.querySelector('div'));
   });
+
+  test('it renders with a block', async function(assert) {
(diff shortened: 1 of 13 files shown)


src/ui/components/ShapeFeature/component-test.ts
```diff
@@ -12,22 +12,4 @@ module('Component: ShapeFeature', function(hooks) {
```
jayjayjpg: Does it make sense to update this test case with an example of a component invocation in which content is yielded now?

jayjayjpg (approved):
This seems great! I left one more suggestion in regards to tests - feel free to take it or leave it according to your own opinion

src/ui/components/ShapeFeature/component-test.ts
marcoow: sure, added

PR #1046, opened 2020-04-08
Wrap markdown files after 80 characters

marcoow:
This makes the files (in particular blog posts) more uniform in style and – which is more important in my opinion even – makes reviews much nicer. With super long lines, it's very hard to comment on a particular part of a sentences when that is located on character 500 of the respective line. Also using Github's suggestions is very hard with these long lines. With everything wrapped after 80 characters that all becomes much simpler.


Turbo87 (approved):
very much in favor of this, though it looks like CI is failing 🤔

PR #1180, opened 2020-06-26
Add "Testing the Mirage.js setup" blog post

### LINKED ISSUE #716: `assert.matchJson()` assertion
based on https://github.com/ozkxr/match-json to assert against Mirage setup responses

Turbo87:
Resolves https://github.com/simplabs/simplabs.github.io/issues/716


_posts/2020-06-23-testing-mirage.md
```diff
@@ -0,0 +1,198 @@
...
+[Mirage.js] is a universal library to mock out HTTP-based APIs. It has proven
+quite useful to us in several client projects, where it helped us write
+a lot of acceptance tests in a concise, but flexible manner.
+
+The issue with tools like this is that you are not testing "the real API"
+though. This is where end-to-end tests are useful, but since those kinds of
+tests are quite slow and complex it would be quite costly to use them for all
+the kinds of tests in a modern web application.
+
+The solution: test the mock API. In this blog post we will show you how we
```
marcoow: There might something be missing between this and the previous paragraph. In the previous paragraph you're talking about how you cannot use end-to-end tests to cover everything which is true of course and this paragraph offers testing the mock API as the solution. Testing the mock API is the solution to potentially testing against an API that is broken in itself and thus makes the test invalid. I think that part is missing – that the mock API might be wrong or broken.

_posts/2020-06-23-testing-mirage.md
```diff
@@ -0,0 +1,198 @@
...
+
+[Mirage.js]: https://miragejs.com/
+
+<!--break-->
+
+
+## Where do we put those tests?
+
+Before we start writing tests we need to figure out where to put all those new
+tests. In a regular Ember.js app there is a top-level `tests` folder, where
```
marcoow: I'd maybe say something like _"In the case of an Ember.js app…"_ or so to make it clear this all is not really specific to Ember.js since Mirage.js is now available standalone as well?

```suggestion
tests. In the case of an Ember.js app there is a top-level `tests` folder, where
```

_posts/2020-06-23-testing-mirage.md
```diff
@@ -0,0 +1,198 @@
...
+      assert.equal(response.status, 200);
+
+      let responsePayload = await response.json();
+      assert.deepEqual(responsePayload, {
+        user: {
+          id: 1,
+          email: 'johnny@dee.io',
+          first_name: 'John',
+          last_name: 'Doe',
+        },
```
marcoow: Is there a way to get this POJO from the `user` created in https://github.com/simplabs/simplabs.github.io/pull/1180/files#diff-d1337d072b6f4c579ba59e1c6783ec6fR63 ?

_posts/2020-06-23-testing-mirage.md
```diff
@@ -0,0 +1,198 @@
+---
+title: 'Testing the Mirage.js setup'
```
marcoow: This seems like it could be improved – _"the"_ makes it sound a bit strange tbh and I think it might be worth it to somehow include the fact that testing the Mirage.js setup can give you more confidence that your tests actually test the _"real"_ behavior.

Mikek2252 (commented):
Looks clear to me, but i agree it would be nice to get opinions on keeping your mirage in sync with the api. But i'm not sure how you would go about that other than running then same tests with passthroughs on the real api

_posts/2020-06-23-testing-mirage.md
mansona:
```suggestion
title: 'Testing your Mirage.js setup'
```
I don't know if this is the kind of tone you're intending to go with but it feels better than `the` in this context :thinking:

_posts/2020-06-23-testing-mirage.md
mansona:
```suggestion
One solution to some of the challenges of using a mock API is to test it and make sure it matches what you would expect from your real API. In this blog post we will show you how we
```

I agree with @marcoow on this one, but my suggestion above proposes a potential solution. I don't know if you would need a better transition section between the two paragraphs or if you can get away with more verbosity in the first line 👍

_posts/2020-06-23-testing-mirage.md
mansona: I think it's probably not a good idea to use the POJO in this case, you probably want to make sure that your assertion is very specific (and possibly it's better that it's verbose 🤔 ) . It might be worth mentioning that though 😂

mansona (approved):
Nice and to the point 👍  I think it's a great post because the value is in the fact that most people might not think to even test something like this and going into too much detail would take away from the impact of the post.

I added a few comments to the existing threads but other than that I think it already reads very well and is good to go 👍

_posts/2020-06-23-testing-mirage.md
sdebarros: I like @mansona's suggestion, it also gives a better idea of why we would test the Mirage setup (which may not make sense if you're not really customizing the handlers), without going into a lot of detail of other options. I like that the post is very straightforward and more focused on the code examples. Discussing the other options could be a different post 🤔

sdebarros (approved):
Great post! This is one of the things we do at Qonto that I really like and I think it's awesome to share it :)

_posts/2020-06-23-testing-mirage.md
Turbo87: we explicitly do not want to share the data, since one is the input data for the Mirage factory and the other is the serialized representation of the backend resource. those do look roughly similar in most cases, but we have encountered various cases at Qonto where there are significant differences, and sharing those POJOs would make those test more complicated than duplicating them.

marcoow (approved):

PR #1230, opened 2020-08-11
node v10.22 pinned with volta

izelnakri:
Same major version as we use in the CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,5 +131,8 @@
       }
     ]
   },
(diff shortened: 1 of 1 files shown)


marcoow (approved):

Turbo87:
might make sense to also use https://github.com/volta-cli/action?

izelnakri:
interesting, I've never used it before, will try it out here!

PR #1233, opened 2020-08-12
Pin volta nodejs on github actions

izelnakri:
This PR introduces volta to our CI pipeline, just like our development environment.

The official volta github action requires `yarn-version` to be declared for some of our CI actions. `node-version` is declared for the action to cache the node version before each step starts.


.github/workflows/ci.yml
```diff
@@ -24,9 +24,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v1
-      - uses: actions/setup-node@v1
+      - uses: volta-cli/action@v1
         with:
-          node-version: 10.x
+          node-version: 10.22.0
+          yarn-version: 1.22.4
```
Turbo87: why do you specify them here? shouldn't it be enough to specify them once in the `package.json` file?

izelnakri: certain CI tasks error out when yarn-version isn't defined. Yes I also thought it would be nice the have it defined once in the package.json, however this makes sure the booted container have the specific v10.22.0 node version installed, so it doesnt have to install it.

marcoow:
@izelnakri please add a description to the PR – see https://github.com/simplabs/playbook/tree/master/src/development-process#pull-requests.

This is what I was talking about regarding the small things we all should be doing in terms of execution that make a big difference in the all-hands call on Monday…

.github/workflows/ci.yml
marcoow: So this is an optimization really and if we didn't have this or updated the package.json but not this all that would happen is the build becoming slower since the Node/Yarn versions would have to be installed every time?

Turbo87: > however this makes sure the booted container have the specific v10.22.0 node version installed, so it doesnt have to install it

do you have a link to the docs? I'm not sure how that is supposed to work tbh.

https://github.com/volta-cli/action#usage seems to suggest that not specifying the values here will simply fall back to the values from the `package.json` file 🤔 

seems to me that the main use case for specifying these values in the CI config is if you use a matrix and want to test on multiple Node versions as shown in the last snippet in their documentation

izelnakri: It seemed to me its only installed when node-version is defined: https://github.com/volta-cli/action/blob/master/src/index.ts#L13
getInput returns an empty string when there isnt an input: https://github.com/actions/toolkit/blob/master/packages/core/src/core.ts

Since my knowledge of github actions are limited I will try it to be sure. Having all the node versions in the container would have increased the container size & transfer, but at the same time github actions are not exactly like containers :sweat_smile:

marcoow (approved):

izelnakri:
without version in yml 3m 33s:

without-version-in-yml

with version in yml 2m 57s:

with-version-in-yml

I'm not sure at this point if node gets cached or not, I cant see it in the logs, and installing fresh node.js version it generally takes 6 seconds. Thus maybe removing it from the `ci.yml` declaration is a better idea, I've made the ci work if I pin yarn as well to the package.json. So I'll keep it simple and remove the optimization(if it does optimize). @marcoow @Turbo87

package.json
```diff
@@ -133,6 +133,7 @@
   },
   "dependencies": {},
   "volta": {
-    "node": "10.22.0"
+    "node": "12.18.3",
```
Turbo87: this shouldn't be changed in the same PR 😉

izelnakri: ok removing that commit :smiley:

PR #1234, opened 2020-08-13
Case study qonto page design issues

izelnakri:
1- `<ShapeWork>` svg clip on the image on mobile doesnt apply to the container.

2- Couldn't figure out a way to properly style `.small` with css-blocks when its yielded to `<ShapeWork>` without introducing a custom css styling to the pages. Another option is:
I can refactor the <ShapeWork /> component to accept both quote text and quote person as attributes than we can have a specific styling on the component with css-blocks.

Preview: https://percy.io/simplabs-com/simplabs.com/builds/6494240 [view it on mobile]

### DIFF
--- a/src/ui/components/ShapeWork/stylesheet.css
+++ b/src/ui/components/ShapeWork/stylesheet.css
@@ -211,4 +211,20 @@
   }
 }
 
+@media (max-width: 888px) {
+  :scope {
+    background: var(--shape-color);
+    color: var(--color-text-inverted);
+  }
+
+  :scope .content {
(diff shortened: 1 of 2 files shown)


marcoow (commented):
This doesn't seem to solve it though since the rop right corner here should be white?

Bildschirmfoto 2020-08-13 um 19 27 22

marcoow:
Ah, I realize we're still using an SVG here instead of a `clip-path` that we use for the other shapes so that's why this is unnecessarily complex. Let's skip this for now and add as an improvement later.

PR #1237, opened 2020-08-15
copy change for qonto case study page

izelnakri:
There was a copy error in the qonto case study page, this PR fixes it.


marcoow (approved):
🤦 ;)

Issue #1242, opened 2020-08-18
Improve case study components in Work page


marhigh:
We included a third case study component in the Work page to highlight the most important examples of our work. 

This task includes: 

- [x] Adding a third case study in the middle of the client tiles
- [x] Update the placement of the components, so they go left, right, left (see the logo shape)
- [x] Update the content, so it displays Qonto, Trainline, and last Timify 
- [x] Update the logo shape in the component on mobile, to make the component more compact

Relevant designs are found here: https://share.goabstract.com/b74672c1-be3c-4395-9aaa-4922c6816f22

Issue #1241, opened 2020-08-18
Add Qonto case study to Homepage


marhigh:
We updated the design of the homepage to highlight Qonto as a case study. 

This task includes: 

- [x]  Update the promotional banner at the top of the homepage to link to Qonto's case study
- [x] Update the logos to add Timfy and re-order 
- [x] Update how the logos respond to smaller viewports 

Relevant designs are found here: https://share.goabstract.com/b74672c1-be3c-4395-9aaa-4922c6816f22

PR #1248, opened 2020-08-24
Link Qonto case teaser

marcoow:
This links the Qonto case teaser from the homepage and the work page. It also rearranges the work page slightly so that we now show three big banners for cases: Qonto, Trainline and Timify.

addresses #1241 
addresses #1242 

## Screenshots

localhost_4200_cases_qonto_
localhost_4200_cases_qonto_(Pixel 2 XL)
localhost_4200_cases_qonto_ (1)
localhost_4200_cases_qonto_(Pixel 2)


src/ui/components/ShapeFeatureQonto/component-test.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const { module, test } = QUnit;
+
+module('Component: ShapeFeatureQonto', function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function(assert) {
+    await render(hbs`<ShapeFeatureQonto />`);
+
+    assert.ok(this.containerElement.querySelector('div'));
```
locks, line 13: Should we keep generated tests?

marcoow: I think they serve as smoke tests and while they're not super valuable there's not really anything beyond what this test asserts that we would assert in a custom written test for a completely static component like this anyway. In real apps, I'd be opposed to keeping generated tests and ask for real ones to be added instead but for a static site these kinds of tests **are** the real tests already.

marhigh:
**Design Review: Approve**
From a design perspective, the tasks that were addressed look like expected 🙌

locks (approved):

marhigh (approved):

PR #1266, opened 2020-09-02
improve tagline on homepage

marcoow:
This improves the tagline to be more concise and more clearly express what we deliver, making sure the aspect of mentoring and leveling up companies and teams is includes as well:

localhost_4200_
localhost_4200_ (1)

### DIFF
--- a/src/ui/components/PageHomepage/template.hbs
+++ b/src/ui/components/PageHomepage/template.hbs
@@ -5,13 +5,13 @@
   >
     <HeaderContent>
       <h1 typography:class="display">
-        Solid Solutions for
+        Guiding projects and teams
         <span typography:class="nowrap">
-          Ambitious Projects
+          to sustainable success
         </span>
       </h1>
       <p typography:class="lead">
-        We deliver ambitious digital products on the web and mobile for our clients to rely on. Our expert team manages projects from idea to release, covering strategy, design and engineering.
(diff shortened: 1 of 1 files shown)


marhigh:
Why is Digital Product Consultancy capitalized? 
And, do we need to say mobile + web? Could we consider saying "mobile applications" maybe? 

Aside from that, I like it. I think it might make sense to highlight the fact that we're a fully remote team, but this is probably not the right place.

marcoow:
> Why is Digital Product Consultancy capitalized?

I don't know – always happy to be corrected by the native speakers ;)

> And, do we need to say mobile + web? Could we consider saying "mobile applications" maybe?

I think web means desktop web apps to most people while _"mobile applications"_ means native mobile apps to most people. I think we could say

_"simplabs is a Digital Product Consultancy that designs and develops web applications."_

but then we're not explicitly mentioning mobile which might turn some people away that are looking for a mobile **web** app.

marhigh:
Let's leave digital product consultancy in all lowercase since it's not a proper noun and it's in the middle of a sentence 👍

My bad, I was thinking "modern applications" but wrote "mobile applications" - I totally understand that the second term would be misleading. I would still favor "web applications" or "modern applications" over the existing phrase because it is more direct, and it sounds more modern. We could highlight the fact that we make mobile web apps in another part of the page.

marcoow:
Both changed – I think we should be saying _"web applications"_ since _"modern"_ doesn't really mean anything.

marhigh (approved):

PR #1358, opened 2020-11-12
remove whitespace around quotes in shape

### LINKED ISSUE #996: Remove extra space from quote
Due to the prettier formatting of templates, newlines are introduced that are then rendered as an empty space between the quotation mark and the content of the quote:

image

## Tasks
- [ ] Remove extra whitespace from quote content.

marcoow:
This removes the whitespace around quotes in shapes that is a result of the template formatting we apply (and enforce via Prettier).

| Before | After |
|-------|---------|
| before | ater |

closes #996

### DIFF
--- a/src/ui/components/ShapeQuoteContent/component-test.ts
+++ b/src/ui/components/ShapeQuoteContent/component-test.ts
@@ -1,0 +1,15 @@
+import hbs from '@glimmer/inline-precompile';
+import { render } from '@glimmer/test-helpers';
+import { setupRenderingTest } from '../../../utils/test-helpers/setup-rendering-test';
+
+const { module, test } = QUnit;
+
+module('Component: ShapeQuoteContent', function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function(assert) {
+    await render(hbs`<ShapeQuoteContent @source="Person, Title, Company">simplabs is so awesome!</ShapeQuoteContent>`);
+
+    assert.ok(this.containerElement.querySelector('q'));
+  });
+});
--- a/src/ui/components/ShapeQuoteContent/component-test.ts
+++ b/src/ui/components/ShapeQuoteContent/component-test.ts
@@ -8,7 +8,7 @@ module('Component: ShapeQuoteContent', function(hooks) {
   setupRenderingTest(hooks);
 
   test('it renders', async function(assert) {
-    await render(hbs`<ShapeQuoteContent @source="Person, Title, Company">simplabs is so awesome!</ShapeQuoteContent>`);
+    await render(hbs`<ShapeQuoteContent @source="Person, Title, Company" @text="simplabs is so awesome!" />`);
 
     assert.ok(this.containerElement.querySelector('q'));
   });
--- a/src/ui/components/ShapeQuoteCardstack/stylesheet.css
+++ b/src/ui/components/ShapeQuoteCardstack/stylesheet.css
@@ -1,7 +1,3 @@
-@block typography from "../../styles/blocks/typography.block.css";
-
 :scope {
   block-name: ShapeQuoteCardstack;
 }
-
-@export typography;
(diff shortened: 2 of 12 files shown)


Turbo87:
why do we not use `~` in the template to achieve that? seems a bit strange to me to use CSS to workaround a limitation in the code formatter 🤔

marcoow:
@Turbo87 I agree this workaround isn't great but the template is this: https://github.com/simplabs/simplabs.github.io/blob/master/src/ui/components/ShapeQuoteQonto/template.hbs – not sure how we'd use `~` there?

Turbo87:
hmm, where are the quote characters defined? I don't see them in `styles/blocks/typography.block.css` 🤔 

and wouldn't it be better to extract a generic quote component instead of building one for each individual quote?

marcoow:
The quotes themselves are standard browser styles. We could extract a quote component as well, yes – the downside would be we'd have the quote content in attributes which would make it harder to add any HTML to that (although I'm not sure we need that at all).

Turbo87:
> the downside would be we'd have the quote content in attributes which would make it harder to add any HTML to that 

you can always put `yield` inside and use it in block form if you really need HTML

> The quotes themselves are standard browser styles.

oh, because it's a `q` element, I see.

> not sure how we'd use ~ there?

the way it's currently built this would indeed be complicated. if we extracted a component then we should be able to use something like `{{~@text~}}` which should remove the additional whitespace.

https://github.com/simplabs/ember-hbs-minifier might potentially also solve this in a somewhat unintentional way 😅

marcoow:
> you can always put yield inside and use it in block form if you really need HTML

🤦 sure

This doesn't solve the problem though:

```hbs
<ShapeQuoteContent @source="Marc-Antoine Lacroix, Qonto CTO">
      simplabs are well known as the Ember.js experts and they absolutely live up to the expectations. They had an immediate as well as significant positive impact on both our velocity and quality of output.
    </ShapeQuoteContent>
```

```hbs
<p>
  <q typography:class="quote" lang="en">
    {{~yield~}}
  </q>
</p>
```

We still end up with:

```html
<q lang="en" class="typography_020c8__quote">
        simplabs' experienced engineers delivered a solid and well architected foundation for our web app. They also helped us establish best practices and a lean process internally. Working with them was a pleasure.
      </q>
```

marcoow:
I extracted a `ShapeQuoteContent` component but I don't think we can use `yield` in that unless we can capture the `yield`'s output and trim leading/trailing whitespace with a helper.

Turbo87:
yeah, with `yield` this might not work as well. but do we actually have a need for using `yield` or is this just nice to have?

marcoow:
No, there's no actual need for `yield` here. I just think passing a super long string like this in an argument isn't such a great API – block invocation would just look nicer in the source…

nickschot:
There is also the remove whitespace with a handlebars comment "hack": `{{~! ~}}`

Turbo87:
😂 😂 😂 😂  I've never seen that before, but that's indeed an awesome hack 😂 😂 😂 😂

marcoow:
@Turbo87 @nickschot can someone 👍/👎 please?

src/ui/components/ShapeQuoteContent/component-test.ts
```diff
@@ -0,0 +1,15 @@
...
+import { render } from '@glimmer/test-helpers';
+import { setupRenderingTest } from '../../../utils/test-helpers/setup-rendering-test';
+
+const { module, test } = QUnit;
+
+module('Component: ShapeQuoteContent', function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function(assert) {
+    await render(hbs`<ShapeQuoteContent @source="Person, Title, Company">simplabs is so awesome!</ShapeQuoteContent>`);
```
Turbo87:
```suggestion
    await render(hbs`<ShapeQuoteContent @source="Person, Title, Company" @text="simplabs is so awesome!" />`);
```

Turbo87 (approved):

PR #1364, opened 2020-11-17
Add plausible analytics

marcoow:
Netlify analytics is pretty limited and does not actually work for SPAs. We only ever get to see the first page that a visitor requests but since Netlify will not actually see any more requests as the user navigates the page, we cannot know whether users visit any of the other pages as well.

[Plausible analytics](https://plausible.io/) seems to be a great alternative – it's cheap, privacy-friendly (hosted in EU, doesn't track any private data, doesn't set cookies – not even Safari reports it as a tracker) and simple.

Bildschirmfoto 2020-11-17 um 09 32 20

### DIFF
--- a/lib/analytics/index.js
+++ b/lib/analytics/index.js
@@ -8,7 +8,7 @@ module.exports = {
   },
 
   contentFor(type) {
-    if (type === 'head-footer') {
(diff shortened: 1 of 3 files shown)


locks (approved):
I was not familiar with Plausible but reading their marketing they seem to be a very nice solution for basic analytics.

PR #1365, opened 2020-11-18
Add OnProduct talk to calendar

marhigh:
Add talk to calendar and also modify the one by Women Techmakers to make distinction between the two obvious


_calendar/2020-12-10-onproduct.md
```diff
@@ -0,0 +1,20 @@
...
+location: Remote
+url: https://www.meetup.com/OnProduct/events/274647325
+kind: meetup
+---
+
+OnProduct is a meetup for product people in Berlin. Their goal is to create a
+living community of designers, product managers, developers, QA engineers that
+exchange ideas, best practices and know how.
+
+In case you miss the previous presentation, Mar High will share once again give
```
marcoow, line 14: Be aware the other one will be gone from the calendar next week (when it's over) so it will not be clear what _"previous"_ might refer to then?

marhigh: I am aware and believe it makes sense as it is written. Do you prefer it phrased in a different way?

marcoow: all good 👍

marcoow (approved):

PR #1370, opened 2020-11-19
Fix column sizing

marcoow:
This fixes the sizing of columns and as an effects prevents the site from growing too much horziontally.

| Before | After |
|--------|------|
| Bildschirmfoto 2020-11-19 um 12 51 08 | Bildschirmfoto 2020-11-19 um 12 51 20 |

closes #1319

PR #1371, opened 2020-11-19
Unify font sizes

### LINKED ISSUE #728: Font-size is too large for wide screens
All headings, sub-headings and even the body of blog posts become unpleasant to read on screens that are wider than 1440px.

image

marcoow:
This bases all font sizes on the same set of variables so they are unified across the entire site.

It also removes the extra scaling factor for huge viewports that scales fonts up from their usual size.


marcoow:
@marhigh this should now use the font sizes that you defined in Abstract (only missing h4/h5 sizes). Regular text on mobile now looks tiny though – not sure that's what we want? Or did I misunderstand something?

marhigh:
@marcoow I only meant to address max font sizes on wide screens  in https://github.com/simplabs/simplabs.github.io/issues/728#issuecomment-730509804 

Mobile should not be updated - even if it was, the font sizes in my table were extracted from the current font sizes right at the point the viewport breaks into mobile. I will take a look but it is likely a misunderstanding if all font sizes in all viewports were refactored (and mobile is even smaller than it is now) - I only wanted to address the issue that was pointed out in the other ticket.

marhigh:
@marcoow I am not sure where you got the mobile body size from - in Abstract and in Notion it's set to 15px (p1) - shall we take a look at this task together? 

***

I noticed the following buggy behavior (viewport width tested: 2506px): 

Text overlaps shape in https://deploy-preview-1371--objective-northcutt-37494c.netlify.app/why-simplabs/
Screen Shot 2020-11-20 at 11 11 29
***
Playbook page also has unwanted overlap: https://deploy-preview-1371--objective-northcutt-37494c.netlify.app/playbook/
Screen Shot 2020-11-20 at 11 16 56
***
Qonto's shape has text outside of edge
https://deploy-preview-1371--objective-northcutt-37494c.netlify.app/cases/qonto/
Screen Shot 2020-11-20 at 11 19 13
***


It would be ideal that we review + eventually release this together with the max-width ticket.

marcoow:
So 15px should equal [1.875rem](https://github.com/simplabs/simplabs.github.io/pull/1371/files#diff-ec31d3e423531db4ae770faca348c76ab7ff1f76657abc0b5cf81241006988d7R245) if I understood @pichfl correctly. I can revert the changes for mobile but if we have the font sizes in Abstract already wouldn't it be better to make sure the code and Abstract match? Also we're missing sizes for h4/h5.

marhigh:
They should already match! I took them from an inventory I made of the website. I don't understand what exactly is not matching. 

Will add H4+H5 asap

marcoow:
I fixed the overflow problem on the playbook page. The problem with the text overflowing the shape on the Qonto page exists in production as well in extreme viewports and will be fixed by #1370 so I wouldn't add a fix to this PR…

marcoow:
Ah sorry – found the problem with mobile font sizes. The scale isn't actually 8px on mobile but 6px – will fix.

marcoow:
@marhigh so for font sizes, we actually have 3 breakpoints:

* width <= 887px
* 887px <= width <= 960px
* 960px <= width <= 1440px

Could you update the typography overview in Abstract to include all font sizes we want to have for each of those breakpoints?

marhigh:
@marcoow I don't think it makes sense to do so at this point since the website will be redesigned with the rebrand and typography will be getting an overhaul then. I just did the inventory to have a stable set to work with since the design files were outdated. I would suggest waiting and actually doing this properly, including bringing in 3 breakpoints to the site as a whole when we refresh the design.

marcoow:
@marhigh I changed this to simply unify font sizes across the entire site – some of the font sizes will change as a result of that. It's ready for review.

marhigh:
Looks good at first glance, let's definitely review together w/ the max-content width before releasing. 

You might want to reduce the size of some of the mobile headlines, they look particularly large and out of place.

E.g. home @ 375px 
Screen Shot 2020-11-20 at 15 12 33

marcoow:
Let's treat this separately from the `max-width` change please. We shouldn't be merging multiple changes together if not necessary and I don't think it is necessary to block this before releasing the other change? What size would you want me to change the headlines to?

marhigh:
Why not keep the existing size for mobile headlines? They are much bigger in this ticket than they are on the live site. E.g. Display H1 is 36px in the live site vs 60px here.

marcoow:
ouch, right – fixed

marhigh:
Font sizes look as expected 👍

The first image on the playbook page is quite small, I suppose that is because the component is now smaller due to smaller font sizes. Have you seen it? It looks a bit buggy especially compared to the other images on that page. Shall we create another issue to fix the size, or should it be done here since this issue caused the change?
Screen Shot 2020-11-20 at 16 14 53

marcoow:
@marhigh the issue on the playbook page was caused by a fix I had added here earlier when the font sizes were smaller which was causing [another bug](https://github.com/simplabs/simplabs.github.io/pull/1371#issuecomment-731084076). I reverted that fix.

marhigh:
Okay, one more request. Please update the text on the client grid (https://deploy-preview-1371--objective-northcutt-37494c.netlify.app/work/) to a default p1 size.

After that, this is ready from my end.

Screen Shot 2020-11-20 at 16 54 19

marhigh (approved):

PR #1404, opened 2020-12-15
Add Karol's interview

marhigh:
Add Karol's interview to our blog, including a YouTube embed & og image


marcoow (approved):

marhigh:
@marcoow what do you do when you optimize the image? I already ran it through ImageOptim. I would be happy to do all the necessary steps for uploading it fully optimized next time 🙌

marcoow:
@marhigh I use ImageOptim as well – think I just run it with max settings

PR #1425, opened 2021-01-13
render clients in ClientsGrid in specified order

marcoow:
The `<ClientsGrid>` component allows rendering only a particular set of clients but will not respect the order that these clients are specified in (instead it renders the set in the order they appear in an internal data structure of that component). This changes that so that e.g.

```hbs
<ClientsGrid @only="erdil,loconet" />
```

will in fact first render ERDiL, then LoCoNET.


marhigh:
@marcoow I am not the right person to review this :/

Mikek2252 (approved):

PR #1432, opened 2021-01-15
Fix capitalization of CLARK

marcoow:
Clark is now called CLARK – see https://clark.de


Turbo87 (approved):
FWIW I'm not sure we really need a review step for trivial changes like this 😉

marcoow:
I agree but it can be hard to draw the line sometime – I'd rather stick to a general rule until it actually hurts us which I don't have the feeling is the case for now.

PR #1528, opened 2021-05-11
use youtube-nocookie for all video embeds

mansona:
This is based on the documentation here: https://support.google.com/youtube/answer/171780?visit_id=637563189687636677-2621067568&rd=1#zippy=%2Cturn-on-privacy-enhanced-mode for "Privacy enhanced mode" 

Essentially if we just do a normal youtube embed then we need a cookie banner again 🙈  so this fixes that 👍 

We should probably have a lint rule to make sure that this stays the case?


Turbo87 (approved):

marcoow (approved):
awesome – didn't even know about this 🙌

PR #1537, opened 2021-07-05
Track contact form submissions in plausible

marcoow:
This tracks submissions of the contact form in plausible so we can better understand what results in conversions and what doesn't.

### DIFF
--- a/lib/analytics/index.js
+++ b/lib/analytics/index.js
@@ -11,6 +11,7 @@ module.exports = {
     if (process.env.CONTEXT === 'production' && type === 'head-footer') {
       return `
         <script async defer data-domain="simplabs.com" src="https://plausible.io/js/plausible.js"></script>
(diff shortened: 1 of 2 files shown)


Turbo87:
looks like there are checks failures on the "Files changed" tab, but for some reason they don't fail CI 🤔

marcoow:
I've always been confused by those… I don't think those are actual issues though, might be warning that are put out during build or so

Turbo87:
> I don't think those are actual issues though

I'm not so sure... I would actually expect `lint:js` or `lint:ts` to trigger because of an undefined variable. ESLint needs to be explicitly told about global variables, unless you call it via `window.plausible(...)`.

Additionally, that snippet that you added which defines `window.plausible` is apparently only inserted for production builds, if I understand the code correctly. That means for non-production builds the `plausible()` call will fail because `window.plausible` is `undefined`.

marcoow:
you're right – fixed

emmasofiah2o (approved):

PR #1561, opened 2021-08-19
bump engines.node to 12 as well

locks:
This should make Netlify use Node 12, and thus go back to a green build.


mansona (approved):
looks good

PR #1775, opened 2022-04-11
Added SAEC Days to calendar

emmasofiah2o:
- Added file


emmasofiah2o:
I think I'll wait until @marcoow is back from vacation for this one. I believe we might want to have it in the appearances section of the website as well, but I wasn't sure if that works without a video/before the talk has been given. Also, I would write this in German, but I don't know if that makes sense since the whole website is in English. Let me know :)

marcoow:
So this should be on the calendar only for now. Once there's a video on YouTube or so we can put it in the talks list as well. I'd say let's write the summary in english since our whole site is in english.

marcoow (approved):
👍

PR #1831, opened 2022-07-26
Added RustConf to calendar

emmasofiah2o:
- created entry
- added description

To-do:
- find and add logo


marcoow (changes requested):
Can we use the SVG? That should have much smaller file size.

emmasofiah2o:
> Can we use the SVG? That should have much smaller file size.

I did use the SVG, changed colors, added a background, and exported it as a png. If you want me to export the image as an svg I can do that.

marcoow (approved):

PR #1885, opened 2022-09-20
Fix sorting Elixir posts

### LINKED ISSUE #1882: Reverse sorting of blog posts on Elixir page
The posts appear oldest first but should start with the latest of course.

marcoow:
This fixes sorting of the Elixir posts by explicitly defining the respective collection. I couldn't reproduce where the previously used collection even comes from – might be some kind of automatic thing that 11ty does based on Tags. This gives us a bit more freedom anyway.

closes #1882

### DIFF
--- a/collections/elixirPosts.js
+++ b/collections/elixirPosts.js
@@ -1,0 +1,8 @@
+const livePosts = require("./posts");
+
+module.exports = collection => {
+  const allPosts = livePosts(collection);
+  return allPosts.filter(
+    post => Array.isArray(post.data.tags) && post.data.tags.some(t => t.match(/^(elixir|phoenix)/))
+  );
+};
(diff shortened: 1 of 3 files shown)


collections/elixirPosts.js
```diff
@@ -0,0 +1,11 @@
+const livePosts = require("./posts");
+
+module.exports = collection => {
+  const allPosts = livePosts(collection);
+  return [
+    ...allPosts.filter(
+      post =>
+        Array.isArray(post.data.tags) && post.data.tags.some(t => t.match(/^(elixir|phoenix)/))
+    ),
+  ];
```
BobrImperator: The spread operator seems to be redundant.

```suggestion
  return allPosts.filter(
      post =>
        Array.isArray(post.data.tags) && post.data.tags.some(t => t.match(/^(elixir|phoenix)/))
    ),
```

It would probably make sense to make the regex case-insensitive.

```suggestion
  return [
    ...allPosts.filter(
      post =>
        Array.isArray(post.data.tags) && post.data.tags.some(t => t.match(/^(elixir|phoenix)/i))
    ),
  ];
```

Lastly, I'm not exactly sure why the "starts with" pattern is needed here (`^`) but it seems fine.

marcoow: right – fixed 👍

BobrImperator (approved):

PR #1886, opened 2022-09-20
Small suggestions

locks:


locks:
@pichfl I couldn't figure out how to suggest these using the interface 😂.

PR #1898, opened 2022-10-18
chore: use volta-cli/action@v4

BobrImperator:


mansona (approved):
Looks good to me! CI is passing which is pretty clear that it's still working 😂

PR #1905, opened 2022-10-24
Add error tracking via Sentry

### LINKED ISSUE #1902: Add error tracking
We should have at least some error tracking in place. Of course this is a static page which little potential for things to go wrong and 99% of the JS is optional anyway really but it would be good to have a feeling for whether things generally work or not. One specific case where we'd absolutely need to know if errors happen is the contact form – if that doesn't submit successfully, we should be aware.

## TODOs

- [x] add general Sentry catchall code for JS
- [x] report submission errors of the contact form to Sentry explicitly (including the original message so that isn't lost)
- [x] add a clause about usage of Sentry to the [privacy policy](https://mainmatter.com/privacy/)
- [x] enable source maps for JS (as we'll need them for Sentry)

marcoow:
This adds Sentry:

* enables sourcemaps for more meaningful logs on Sentry
* tracks any uncaught errors with Sentry
* explicitly logs failed contact form deliveries to Sentry
* adds Sentry to our privacy policy

closes #1902

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "last 1 Edge versions"
   ],
   "dependencies": {
+    "@sentry/browser": "^7.16.0",
     "lax.js": "^2.0.3"
   },
   "engines": {
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -1,9 +1,16 @@
+import * as Sentry from "@sentry/browser";
 import "wicg-inert";
 import { Animations } from "./animations";
 import { Nav } from "./nav";
 import { ContactForm } from "./contact-form";
 import { LogoList } from "./logo-list";
 import bindSelectDropdowns from "./select";
+
+if (window.location.host === "mainmatter.com") {
+  Sentry.init({
+    dsn: "https://43f7627909d94dc4a769340ad730f1a2@o68744.ingest.sentry.io/4504039028817920"
+  });
+}
 
 const navElement = document.getElementById("nav");
 new Nav(navElement);
--- a/src/privacy.njk
+++ b/src/privacy.njk
@@ -488,7 +488,6 @@ noindex: true
       burden of proof in a procedure under the General Equal Treatment Act (AGG).
     </p>
     <h2>
-    <h2>
       9. Data protection provisions about the application and use of Plausible Analytics
     </h2>
     <p>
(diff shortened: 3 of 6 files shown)


src/assets/js/app.js
```diff
@@ -1,10 +1,24 @@
...
+if (host === "mainmatter.com") {
+  environment = "production";
+} else if (host.endsWith("netlify.app")) {
+  environment = "staging";
+}
+
+Sentry.init({
+  dsn: "https://43f7627909d94dc4a769340ad730f1a2@o68744.ingest.sentry.io/4504039028817920",
+  environment
+});
```
Turbo87: sure that we want to enable it in development mode too? Sentry by default shows all environments and seeing devtime errors in there could be a bit confusing.

src/privacy.njk
```diff
@@ -488,7 +488,22 @@ noindex: true
...
+      ensure a smooth experience for all users.
+    </p>
+    <p>
+      The operator of the Sentry component is Functional Software Inc., 45 Fremont Street 8th Floor, San Francisco, CA 94105, United States.
+    </p>
+    <p>
+      For Sentry's policies and practices around analysis and storage of user data, refer to <a href="https://sentry.io/legal">https://sentry.io/legal</a>.
+    </p>
+    <h2>
+      10. Data protection provisions about the application and use of Google Analytics (with anonymization function)
```
Turbo87: seems unrelated?

package.json
```diff
@@ -63,6 +63,8 @@
     "last 1 Edge versions"
   ],
   "dependencies": {
+    "@sentry/browser": "^7.16.0",
+    "@sentry/tracing": "^7.16.0",
```
Turbo87: this does not seem to be used anywhere?

src/privacy.njk
Turbo87: nevermind, I had mistaken this for the changes in the other two PRs 🤦

package.json
marcoow: ah. right – should be removed 👍

src/assets/js/app.js
marcoow: fair point – let's enable this on mainmatter.com only

BobrImperator (approved):

PR #1913, opened 2022-10-26
Fix dimensions of monochrome logos

marcoow:
Some of the monochrome logos had such small viewports that their maximum display size was limited by that and not by the size we assign via CSS. This makes them all have a width of 1000 which doesn't have an effect on file size as they are vectors but allows them to grow larger if there's space in the viewport.

## TODO

- [ ] add max size via CSS for logo lists


marcoow:
nvm

PR #1921, opened 2022-10-27
Stabilize Percy

marcoow:
This stabilizes the Percy integration which was pretty flaky so far and thus caused quiet a bit of noise.

* disable all animations for Percy
* remove all `loading="lazy"` attributes from images which might cause images not to have loaded when Percy takes the snapshot (we had this before but it never worked – see e.g. https://github.com/mainmatter/mainmatter.com/actions/runs/3337295611/jobs/5523460437#step:5:223)

closes #1899

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,10 +35,7 @@ jobs:
         run: yarn install
 
       - name: build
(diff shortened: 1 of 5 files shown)


marcoow:
It does look indeed much better 🎉

BobrImperator (approved):

Issue #1914, opened 2022-10-27
Add new case studies


marcoow:
We have case studies for 3 projects that we haven't put on the website yet: Experteer, MVB, and AIS. Those need to be added to the site.

## TODOs

- [x] add case study page for Experteer
- [x] add case study page for MVB
- [x] add case study page for AIS
- [x] update featured case studies on `/work` page so that the order of the featured case studies is Qonto, MVB, Trainline, Experteer, AIS, Timify, dd-wrt, Expedition – the rows of small project teasers in between remain the same
- [x] replace the small features of the Timify and dd-wrt and case studies on the landing page with Experteer and MVB so that the order of the small ones below the large Qonto one is Experteer, Trainline, MVB
- [x] add the new case studies to the list under _"Work"_ in the footer (order being the same as on the `/work` page)

## Resources

* [Designs](https://www.sketch.com/s/7baa7317-7ec9-48b5-9584-4edb6163c143/p/09ABE235-F655-4723-9ADF-FCD5D2CF0F9F/canvas)
* Case study text, original quotes and screenshots:
  * [AIS](https://www.notion.so/mainmatter/AIS-505b59ed44404128bf87ffafe189590b)
  * [MVB](https://www.notion.so/mainmatter/MVB-b9b2a96236a24a6eaf2b4c4dbfc2a64b)
  * [Experteer](https://www.notion.so/mainmatter/Experteer-CV-to-Profile-8d32a27c5c844280a9a4af8434d1ebab)

## To Prepare

- [x] The AIS case study design is missing the quote

marcoow:
done

PR #1940, opened 2022-11-03
add new case studies to work page

marcoow:
This adds the new Experteer, MVB, and AIS case studies to the `/work` page:

localhost_8080_work_

see #1914

### DIFF
--- a/src/work.njk
+++ b/src/work.njk
@@ -23,14 +23,20 @@ eleventyNavigation:
   <ul class="work__list">
     {{ featuredCaseStudy('qonto', 'purple', 'work__featured-case-study') }}
     {{ clientCards(['auditboard', 'clark', 'sage']) }}
-    {{ featuredCaseStudy('trainline', 'black', 'work__featured-case-study') }}
-    {{ clientCards(['mvb', 'camilyo', 'hop-skip-drive', 'ais', 'phorest', 'experteer']) }}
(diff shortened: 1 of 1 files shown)


davidhajba:
My suggestions:
- the Experteer background picture should be positioned to see the heads 
- The MVB background picture became very pale, the attached pic seems better


MVB - background

marcoow:
This isn't built in a way that would allow that level of control I'm afraid – we'd have to refactor things quite a bit and I'm not sure it's worth the effort?

davidhajba:
Could you please explain how exactly the pictures appear on the website and what we can control?

Does it add a purple overlay / gray transparency on images?
If yes, in my opinion, it would be better to not do it this way (or have the option to not do it this way) and upload just the final image with the final colors.

So we cannot control the position of the pictures in the picture boxes? Does it center align automatically?
If yes, we can crop the picture before uploading to be able to see the faces.

marcoow:
Yes, the images are centered automatically and the overlay is added via CSS. We could of course change that but the question is whether it's worth blocking the new case studies from being merged and released for that. It will likely take quite some time until anyone finds the time to make the necessary changes. I'd recommend we release this (unless it's a critical design problem) and change the things you mention in a second step afterwards?

davidhajba:
Yes, it is not critical.
Change these in second step - Sounds good to me.

BobrImperator (approved):

PR #1955, opened 2022-11-25
Add diff-js styles

beerinho:
This PR adds the styles to `diff-js` codeblocks so that they can now be used to emphasise code that has been changed in blog posts

Example:
image

I just took the colours directly from the 11ty example, so I'm happy to adapt them if needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "@11ty/eleventy-img": "^2.0.1",
     "@11ty/eleventy-navigation": "^0.3.2",
     "@11ty/eleventy-plugin-rss": "1.2.0",
-    "@11ty/eleventy-plugin-syntaxhighlight": "4.1.0",
+    "@11ty/eleventy-plugin-syntaxhighlight": "4.2.0",
     "@lhci/cli": "^0.9.0",
     "@percy/cli": "^1.10.2",
     "@quasibit/eleventy-plugin-sitemap": "^2.1.5",
--- a/src/assets/css/base/_code.scss
+++ b/src/assets/css/base/_code.scss
@@ -189,3 +189,30 @@ code {
     font-weight: 700;
   }
 }
+
+.token.deleted {
+  background-color: hsl(350deg 100% 88% / 47%);
+}
+
+.token.inserted {
+  background-color: hsl(120deg 73% 75% / 35%);
+}
+
+.token.deleted-arrow,
+.token.deleted-arrow>.deleted {
+  background-color: initial;
+}
+
+/* Make the + and - characters unselectable for copy/paste */
+.token.prefix.unchanged,
+.token.prefix.inserted,
+.token.prefix.deleted {
+  -webkit-user-select: none;
+  user-select: none;
+}
+
+/* Optional: full-width background color */
+.token.inserted:not(.prefix),
+.token.deleted:not(.prefix) {
+  display: block;
+}
--- a/src/posts/2018-01-24-ember-freestyle.md
+++ b/src/posts/2018-01-24-ember-freestyle.md
@@ -67,7 +67,7 @@ This is obviously not what we want. To fix this we will need to adjust the
 `app/templates/application.hbs` file and wrap the existing contents in a
 condition:
 
-```diff
+```diff-hbs
 {% raw %}
 +{{#if onFreestyleRoute}}
 +  {{outlet}}
--- a/yarn.lock
+++ b/yarn.lock
@@ -44,13 +44,13 @@
     posthtml "^0.16.6"
     posthtml-urls "1.0.0"
 
-"@11ty/eleventy-plugin-syntaxhighlight@4.1.0":
-  version "4.1.0"
-  resolved "https://registry.yarnpkg.com/@11ty/eleventy-plugin-syntaxhighlight/-/eleventy-plugin-syntaxhighlight-4.1.0.tgz#742fb1910c5f04d76421c6c95c38804b54155fc1"
-  integrity sha512-bLpV8DKFZRgh0kToh8JPCjABfalL5ydyP6rxj/aUgrlR2v9TheLGRNqoKMhfgwUETOas2nMo/rd7sCE4kSvBNQ==
+"@11ty/eleventy-plugin-syntaxhighlight@4.2.0":
+  version "4.2.0"
+  resolved "https://registry.yarnpkg.com/@11ty/eleventy-plugin-syntaxhighlight/-/eleventy-plugin-syntaxhighlight-4.2.0.tgz#f0c2d059ce4daabce06af4ae4c48f0084c7c9f20"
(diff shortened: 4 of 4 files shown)


marcoow (commented):
Hm, the screenshot seems a bit strange? I think it's `diff-js`, not just `diff` as it should have JS syntax highlighting **and** Red/Green for the Diff?

beerinho:
Yeah I was expecting the same so I played around a bit to see if I could get it working, but for some reason the tokens aren't added to the individual words like they are when just using `js`, it instead shows that a whole line has changed and then just spits it out as plain text
image

image
 
 Which is also strange, because it's different to the example in the 11ty [docs](https://www.11ty.dev/docs/plugins/syntaxhighlight/#show-changes-using-diff-syntax)

The 11ty [docs](https://www.11ty.dev/docs/plugins/syntaxhighlight/#show-changes-using-diff-syntax) suggested looking at the [Prism.js highlighting syntax](https://prismjs.com/plugins/diff-highlight/) but that just resulted in this error
```
"language-diff-javascript" is not a valid Prism.js language for eleventy-plugin-syntaxhighlight
```

I tried researching the issue but nothing useful came up

While playing around, I also noticed that the 11ty example isn't great either as the highlighting is a block and doesn't wrap the entire content, which is the same issue we have trying to implement it in the same way
Screen Recording 2022-12-02 at 10 25 46

marcoow:
Do we need to configure additional syntaxes or so maybe?

beerinho:
Possibly, I couldn't see anything in the docs about configuring additional syntaxes, the docs mainly just say "here are the ones that work" and "diff-js" is supposedly one of them (which does work, but not the JS syntax)

beerinho:
Hey @marcoow, how did you want to progress with this?

marcoow:
We should fix this – I don't think it's just completely broken tbh…

beerinho:
Hey @marcoow, I got it working nicely but using the old syntax ```[language]/[added_lines]/[deleted_lines]``` the benefit of that means that it _should_ work for all languages and we don't technically need the `-` or `+` anymore, it also wraps the entire content instead of having the weird part at the end that isn't highlighted if you scroll.
The drawback is the way that you have to write it: you have to count the lines that have changed and write them out.
Below is the `diff-js` format, then the new format, then the original code without the diff.

With the new but currently not really working `diff-js` syntax
image
Screenshot 2022-12-09 at 10 51 14

With the old `js/0/2` syntax
image
Screenshot 2022-12-09 at 10 51 20

Normal `js` tag
image
Screenshot 2022-12-09 at 10 51 27



If you're happy with this I'll add it to the PR, and if so, should we not have the `-` and `+` when showing a diff? i.e.
image
image

marcoow:
I don't fully understand – can you share an example for how you'd use this in markdown?

beerinho:
I have updated the previous comment with the markdown to hopefully better explain it

marcoow:
I see – don't think that's a good approach though tbh. Is that the suggested way to use this?

beerinho:
Yeah I'm not a fan of how you need to write it, the 11ty [docs](https://www.11ty.dev/docs/plugins/syntaxhighlight/#usage) say that the `diff-js` should work but it just doesn't for some reason. The outputted HTML is missing all of the spans that do the token highlighting
image

Compared to the old `js/2/0` syntax that still has all of the token highlighting
Screenshot 2022-12-09 at 11 15 02

The old version also adds `ins` and `del` elements which mean we could easily style the diff for all languages in one go

marcoow:
> The old version also

what's the old version?

beerinho:
The old version is the `js/2/0` syntax. I call it the old version because in the 11ty docs it is the older version of the highlighting

marcoow:
I'd debug into this to see what's going on. I don't think what's being teased on the 11ty site is just totally broken. Most likely we have an old version of something, are missing a configuration setting or something?

beerinho:
I've been banging my head against this for a couple of hours now and I'm really not sure where to go next, I've tried following the prism docs but they are very limited, I can't find anything online about any configuration that needs to be added apart from adding the style files to the project and setting up the `Prism.plugins.autoloader.languages_path` as mentioned [here](https://prismjs.com/plugins/autoloader/) (it seems like a requirement [here](https://prismjs.com/plugins/diff-highlight/)) but I still get the same error:
`"language-diff-javascript" is not a valid Prism.js language for eleventy-plugin-syntaxhighlight`
If you have any other ideas I'd love your input. Also, we are on a version of `eleventy-plugin-syntaxhighlight` that supports this so that shouldn't be the issue

beerinho:
There is an open issue for this on StackOverflow that has no responses - https://stackoverflow.com/questions/72097395/how-do-i-do-the-prismjs-diff-highlight-with-language-in-my-markdown-file/74821991#74821991

beerinho:
I've just asked on the 11ty discord - https://discord.com/channels/741017160297611315/1053236500613046292/1053236500613046292

marcoow:
I think this is a problem with how the 11ty plugin or Prism load (and perhaps cache) languages. If you remove all other blog posts and just keep one with a ` ```diff-js` that gets highlighted correctly. When there are other posts as well though (so there were blocks before that one that got highlighted for the `js` language), it stops working and only highlights the Diff.

beerinho:
That sounds like a real pain, do you have any thoughts on how to move forward with it? From a styling perspective I feel that `js/0/2` looks the best (because it uses `del` and `ins` elements that wrap the whole content), but it's just a pain to have to count the lines

marcoow:
Yeah, I don't think that's an option at all – let's see whether I can find out more, otherwise we'll just have to go with ` ```diff` for now.

marcoow:
I fixed it – the problem is you cannot use ` ```diff` **and** ` ```diff-js` (or any other language-specific diff grammar) at the same time because of [this line](https://github.com/11ty/eleventy-plugin-syntaxhighlight/blob/master/src/PrismLoader.js#L30). The first time you highlight anything with `diff`, Prism loads the language grammar for that and from then on knows that language. However, that's the `prism-diff` grammar then and since it's known, 11ty won't load the `diff-highlight` plugin so that that won't register it's callbacks etc.…

marcoow:
see https://github.com/11ty/eleventy-plugin-syntaxhighlight/issues/76

marcoow (approved):

beerinho:
Just to confirm, what do you mean when you say you fixed it? Are we just accepting that we can't have both `diff` _and_ `diff-js` highlighting?

marcoow:
I'd say for now, yes

beerinho:
Okay, thanks. 
This is my first PR for the mainmatter website, what happens next? Do I as the author merge this PR? And is there anything I need to do about the failing checks?

marcoow:
sure, feel free to merge – you can ignore the failed Netlify build in this case as that's caused by someone impersonating @simplabs which broke the repo redirects.

PR #1956, opened 2022-11-30
Fix text on product development page

marcoow:
I realized we were saying some things that were just wrong on the page for digital product development – this fixes that:

Bildschirm­foto 2022-11-30 um 08 22 04

### DIFF
--- a/src/services/digital-products-and-design.njk
+++ b/src/services/digital-products-and-design.njk
@@ -38,7 +38,7 @@ title: Digital Products & Design | Our Services
 {% set content = {
   "title": "Product design",
   "subtitle": "Experiences that users love",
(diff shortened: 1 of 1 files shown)


SaeLLo (commented):
I am not sure if it got changed....It should not be a numeric list but a list with arrows like in the content blocks below.

marcoow:
only the text changed: https://github.com/mainmatter/mainmatter.com/pull/1956/files

SaeLLo:
Okay. 

Do I have to open a new request so the numeric list view gets changed into an arrow list?

marcoow:
yes please :)

SaeLLo (approved):

PR #1960, opened 2022-12-05
Conditional header in mainmatter.com

candunaj:
This PR makes the header
image
visible only when you visit the page from simplabs.com. When you visit the page from any other domain, the header will be removed by a netlify edge function.

# How conditionalHeader edge function works

When the user navigates to simplabs he will be redirected to mainmatter based on
redirect rules set in `netlify.toml`. It is redirect to a different domain so
the only way how we can sent an information that header should be visible is by
setting a query param `simplabs=true`. When user navigates to a different link
on mainmatter, the query param is removed, so when there is a query param
`simplabs=true` session cookie `simplabs=true` is set. So `conditionalHeader`
edge function checks if the session cookie `simplabs=true` is present, if it is,
the header is not removed.

## How to test `conditionalHeader` edge function

1. add `simplabs.com` and `mainmatter.com` to your /etc/hosts file
2. Uncomment the redirect rule in `netlify.toml` file
   > note: it will redirect `http://simplabs.com:8888/contact` to
   > `http://mainmatter.com:8888/` that way we can test redirect to a different
   > domain

```
[[redirects]]
  from = "/contact"
  to = "http://mainmatter.com:8888/"
  status = 301
  force = false
```

3. run `netlify dev`
4. Navigate to [simplabs.com:8888/contact](http://simplabs.com:8888/contact)
5. You will be redirected to `mainmatter.com:8888` with a query param
   `simplabs=true` and header will be wisible
6. Navigate to any other link on `mainmatter.com:8888`, the header will be visible
   until the browser is closed and navigated to mainmatter.com:8888 again

For debugging, run the following command
`netlify dev --edgeInspect=127.0.0.1:9229`, then open chrome and navigate to
`chrome://inspect`.


netlify/edge-functions/conditionalHeader.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+}
+
+export default async (request, context) => {
+  // 1. start loading response text and wasm in parallel
+  const promises = [loadText(context), init()];
+
+  // 2. while promises are loading, do some work
+  // month is 0-indexed
+  const headerShownUntil = BigInt(new Date(2024, 0, 1).getTime());
```
marcoow: I think we could leave this out even – we can just sent a reminder to switch this off at a particular date

netlify/edge-functions/conditionalHeader.js
```diff
@@ -0,0 +1,36 @@
...
+  const url = new URL(request.url);
+  const isMainmatter = url.hostname === "mainmatter";
```
marcoow: I'd invert that and check for whether the hostname is simplabs. We might have additional domains at some point for which we don't necessarily want the teaser to show and that way we make the code more resilient to these kind of situations.

netlify/rust-rewriter/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+wasm-bindgen = "0.2"
+chrono = { version = "0.4", features = ["unstable-locales"] }
+kuchiki = "0.8.1"
```
marcoow: 👀

netlify/rust-rewriter/pkg/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "version": "0.1.0",
+  "files": [
+    "rust_rewriter_bg.wasm",
+    "rust_rewriter.js",
+    "rust_rewriter.d.ts"
+  ],
+  "module": "rust_rewriter.js",
+  "types": "rust_rewriter.d.ts",
+  "sideEffects": false
+}
```
marcoow: 👀

netlify/rust-rewriter/pkg/README.md
```diff
@@ -0,0 +1,10 @@
+# Build command
+
+```
+wasm-pack build --target web
+```
+
+## Run tests with debug output
+```
+cargo test -- --nocapture
+```
```
marcoow: The README seems to be duplicated?

netlify/rust-rewriter/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+    let is_in_past = date < now;
+    return is_in_past || is_mainmatter;
+  }
+  
+  return false;
+
+}
+
+fn remove_element(id: &str, html: &str) -> String {
+  let document = kuchiki::parse_html().one(html);
```
marcoow: I think we could maybe make this simpler and faster with just some special comments like `<!-- SIMPLABS_HEADER_START -->`, `<!-- SIMPLABS_HEADER_END -->` and we just remove all text in between those? That way we don't need to parse the HTML which I assume could be slow(ish) even in Rust?

marcoow (commented):
Nice! Can we try this locally or for a PR somehow?

Also should the `pkg` folder be committed or is that auto-generated by wasm-pack at build time?

netlify/edge-functions/conditionalHeader.js
candunaj: Thank you for the suggestion, I am going to check the hostname to see if it is simplabs. And I will remove the date condition.

netlify/rust-rewriter/Cargo.toml
candunaj: Chrono is for working with dates. I can remove it because date condition will be removed. Kuchiki is for parsing and working with HTML. I am removing element by selector.

netlify/rust-rewriter/pkg/package.json
candunaj: pkg was generated by wasm-pack. I have included it as part of this repo and was able to import the wasm file. Do you think that rust_rewriter should be in a separate repo?

netlify/rust-rewriter/pkg/README.md
candunaj: wasm-pack generate pkg directory and it copy README.md.

netlify/rust-rewriter/src/lib.rs
candunaj: You are right it will definitely be faster. I am going to implement it.

netlify/rust-rewriter/Cargo.toml
marcoow: sorry for the confusion – I meant missing newline at EOF

netlify/edge-functions/conditionalHeader.js
```diff
@@ -0,0 +1,41 @@
+import init, {
+  remove_header_start_end,
+} from "https://unpkg.com/rust-rewriter@1.2.0/rust_rewriter.js";
```
marcoow: What's this?

marcoow (commented):
Is all of the Rust code now gone from this repo?

Also, I'd be interested how we can test this? Is there a local setup?

netlify/edge-functions/conditionalHeader.js
candunaj: Netlify edge function can import external packages only directly from a URL. It can't be installed as a dependency in package.json.

https://deno.land/manual@v1.28.3/examples/manage_dependencies#managing-dependencies

candunaj:
> Is all of the Rust code now gone from this repo?
> 
> Also, I'd be interested how we can test this? Is there a local setup?

- Yes rust code is in separate package rust-rewriter https://github.com/mainmatter/rust-rewriter
- You can test it with the command ```netlify dev```. It will open the page on localhost:xxx so the domain is not simplabs and header will be removed.

netlify/edge-functions/conditionalHeader.js
marcoow: Hm, is there no way we can have the Rust code in this repo as well? It seemed to work before?

Also are Netlify edge functions running on Deno? I was assuming V8…

candunaj: - I will move the rust code back to this repo.
- I have found link where it is explained, that netlify edge functions are build on deno: https://www.netlify.com/blog/announcing-serverless-compute-with-edge-functions/

.gitignore
```diff
@@ -26,3 +26,9 @@ doc/
 _site
 .DS_Store
 .sass-cache
+
+# vscode
+.vscode
+
+# rust-rewriter
+**/rust-rewriter/target
```
marcoow:
```suggestion
**/rust-rewriter/target
**/rust-rewriter/build
```

Should this not be ignored as well?

README.md
```diff
@@ -7,3 +7,24 @@ The source code for [https://mainmatter.com](https://mainmatter.com).
 Copyright &copy; 2019-2022 Mainmatter GmbH (https://mainmatter.com), released
 under the
 [Creative Commons Attribution-NonCommercial 4.0 International license](https://creativecommons.org/licenses/by-nc/4.0/).
+
+## "Smartroom is mainmatter" header is removed by netlify edge function
```
marcoow:
```suggestion
## "simplabs is mainmatter" header is removed by netlify edge function
```

netlify/edge-functions/rust-rewriter/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+      assert_eq!(result, "<html><head></head><body><div id='simplabsheader'>simplabs is Mainmatter now!</div></body></html>");
+    }
+
+    #[test]
+    fn it_do_nothing_if_element_doesnotexist() {
+      let response_text = "<html><head></head><body></body></html>";
+      let result = remove_element("#simplabsheader", response_text);
+      assert_eq!(result, "<html><head></head><body></body></html>");
+    }
+}
```
marcoow: I don't think the tests are in sync with the latest code?

marcoow: Let's also add these tests to our CI setup please – that way you'd have noticed this I think?

package.json
```diff
@@ -13,6 +13,10 @@
     "build:html": "eleventy",
     "build:js": "rollup --config ./rollup.config.js",
     "build:sass": "sass src/assets/css/app.scss ./assets/css/app.css --style=compressed",
+    "rust-rewriter:remove-files": " rm ./netlify/edge-functions/rust-rewriter/build/.gitignore && rm ./netlify/edge-functions/rust-rewriter/build/package.json",
+    "rust-rewriter:ignore": "echo '/* eslint-disable */' | cat - ./netlify/edge-functions/rust-rewriter/build/rust_rewriter.js > temp && mv temp ./netlify/edge-functions/rust-rewriter/build/rust_rewriter.js",
```
marcoow: Do we need this as a script?

package.json
```diff
@@ -13,6 +13,10 @@
     "build:html": "eleventy",
     "build:js": "rollup --config ./rollup.config.js",
     "build:sass": "sass src/assets/css/app.scss ./assets/css/app.css --style=compressed",
+    "rust-rewriter:remove-files": " rm ./netlify/edge-functions/rust-rewriter/build/.gitignore && rm ./netlify/edge-functions/rust-rewriter/build/package.json",
```
marcoow: I guess this can be removed altogether when we ignore the `build` directory?

package.json
candunaj: wasm-pack generates the build directory prepared to be deployed as an npm package. In our case, a js file from that directory is directly imported in the netlify edge function. This is why I believe that the build directory needs to be committed to GitHub. We are not deploying the package to npm, so I created a script to remove package.json and gitignore from the build directory. There were also some eslint errors in generated rust_rewriter.js file so I have added eslint-disable at the top.

package.json
candunaj: web-pack generates the build directory where is the wasm and js file. Js file is imported into the netlify edge function. So this is why I think that build needs to be also committed.

netlify/edge-functions/rust-rewriter/src/lib.rs
candunaj: Tests are successfully running and in sync with the code. There are two functions exported from the wasm file. remove_header will remove an element by Id, and remove_header_start_end remove text in between two texts. The first function is no longer used, but I left it there in case we decide to remove an element in the future by Id. Would you like me to remove that function?

README.md
candunaj: Thank you. Smartroom was emberjs application I was working on a few years.

.gitignore
candunaj: wasm-pack generates the build directory. There is a rust_rewriter.js file from the build directory imported to the netlify edge function. This is why I believe the build directory needs to be also committed.

netlify/edge-functions/rust-rewriter/src/lib.rs
candunaj: I have now added tests to CI.

package.json
marcoow: Can't we run `wasm-pack` during deployment of the Netlify function?

netlify/edge-functions/rust-rewriter/src/lib.rs
marcoow: > but I left it there in case we decide to remove an element in the future by Id

I don't think we need that

package.json
candunaj: I am not as experienced with netlify, but thank you for the suggestion. I am going to run wasm-pack during the deployment.

netlify/edge-functions/rust-rewriter/src/lib.rs
candunaj: Yes, you are right, I will remove unnecessary bits from the code.

marcoow:
@candunaj running `netlify dev` doesn't seem to work?

```
◈ Starting Netlify Dev with Eleventy
$ npm run build:rust-rewriter && npm run watch
npm WARN lifecycle The node binary used for scripts is /var/folders/1g/y9k16qbd16nft_y0cvn_zgq40000gn/T/yarn--1671454841513-0.0006426149415563742/node but npm is using /Users/marcoow/.volta/tools/image/node/14.16.1/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.

> mainmatter-website@1.0.0 build:rust-rewriter /Users/marcoow/Code/mainmatter.com
> wasm-pack build ./netlify/edge-functions/rust-rewriter --target web

You must install wasm-pack before you can run it
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! mainmatter-website@1.0.0 build:rust-rewriter: `wasm-pack build ./netlify/edge-functions/rust-rewriter --target web`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the mainmatter-website@1.0.0 build:rust-rewriter script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

marcoow:
seems to be a known issue: https://github.com/rustwasm/wasm-pack/issues/1189

netlify/edge-functions/conditionalHeader.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+}
+
+export default async (request, context) => {
+  // 1. start loading response text and wasm in parallel
+  const promises = [loadText(context), init()];
+
+  // 2. while promises are loading, do some work
+  const url = new URL(request.url);
+  const isSimplabs = url.hostname === "simplabs";
```
marcoow: `hostname` includes the TLD so it's `"simplabs.com"` actually and this will never be true. It should cover both simplabs.com and simplabs.de

marcoow (commented):
I'm not 100% sure how to test this. Downgrading `wasm-pack` to 0.10.1 prevents the issue (see rustwasm/wasm-pack#1189) but the problem is the redirects don't seem to work when running `netlify dev`?

So while I can add

```
127.0.0.1	simplabs.com
127.0.0.1	mainmatter.com
```

to my `/etc/hosts` and then I see the banner at http://simplabs.com:8888 and not at http://mainmatter.com:8888 (that is, after changing the check for `hostname` to check for `"simplabs.com"`), what I really want to see is simplabs.com redirecting to mainmatter.com and then showing the banner there while it does not show when going to mainmatter.com directly.

How did you validate this?

netlify/edge-functions/conditionalHeader.js
candunaj: Thank you for highlighting this. It is fixed now.

candunaj:
I wasn't able to test the redirect from simplabs to mainmatter directly, but I have found way how to test it. I have described it in the description of this PR and also in the project `netlify/edge-functions/conditionalHeader.md`.

.gitignore
marcoow: No, it's a build artifact that will be re-created each time based on the code –no need to commit that.

netlify.toml
```diff
@@ -6,6 +6,12 @@
 [build]
   environment = { NODE_VERSION = "16.17.0" }
 
+# rule for testing conditionalHeader edge function redirect from simplabs to mainmatter
+#  [[redirects]]
+#    from = "/simplabs.com"
+#    to = "/"
+#    force = true
```
marcoow: Redirecting a path is a pretty different thing from redirecting a domain – don't think we should rely on that for testing.

marcoow (commented):
@candunaj I made some smaller fixes. Testing still remains a problem and I don't think we can be sure this works because:

I understand why the cookie is needed as you'll want the header to be shown on subsequent requests in the same session as well although those would be served from mainmatter.com. However, when you're testing the cookie functionality by redirecting a path, you're setting the cookie for the same domain the request came in through (localhost in your case). However, when redirecting the domain, the requests comes in for simplabs.com but you want to set the cookie for mainmatter.com (but you cannot do that from the response to a request to simplabs.com as you cannot set cookies cross-domain). I think figuring out a way to test this locally is essential to understand whether this actually works.

You can serve simplabs.com and mainmatter.com from `netlify dev` by pointing the domains to localhost via `/etc/hosts` btw:

```
127.0.0.1 simplabs.com
127.0.0.1 mainmatter.com
```

You can also run `netlify dev` on port 80 and via HTTPS so generally this should be possible to test locally.

candunaj:
I have found that netlify bundles the netlify edge functions before the build command from package.json runs. This is why there is an error:

```
Could not find file: file:///opt/build/repo/netlify/edge-functions/rust-rewriter/pkg/rust_rewriter.js
```

https://app.netlify.com/sites/objective-northcutt-37494c/deploys/641429e7063b99000816b986#L102

This is why I believe that the right solution is to have rust rewriter as a separate package. Then It can be imported into the netlify edge function without any problem.

marcoow:
sure - separate package is fine I guess 👍

candunaj:
I don't have access to percy/mainmatter.com, so I can't see why it is failing. Probably because the header is removed.

marcoow:
invited you to Percy

PR #1962, opened 2022-12-10
[WIP] Blog post about Implementing solution to show header only when redirected from simplabs.com

candunaj:
The blog post about implementing a solution to show the "simplabs is mainmatter now!" header on mainmatter.com when the user is redirected from simplabs.com. This was achieved by using a Netlify edge function to check the URL of the incoming request and, if it is from mainmatter.com, calling a WebAssembly (wasm) function written in Rust to remove the header. The Rust code was chosen for its performance and static typing and was compiled to wasm for portability and execution on the web.


src/posts/2022-12-09-rewrite-responses-with-edge-function-and-rust.md
```diff
@@ -0,0 +1,218 @@
...
+authorHandle: candunaj
+tags: rust
+bio: "Stanislav Dunajčan"
+description:
+  "Stanislav Dunajčan exploring potential of webassembly (wasm) written in Rust
+  and netlify edge functions"
+og:
+  image:
+tagline: |
+  <p>Implemented a solution to show the "simplabs is mainmatter now!" header on mainmatter.com when the user is redirected from simplabs.com. This was achieved by using a Netlify edge function to check the URL of the incoming request and, if it is from mainmatter.com, calling a WebAssembly (wasm) function written in Rust to remove the header. The Rust code was chosen for its static typing and performance and compiled to wasm for portability and execution on the web.</p>
```
marcoow, line 12: This is very specific already. But since this is the intro, put yourself in the shoes of the reader: they have 0 context, don't know neither simplabs nor Mainmatter and might also not care much. You'll want to make the intro so that someone who has never heard of us can still get excited about the blog post as they care about the general topic which is rewriting responses on an edge function with Rust – what's being rewritten specifically and why isn't a relevant aspect there necessarily.

src/posts/2022-12-09-rewrite-responses-with-edge-function-and-rust.md
```diff
@@ -0,0 +1,218 @@
...
+---
+
+When a user visits mainmatter.com, they see a header that says "simplabs is
+mainmatter now!". The task was to show this header only when the user visits the
+page from simplabs.com and is redirected to mainmatter.com. There are several
```
marcoow, line 17: Might make sense to share that the company was recently renamed and link to the post so it's clear why we're even doing this. Also might make sense to explain why we're not doing this with JS (since we want to serve static pages if possible and also we like Rust (link to `/rust-consulting/` and don't want to let an opportunity to use it pass unused))

src/posts/2022-12-09-rewrite-responses-with-edge-function-and-rust.md
```diff
@@ -0,0 +1,218 @@
...
+2. Compile the Rust code to WebAssembly (wasm) to make it portable and
+   executable on the web.
+3. Create and set up a Netlify edge function to remove the header when
+   appropriate.
+
+### 1. Implement a function in Rust to remove the header, taking advantage of the language's performance and static typing.
+
+#### Creating the rust library
+
+To create the rust library, we first need to use the `wasm-pack` tool with the
```
marcoow, line 35: > To create the rust library

for creating a Rust library, you'd usually use `crate`, `wasm-pack` is used here since this is WASM – might be good to mention that to make it clear

src/posts/2022-12-09-rewrite-responses-with-edge-function-and-rust.md
```diff
@@ -0,0 +1,218 @@
...
+
+To implement the function that removes the header when visited from a domain
+other than simplabs.com, we have two options:
+
+1. Parse the response text into HTML using a library like kuchiki, and then
+   remove the element by ID.
+2. Add specific comments before and after the header, and then remove everything
+   in between.
+
+We choose the second option because it is faster than parsing the response text.
```
marcoow, line 54: I'd add sth. like _"…and we didn't need the additional flexibility that parsing the HTML response would have allowed for"_ or so

src/posts/2022-12-09-rewrite-responses-with-edge-function-and-rust.md
```diff
@@ -0,0 +1,218 @@
...
+pub fn remove_header_start_end(response_text: &str, start_text: &str, end_text: &str, is_simplabs: bool) -> String {
+  if is_simplabs {
+    return response_text.to_string();
+  }
+
+  let start_index = response_text.find(start_text);
+  let end_index = response_text.find(end_text);
+
+  if let Some(si)=start_index {
+    if let Some(ei)=end_index {
```
marcoow, line 70: Might be able to combine these 2 like

```rs
if let (Some(si), Some(ei)) = (start_index, end_index) {
  …
```

src/posts/2022-12-09-rewrite-responses-with-edge-function-and-rust.md
```diff
@@ -0,0 +1,218 @@
...
+`netlify.toml` config file. We do this by adding the following code to the file:
+
+```toml
+[[edge_functions]]
+  path = "/*"
+  function = "conditionalHeader"
+```
+
+This tells Netlify to run the `conditionalHeader` function for every incoming
+request to any path on the site.
```
marcoow, line 153: It might be helpful to add here that Netlify will look for a file with a particular name in a particular folder then (or I assume that's how it works)?

src/posts/2022-12-09-rewrite-responses-with-edge-function-and-rust.md
```diff
@@ -0,0 +1,218 @@
...
+
+To remove the header from the HTML response, we first need to add specific text
+markers above and below the header. This allows us to easily find and remove the
+header using the `remove_header_start_end` function from the wasm file.
+
+Here's an example of what the HTML with the text markers might look like:
+
+```html
+<!-- SIMPLABS_HEADER_START -->
+Header to be removed
```
marcoow, line 165: I'd add a bit of the actual HTML here – might make it clearer

marcoow (commented):
Nice 🙌 I left some comments mostly regarding stylistic aspects but I think this is a good overview

src/posts/2022-12-09-rewrite-responses-with-edge-function-and-rust.md
candunaj: I was searching for how to combine these two but did not find an example. Thank you for the suggestion!

src/posts/2022-12-09-rewrite-responses-with-edge-function-and-rust.md
candunaj: Thank you for the suggestion.

emmasofiah2o:
@candunaj can you please give me a date when this would be ready? I have been pushing release for weeks now 😅

PR #1963, opened 2022-12-15
Update 2022-11-18-sveltekit-super-rentals.md

BobrImperator:


beerinho:
We decided to just go with the standard `diff` styling for now, ([PR](https://github.com/mainmatter/mainmatter.com/pull/1955)) so I'm going to close this. Thanks for your input anyway @BobrImperator

marcoow:
> We decided to just go with the standard diff styling for now, (https://github.com/mainmatter/mainmatter.com/pull/1955) so I'm going to close this.

That's not really the resolution of #1955 @beerinho – the solution here is that if you **never** use `diff` and **always** use language specific diff-styling (e.g. `diff-js`) it works as intended. Once 11ty/eleventy-plugin-syntaxhighlight#76 gets fixed, that limitation might get resolves – just adding this here for clarification and to avoid misunderstandings

PR #1973, opened 2023-01-12
`fix(htmlmin)`: escape when double `'<<'` with `'&lt;<'`

oscard0m:
# Description

`html-minifier` is receiving html with `'<'` characters sorrounded by a `span` tag. This results in `<span><<span/>`. 

We need to escape the `'<'` character to bypass the issue. To do that, I use the equivalent code: `&lt;`.

# Context
Unblocks https://github.com/mainmatter/mainmatter.com/pull/1952

When using `diff-js` syntax-highlighting and the highlighted code contains an HTML `<script>` tag, we experience this issue.

# Alternatives I have tried
Eleventy uses [`Prism`](https://prismjs.com/) to generate syntax highlights. Checking [the supported languages by `Prism`](https://prismjs.com/#supported-languages), we can see `Svelte` is not one of them.

The alternatives I tried are:

## 1. `diff-html` of `diff-jsx` syntax highlight
I have tried using `diff-html`, but the resulting CSS colours are worse than the ones we have when using `diff-js`.

Same for `diff-jsx`.

## 2. [`prism-svelte`](https://github.com/pngwn/prism-svelte/)
I tried to apply [`prism-svelte`](https://github.com/pngwn/prism-svelte/), but the results were not satisfactory in terms of colours, and it did not solve the issue with unescaped `<<` char sequence.

## 3. [`Torchlight`](https://torchlight.dev/#pricing?)
[`Torchlight`](https://torchlight.dev/#pricing?) is an alternative to `Prism`. They promise to support all the syntax highlights supported by VSCode (this includes `Svelte`), and there is an existing [plugin for Eleventy](https://www.npmjs.com/package/eleventy-plugin-torchlight).

The lowlight is this plugin [is not free](https://torchlight.dev/#pricing). There is a free plan for OSS, but I don't think we apply, considering this is Mainmatter's website.


BobrImperator (approved):

utils/transforms/htmlmin.js
```diff
@@ -2,7 +2,9 @@ const htmlmin = require("html-minifier");
 
 module.exports = function (content, outputPath) {
   if (outputPath && outputPath.endsWith(".html")) {
-    let minified = htmlmin.minify(content, {
+    const sanitizedContent = content.replaceAll("<<", "&lt;<");
```
oscard0m, line 5: > wouldn't the fix be to replace < and > in code snippets before the highlighted code gets passed to html minifier? (as in doesn't the proposed fix fix the symptom intsead of the cause?)
[via @marcoow [in Slack](https://mainmatter.slack.com/archives/C0DBS7FCM/p1674129471255939?thread_ts=1674128000.582309&cid=C0DBS7FCM)]


> well, it’s more intuitive for the writer to just use << in the text.
I agree with you too though, as long as someone remember that this could break :smile:
[via @BobrImperator  [in Slack](https://mainmatter.slack.com/archives/C0DBS7FCM/p1674129790852629?thread_ts=1674128000.582309&cid=C0DBS7FCM)]

@marcoow with your comment it's not clear to me where you want to apply the fix:
a) At Eleventy plugin level.
b) At Blog post level.

marcoow: I assume if the blog post contains e.g. `&lt;<` in diffs as opposed to `<<`, the highlighting wouldn't work?

I guess the correct place to implement this would be as a wrapper around the syntax highlighting plugin (see https://github.com/mainmatter/mainmatter.com/blob/45944ec42cbd9d8317e6036b10f6710edb7a953b/.eleventy.js#L147) so that if we receive highlighted HTML that contains `<` or `>`, we replace with `&lt;`/`&gt;`?

marcoow: Of course as a quickfix we can also disable htmlnifier…

oscard0m: > I assume if the blog post contains e.g. `&lt;<` in diffs as opposed to `<<`, the highlighting wouldn't work?

~I didn't try~ I just drafted a PR (https://github.com/mainmatter/mainmatter.com/pull/1976), and it doesn't work.
I don't think we want writers to be aware of this bug and write escaped code in their code snippets when writing a blog post.

> I guess the correct place to implement this would be as a wrapper around the syntax highlighting plugin (see
> 
> https://github.com/mainmatter/mainmatter.com/blob/45944ec42cbd9d8317e6036b10f6710edb7a953b/.eleventy.js#L147
> 
> ) so that if we receive highlighted HTML that contains `<` or `>`, we replace with `&lt;`/`&gt;`?

The root cause of this issue is we are trying to highlight a code snippet with `diff-js`, but the code snippet contains `JS` and `HTML`. The right solution would be to implement support for `Svelte` or [fix the existing one](https://github.com/pngwn/prism-svelte/). [Torchlight](https://torchlight.dev/#pricing?) solves this issue but is not free (see PR description).

As a hot-fix, I did not find a way to hook a transformation for [eleventy-plugin-syntaxhighlight](https://github.com/11ty/eleventy-plugin-syntaxhighlight). I checked the `options` of this plugin[^1][^2], but I don't see they offer one. I can research more about it, but it will take me time. 

**Any idea how to hook the wrapper you propose and make it receive the `HTML content` to be sanitized?**

[^1]: https://github.com/11ty/eleventy-plugin-syntaxhighlight/blob/master/.eleventy.js#L18
[^2]: https://www.11ty.dev/docs/plugins/syntaxhighlight/

> Of course as a quickfix we can also disable htmlnifier…

I think this solution is not good. We would punish the performance for all the website pages to skip this issue. As a quick fix, my proposed solution does the job without punishing performance for end users (only an extra step at compile time of eleventy).

marcoow: HTML minification is a totally minor thing when it comes to frontend performance so disabling that doesn't really have a relevant impact.

What I meant by writing a wrapper is sth. like (no idea what the actual API is but I guess it's something like the following):

```js
eleventyConfig.addPlugin({
  callPlugin(input) {
    const output = syntaxHighlightPlugin(input);
    return output.replaceAll("<span><<span/>", "<span>&lt;<span/>").replaceAll("<span>><span/>", "<span>&gt;<span/>")
  }
}); 
```

It's not really worth making a huge deal of this though – if the blog post ends up being blocked without this, fine let's merge but we should eventually clean this up properly.

oscard0m: > HTML minification is a totally minor thing when it comes to frontend performance so disabling that doesn't really have a relevant impact.

I opened a PR (https://github.com/mainmatter/mainmatter.com/pull/1977) removing the `html-minifier` transformation for the website. Feel free to use it if this is the alternative you want to use.

> What I meant by writing a wrapper is sth. like (no idea what the actual API is but I guess it's something like the following):
> 
> ```js
> eleventyConfig.addPlugin({
>   callPlugin(input) {
>     const output = syntaxHighlightPlugin(input);
>     return output.replaceAll("<span><<span/>", "<span>&lt;<span/>").replaceAll("<span>><span/>", "<span>&gt;<span/>")
>   }
> }); 
> ```

I've not been able to hook something easily. The plugin does not expose the `markdownPrismJs()` method [used in the `configFunction`](https://github.com/11ty/eleventy-plugin-syntaxhighlight/blob/ac92151c42996faef19650f34f54a77e4ac210f2/.eleventy.js#L45) so I don't see an easy way to do this without copy =-pasting the content of the library.

> It's not really worth making a huge deal of this though – if the blog post ends up being blocked without this, fine let's merge but we should eventually clean this up properly.

I still prefer the solution proposed in this PR as a hot fix and we can do deeper research into existing plugins without losing minified HTML but feel free to use the one you prefer.

oscard0m:
Solved via https://github.com/mainmatter/mainmatter.com/pull/1979

PR #1977, opened 2023-01-19
remove html minifier

oscard0m:
- feat(eleventy): remove 'html-minifier' as code transformation


marcoow:
See https://github.com/mainmatter/mainmatter.com/pull/1952#issuecomment-1398025018 – there seems to be a simpler solution…

oscard0m:
Solved via https://github.com/mainmatter/mainmatter.com/pull/1979

PR #1952, opened 2022-11-18
Add sveltekit super rentals blog post

beerinho:
This PR adds a new blog post that goes through building the Ember super rentals tutorial app with SvelteKit.

If there is a way to have both a title and subtitle, I'd like to change these around to be:
title - "SvelteKit super rentals"
subtitle - "A practical first look at the new Svelte framework"

Reviewers be warned, it's a long one!

Also, @emmasofiah2o, do you normally add the og images for these? Or how does that work?

## TODO

- [x] add og:image
- [x] add header image

PR #1979, opened 2023-01-20
run content parser scripts before minification

marcoow:
This runs the `contentParser.js` script that does a number of modifications on the HTML that 11ty generates before the minification step. While it makes sense to only minifiy at the very end, this also seems to fix the build problem that @beerinho encountered in #1952 for some reason…


beerinho (approved):

beerinho:
Very odd, thanks for sorting this @marcoow, once this is merged I'll rebase my branch

PR #2010, opened 2023-03-16
Add ElixirConf

emmasofiah2o:
- Add event
- Add banner

Screenshot 2023-03-16 at 16 00 37


emmasofiah2o:
@marcoow feel free to change the text if you need to. I will only be able to merge this on the 27th cause I am off on vacation until then so please take over if you would like this to be published before 🙏🏻.

marcoow (approved):

PR #2014, opened 2023-03-22
Enable gravity

marcoow:
…so we can get more real-world test coverage while completing the service


marcoow:
(Gravity check remains pending here because we don't have the baseline established yet)

oscard0m (approved):

PR #2015, opened 2023-03-22
Optimize CI

marcoow:
We're doing a bunch of things more than once and run some steps even if a fundamental one (e.g. linting) fails which makes things slower than they need to be.

This runs linting before everything else (and does not run anything else if linting fails) and caches the build between build steps.


marcoow:
this is slower than just building again and again…

PR #2022, opened 2023-04-12
Add This week in os #22

emmasofiah2o:
- Add file 

To-do:
- Categorize Unknown links


src/twios/2023-04-18.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Octokit
+
+- [octokit/app.js] [#405](https://github.com/octokit/app.js/pull/405) WIP
+  build(deps): upgrade @octokit/webhooks to 10.7.1 (next) ([@oscard0m])
+- [octokit/graphql-action]
+  [#207](https://github.com/octokit/graphql-action/pull/207) migrate zeit ncc to
+  vercel ncc ([@oscard0m])
+- [octokit/webhooks.js] [#835](https://github.com/octokit/webhooks.js/pull/835)
+  Next ([@oscard0m])
```
oscard0m:
```suggestion
```

oscard0m (changes requested):

src/twios/2023-04-18.md
emmasofiah2o: Why the double dash? :)

oscard0m: I'm suggesting removing these 2 lines.

The way this view works (it's called `diff`) is by adding `-` in front of each line of code removed and adding `+` in front of each line of code added.

See this example:

```diff
+ line 1 added
+ line 2 added
- line 3 removed
- line 4 removed
```

In this case, the double dash is because the line deleted (`-`) starts with a dash:

```diff
- - line deleted starting with a dash
```

emmasofiah2o: Ah I see. Is this repo to be excluded from all posts? In that case I would update the config

oscard0m: No, only this PR

emmasofiah2o:
@oscard0m could you reapprove Gravity if the changes are fine to you?

oscard0m (approved):

PR #2037, opened 2023-05-09
CLI Tricks for GitHub

emmasofiah2o:
- Add blog post

To-do:
- [ ]  Add og-image
- [ ]  Add illustration


emmasofiah2o:
@marcoow Bart had this idea about a TIL post a while back. I have helped him a little structuring the post and this is what came out. It's short, but I think it could work. What do you say?

src/posts/2023-05-15-cli-tricks.md
```diff
@@ -0,0 +1,48 @@
+---
+title: "CLI Tricks for GitHub"
```
BobrImperator:
```suggestion
title: "CLI Tricks for Git with FZF"
```

Github is a service that hosts Git repositories.
Some people could be using e.g. Gitlab or have a self-hosted solution to host their repositories but still use `Git` as their version control tool.

src/posts/2023-05-15-cli-tricks.md
```diff
@@ -0,0 +1,48 @@
...
+authorHandle: BobrImperator
+tags: [misc]
+bio: "Bartłomiej Dudzik"
+description:
+  "Bartłomiej Dudzik shows a trick to checkout a GitHub branch faster in the
+  CLI."
```
BobrImperator:
```suggestion
  "Bartłomiej Dudzik shows a trick to checkout a Git branch faster in the
  CLI."
```

src/posts/2023-05-15-cli-tricks.md
```diff
@@ -0,0 +1,48 @@
...
+og:
+  image:
+tagline: |
+  <p>Are you looking to be more efficient with Git while wokring with the CLI? There is an easy way to make checking out branches a one-step process.</p>
```
BobrImperator:
```suggestion
  <p>Are you looking to be more efficient with Git while working with the CLI? There is an easy way to make checking out branches a one-step process.</p>
```

src/posts/2023-05-15-cli-tricks.md
```diff
@@ -0,0 +1,48 @@
...
+
+image: ""
+imageAlt: ""
+---
+
+Does your current process to checkout a GitHub branch look something like this?
```
BobrImperator:
```suggestion
Does your current process to checkout a Git branch look something like this?
```

src/posts/2023-05-15-cli-tricks.md
```diff
@@ -0,0 +1,48 @@
...
+- `git branch` or `git branch --list`
+- outputs a non-interactive list of branches
+- Double-click an entry
+- do `CTRL-C` or right click and copy
+- close the list
+- paste the entry to the terminal
+- prepend entry with `git checkout`
+
+Going through these many steps decreases productivity and makes working with
+GitHub a real struggle. The good news is, there is a way to work around this and
```
BobrImperator:
```suggestion
Git a real struggle. The good news is, there is a way to work around this and
```

src/posts/2023-05-15-cli-tricks.md
```diff
@@ -0,0 +1,48 @@
...
+make the whole process much leaner!
+
+Using FZF, the seven steps listed above become a one-liner:
```
BobrImperator: FZF is a tool that is usually not included on new systems so people will have to install it themselves
```suggestion
Using [FZF](https://github.com/junegunn/fzf), the seven steps listed above become a one-liner:
```

src/posts/2023-05-15-cli-tricks.md
```diff
@@ -0,0 +1,48 @@
...
+
+`~/.bashrc`
+
+`alias gb="git branch | fzf | xargs git checkout"`
+
+You should now be able to use this with a `gb` shorthand, making your experience
+smoother. Happy coding!
+
+If you are looking for a guide on using Fork to manage your Git actions, check
+out this blog post by Chris Manson:
```
BobrImperator:
```suggestion
If you are looking for a guide on using Git with visual tools like Fork, check out this blog post by Chris Manson:
```

src/posts/2023-05-15-cli-tricks.md
marcoow: Checking out branches is a one step process?

```bash
git checkout <branch>
```

marcoow (commented):
I'm not sure this is a blog post tbh 😬 It's definitely another signal that having the TIL would be cool though!

We can make it into a blog post if we want to but I think we'd have to explain a bit better what this is about. I didn't get it the first time I read it (see my comment)

emmasofiah2o:
> I'm not sure this is a blog post tbh 😬 It's definitely another signal that having the TIL would be cool though!
> 
> We can make it into a blog post if we want to but I think we'd have to explain a bit better what this is about. I didn't get it the first time I read it (see my comment)

I see. Would you recommend turning it into a blog post or rather using it in another way (for example LinkedIn article, but that way we only post on one platform)? I am trying to understand if it makes sense to invest more time into this since I would also need Bart's support.

BobrImperator:
It really feels like the tagline is the only issue here.
IMO checking out a branch is not a one-step process if you don't know the name of it or at least the ticket number it relates to (if it has one). I'd often forget the ticket number, or knew only a part of the name that the usual autocompletion wouldn't pick up easily. While the script from the post gives me an ability to have a searchable list where i just need to select an item I'm interested in.

If we were to say:
```
Are you looking to browse your Git branches locally in more efficient way? There is an easy way to make checking out branches a one-step process.
```
Then it feels to me like part of the issue goes away.

marcoow:
we can grow this into a blog post or use differently

emmasofiah2o:
@BobrImperator I'd say let's use this as a LinkedIn article. That way we don't need to have a TIL section on the website for these short blog posts and we can have a new "series" for LinkedIn to replace carousels :)

I'll work through Marco's suggestions later today and see if I can make the text clearer.

emmasofiah2o:
Posted on LinkedIn as article ✅

PR #2061, opened 2023-06-10
Fix text reveal animation for tall screens

lolmaus:
# Before

The animation was starting and finishing when the element reached certain distance to the TOP of the screen. As a result, the animation would not complete on a portrait screen:

censorship_before

# After

The animation is now tied to the element emerging from the BOTTOM of the screen, so it now works the same regardless of screen height:

censorship_after


davidhajba (approved):

SaeLLo (approved):

mansona (approved):
looks good to me 🎉

oscard0m (approved):

PR #2070, opened 2023-06-29
Add Skovik as Embroider initiative sponsor

marcoow:
Bildschirmfoto 2023-06-29 um 17 51 16


mansona (approved):
The code looks good, but I think we're hitting the caching issue that we hit last time 🤔 the background images don't seem to update on the preview

I'm approving because I assume you can flush the cache on the production build to fix this, but it's something that we should probably investigate as to why this doesn't update on the netlify preview

marcoow:
ah, right – redeployed and fixed now

PR #2078, opened 2023-07-15
 Add intro post Mario Gintili

shellandbull:
I have added a post to introduce myself to mainmatter.com 😄


marcoow:
I think this is more sth. for our Slack?

shellandbull:
@marcoow oh I had already posted something similar on slack but I'll close this as we don't seem to post this type of content

PR #2087, opened 2023-07-22
The Case for Rust on the Web

marcoow:
This is the blog post version of my talk at WeAreDevelopers World Congress: https://www.wearedevelopers.com/world-congress/program#sz-tab-45134

Bildschirmfoto 2023-08-03 um 14 18 09

## TODO

- [x] add `og:image` 
- [x] update title image


davidhajba (approved):

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+image: "/assets/images/posts/2023-06-09-securing-the-ecosystems-investment-in-emberjs/header-illustration.jpg"
+imageAlt: "The Ember logo on a gray backround picture"
+---
+Rust has received a lot of developer attention and love ever since it entered the stage around a decade ago. And it's not only the developers that love the language – the decision makers at the big corporates agree Rust is a great technology to build on and the language has seen a lot of adoption throughout the industry over the past few years. AWS uses Rust heavily for their platform, Google uses it for example in Android, Microsoft uses it in Windows, it's used in the Linux kernel now and Rust flies planes and satellites. Essentially, Rust is replacing C and C++ everywhere those were used before: systems programming, operating systems, all kinds of embedded systems, low level tooling, as well as in games and game engines.
```
LukeMathWalker: Added a few links to substantiate and toned down a bit the statements on replacing C/C++:

```suggestion
Rust has received a lot of developer attention and love ever since it entered the stage around a decade ago. And it's not only the developers that love the language – the decision makers at the big corporates agree Rust is a great technology to build on and the language has seen a lot of adoption throughout the industry over the past few years. AWS uses Rust heavily [for their platform](https://aws.amazon.com/blogs/opensource/sustainability-with-rust/), Google uses it [in Android](https://security.googleblog.com/2022/12/memory-safe-languages-in-android-13.html), Microsoft uses it [in Windows](https://www.theregister.com/2023/04/27/microsoft_windows_rust/). Essentially, Rust is on track to replace C and C++ in many of the domains those were used before: systems programming, operating systems, all kinds of embedded systems, low level tooling, as well as in games and game engines.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+Yet, we believe Rust has even more potential and has a great future ahead of it in other areas like the web and cloud as well. While those areas aren't necessarily areas where many teams used C or C++ before and Rust might not be an obvious choice, we believe it can in fact take development of web backends to the next level as it gives teams access to capabilities that were previously out of reach – what those are we'll see in more detail below. And while the web and cloud space is a young and evolving area for Rust still, there are quite a few companies that already use Rust in that area and quite successfully so like [Truelayer](http://truelayer.com), [Discord](http://discord.com), [Temporal](http://temporal.io), [Nando's](https://nandos.co.uk), [svix](http://svix.com), and many others.
```
LukeMathWalker:
```suggestion
Yet, we believe Rust has even more potential and a great future in other areas: the web and the cloud. Rust might not be an obvious choice there - those are not areas where many teams were previously using a low-level language like C or C++. Nonetheless, we believe that Rust can bring the development of web backends to the next level, giving teams access to capabilities that were previously out of reach – what those are we'll see in more detail below. We can already see quite a few companies using Rust successfully in the web and the cloud, even though Rust is still young and evolving in these areas: [Truelayer](http://truelayer.com), [Discord](http://discord.com), [Temporal](http://temporal.io), [Nando's](https://nandos.co.uk), [svix](http://svix.com), [Wingback](https://www.wingback.com/), and many others.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+## What?
+
+Around 8 years after its 1.0 release, Rust is still a relative young ecosystem. Yet, it's reached a level of maturity that makes it a viable choice to build real applications and Rust's web ecosystem is not exception there: Rust is clearly ready for the web as [arewewebyet.org](http://arewewebyet.org) confirms:
```
LukeMathWalker:
```suggestion
Around 8 years after its 1.0 release, Rust is still a relative young ecosystem. Yet, it's reached a level of maturity that makes it a viable choice to build real applications and Rust's web ecosystem is no exception there. Rust is clearly ready for the web as [arewewebyet.org](http://arewewebyet.org) confirms:
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+![arewewebyet.org](/assets/images/posts/2023-07-22-the-case-for-rust-on-the-web/arewewebyet.png)
+
+First of all, there's [tokio](http://tokio.rs), an asynchronous runtime that's a solid and performant foundation for networking applications. On top of that, there's mature and well-maintained web frameworks like [axum](http://github.com/tokio-rs/axum/), as well as middleware solutions like tower. There's mature drivers for all of the relevant datastores as well as ORMs and of course some datastores that were written in Rust itself. Finally, all other relevant aspects for building web apps are covered as well like (de-)serialization, i18n, templating, observability, etc. So overall, Rust provides solid and stable building blocks for building ambitious web backends.
```
LukeMathWalker:
```suggestion
First of all, there's [tokio](http://tokio.rs), an asynchronous runtime that's a solid and performant foundation for networking applications. On top of that, there's mature and well-maintained web frameworks like [axum](http://github.com/tokio-rs/axum/) and [actix-web](https://actix.rs). There's mature drivers for all of the relevant datastores as well as ORMs. Finally, you can find libraries covering all other relevant aspects for building web apps, such as (de-)serialization, internationalisation, templating, observability, etc. Overall, Rust provides solid and stable building blocks for building ambitious web backends.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Why?
+
+Of course one might ask: why should you care? What could motivate a team that already uses Ruby, Java, Elixir, TypeScript, Go, or whatever else, to adopt Rust? The answer to that question is straight forward really as there are essentially two main aspects that make Rust a great choice for building for the web: its efficiency and performance and the reliability and maintainability gains it comes with.
```
LukeMathWalker:
```suggestion
Of course one might ask: why should you care? What could motivate a team that already uses Ruby, Java, Elixir, TypeScript, Go, or whatever else, to adopt Rust? There are two key aspects that make Rust a great choice for building for the web: 

- its efficiency and performance
- the reliability and maintainability gains brought by its type system
```

LukeMathWalker: Moved the gains to a list to clearly delineate the "two", it felt a bit muddy with all those "and"s in a row.

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Efficiency and Performance
+
+Rust is famously efficient and performant. It will easily outperform languages like JavaScript, Ruby, Python, etc. that are commonly used for web applications, by several orders of magnitude. Also languages like Java or Go can't really reach the same levels of performance and efficiency that Rust offers. In addition, what separates Rust from all of these languages is its lack of a garbage collector. While garbage collected languages can be performant as well, they can never be performant consistently at all times. In regular intervals, execution will need to be paused so that the garbage collector can run and free unused memory, resulting in p95 performance numbers always being significantly worse than average performance. Rust not having a garbage collector means it can deliver consistent performance without those spikes.
```
LukeMathWalker:
```suggestion
Rust is famously efficient and performant. It will easily outperform languages like JavaScript, Ruby, Python, etc. that are commonly used for web applications, by several orders of magnitude. Other languages may have a higher performance ceiling (e.g. Java or C# or Go), but you need to invest serious engineering effort to come close to the level of performance offered by Rust's toolkit out of the box. In addition, Rust has a key advantage: it doesn't bundle a garbage collector. Garbage-collected languages can be plenty fast, but they are not **consistently** performant at all times. The garbage collector will introduce pauses in order to free unused memory, negatively impacting the tail latency (p95/p99) of your applications. Rust doesn't have this problem: it can deliver consistent performance without those spikes.
```

LukeMathWalker: Reworded things slightly to avoid the "XXX can be plenty fast as well if you write "absolutely unidiomatic code"!".

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+![Performance charts of garbage-collected languages and non-garbage-collected languages](/assets/images/posts/2023-07-22-the-case-for-rust-on-the-web/perf-charts.png)
+
+The only other languages that allowed for such stable and consistent performance on the same level as Rust does, were C and C++ before. The problem with those though is they are full of pitfalls and opportunities for shooting yourself in the foot, in particular by making mistakes doing manual memory management. As Linus Torvalds, who must know as he must have written millions of lines of code in both languages over the years, says:
```
LukeMathWalker:
```suggestion
C and C++ are the only other languages that allowed for such stable and consistent performance. Unfortunately, they are full of pitfalls and opportunities for shooting yourself in the foot, in particular by making mistakes doing manual memory management. As Linus Torvalds, the creator of Linux, says:
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+> "[…] it is so close to the hardware that you can do anything with it. It is dangerous. It's like juggling chainsaws. I also see that it does have a lot of pitfalls and they're easy to overlook. <author>Linus Torvalds</author>
+
+Because of these dangers of C and C++, both languages are really only usable when you're an expert in them or have an expert team already – which, in particular in the web field, nobody has as everyone mostly uses very different languages like JavaScript, Python, Ruby, Elixir, etc. That way, Rust not coming with the same pitfalls, empowers developers to build software at levels of efficiency that were previously out of reach for them. Rust will typically outperform other technologies commonly used to build web backends by several orders of magnitude while at the same time keeping a significantly lower memory footprint.
```
LukeMathWalker: I think we should make a point about security here as well: those "mistakes" can easily turn into vulnerabilities, which is a big no-no for software that is exposed to the internet.

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+Of course, being able to write e.g. web servers that can respond to requests in a fraction of the time compared to other technologies, also means being able to respond to the same number of requests with a fraction of the servers, which again means a fraction of the hosting cost. That difference has real financial consequences – as great as cloud hosting is, it's certainly not cheap. Being able to just switch off a substantial amount of cloud servers can easily be a reduction of hundreds of thousands of Euros per month even for small to mid-sized products and companies.
```
LukeMathWalker:
```suggestion
Of course, if your Rust web server can respond to requests in a fraction of the time compared to other technologies, it also means it can respond to the same number of requests with a fraction of the servers, which again means a fraction of the hosting cost. That difference has real financial consequences – as great as cloud hosting is, it's certainly not cheap. Being able to just switch off a substantial amount of cloud servers can easily be a reduction of hundreds of thousands of Euros per month even for small to mid-sized products and companies.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+Financial savings aren't even the whole story though. Using fewer servers also means using less energy. As good as running data centers on renewables is, the greenest energy still is the energy we don't use. And while it would be ridiculous to claim Rust is going to solve the climate crisis, it's important to acknowledge that the software we all write uses real resources and thus has an impact on the real world. Our industry has a tendency to forget about the resources our work consumes, although the [ICT sector accounts for 8-10% of the European CO2 emissions](https://ictfootprint.eu/en/about/ict-carbon-footprint/ict-carbon-footprint) – if we can use resource more efficiently and get the same output for less input, that's an aspect to consider when choosing a technology.
+
+### Reliability and Maintainability
+
+While Rust's performance and efficiency might be the most obvious reason to choose it for a project, it's not the only one. Other, maybe more relevant reasons in many cases, are the reliability and maintainability gains that Rust's strong type system unlocks.
```
LukeMathWalker:
```suggestion
While performance and efficiency might be the most obvious reasons to choose Rust for a project, they are not the only ones. Other, maybe more relevant reasons in many cases, are the reliability and maintainability gains that Rust's strong type system unlocks.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+Everyone has seen code snippets like the following one (which is Ruby on Rails) that's quite typical for a web application:
```
LukeMathWalker:
```suggestion
Snippets like this one are fairly typical for a web application (Ruby on Rails):
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+    …
+  end
+end
+
+…
+
+user.activate(Time.now)
+```
+
+While this code is quite expressive and writing code like this will lets you reach your goal fast, there's problems. In this (admittedly simple) example, while we can see the user's attributes, we don't know what rules there might be around those (e.g., if `active` is `true`, `activation_date` must be set as well probably? If `active` is `false`, presumably `activation_date` should be `nil`?). In order to validate these assumption, one has to look into the implementation of the `activate` method which means relatively high effort is required to get to the information.
```
LukeMathWalker:
```suggestion
While this code is quite expressive and writing code like this will let you reach your goal fast, there are problems. In this (admittedly simple) example, while we can see the user's attributes, we don't know what rules there might be around those (e.g., if `active` is `true`, `activation_date` must be set as well probably? If `active` is `false`, presumably `activation_date` should be `nil`?). In order to validate these assumption, one has to look into the implementation of the `activate` method which means relatively high effort is required to get to the information.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+Looking at the invocation of the `activate` method, we can't know whether it might raise an error or what timezone we're supposed to pass the time in. And while Ruby is a bit of an extreme example maybe as it's famously flexible, many of these problems can be found in other languages as well, for example in Java where there would also be no way to express the rules around the `active` and `activation_date` attributes of the user in the types directly, where things can be `null` and we can get `NullPointerException`s at runtime, or where we can't see at development time whether a method that's being called might throw an exception.
```
LukeMathWalker:
```suggestion
Looking at the invocation of the `activate` method, we can't know whether it might raise an error or what timezone we're supposed to pass the time in. And while Ruby might be a bit of an extreme example given its notorious flexibility, many of these problems can be found in other languages as well. Let's look at Java, as an example. We would still have no way to encode the rules around the `active` and `activation_date` attributes in the type system, nor when things can be `null` and we risk getting `NullPointerException`s at runtime.
```

Java technically does have checked exceptions, although as far as I understand they are not extensively used. So the last sentence might not be entirely accurate 🤔

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+While these problems might seem a bit made up and not too relevant in reality, they become bigger and more relevant over time as the codebase grows and the team working on the codebase grows as well or just changes as people leave and others join. There isn't a a small, tightly-knit team working on a limited size codebase anymore where everyone has a mental model of the entire application and is aware of all of the concepts and rules that are somehow present in the code but not encoded in any interfaces or data structures. Instead understanding those concepts requires people to read through classes and methods and tests. That task becomes more and more unrealistic over time, in particular for new joiners of which there tend to be plenty in long living codebases over time. That not only decreases efficiency but also likely results in increased error rates in production as people not familiar with the application make mistakes.
```
LukeMathWalker:
```suggestion
These issues become more and more relevant over time as the codebase grows and the team working on it grows as well, or just changes as some people leave and others join. It is no longer true that everyone working on the codebase has a perfect mental model of the entire application and all the implicit assumptions being made throughout the codebase. Instead understanding those concepts will require people to carefully read through classes and methods and tests. That task becomes more and more unrealistic over time, in particular for new joiners of which there tend to be plenty in long living codebases over time. That not only decreases efficiency but also likely results in increased error rates in production as people not familiar with the application make mistakes.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+
+    fn save(&self) -> Result<(), DBError> {
+        …
+    }
+}
+```
+
+First of all, for the user model, we can use Rust's `enum` with associated data. That way, it's completely clear what inactive and active users look like and what attributes can be expected to be set in what scenarios – in fact, active and inactive users do not even have the same attributes but each only has the ones that make sense for the respective user state they represent. Also the attribute's types are clearly defined – not only is Rust typed while Ruby is untyped obviously but the types are also very precise so that e.g. for the `activation_date` field, the expected Timezone is right in the type as well.
```
LukeMathWalker:
```suggestion
First of all, for the user model, we can use Rust's `enum` with associated data. That way, it's completely clear what inactive and active users look like and what attributes can be expected to be set in what scenarios – in fact, active and inactive users do not even have the same attributes but each only has the ones that make sense for the respective user state they represent. Also the attribute's types are clearly defined – not only is Rust typed while Ruby is untyped obviously but the types are also very precise so that e.g. for the `activation_date` field, the expected timezone is right in the type as well.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
pablobm: > We wouldn't also be able to check, just by looking at the signature, whether a method that's being called might throw an exception

If I'm understanding this correctly, I agree with @LukeMathWalker: Java does have the ability of adding the `throws` clause to a method signature, indicating that such exception may be thrown. Back when I did Java (early 2000s) I'm pretty sure this was required!

In fact, looking this up:

> In the Java class hierarchy, an exception is a checked exception if it inherits from java.lang.Throwable, but not from java.lang.RuntimeException or java.lang.Error. All the application or business logic exceptions should be checked exceptions. 

That's from https://en.wikibooks.org/wiki/Java_Programming/Checked_Exceptions

So there can still be other unexpected exceptions not declared in the signature, but they should (in theory) be rare.

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
marcoow: good point – added a short note on what exactly the risks are

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+The definition of the `activate` function also communicates much more detailed information explicitly that was all implicit in the Rails example. Again, the expected timezone for the `activation_date` is right in the type, and also the function returns [Rust's `Result`](https://kapeli.com/dash_share?docset_file=Rust&docset_name=Rust&path=doc.rust-lang.org/1.68.2/std/result/enum.Result.html&platform=rust&repo=Main&source=doc.rust-lang.org/1.68.2/std/result/enum.Result.html&version=1.68.2) which clearly communicates that errors might occur when calling it. In fact, the Rust compiler will require both the success and error variant of the result to be handled so that no unhandled runtime exceptions can occur.
```
LukeMathWalker:
```suggestion
The signature of the `activate` function also encodes, explicitly, a lot of the information that was implicit in the Rails example. Again, the expected timezone for the `activation_date` is right in the type, and also the function returns [Rust's `Result`](https://kapeli.com/dash_share?docset_file=Rust&docset_name=Rust&path=doc.rust-lang.org/1.68.2/std/result/enum.Result.html&platform=rust&repo=Main&source=doc.rust-lang.org/1.68.2/std/result/enum.Result.html&version=1.68.2) which clearly communicates that errors might occur when calling it. In fact, the Rust compiler will require both the success and error variant of the result to be handled so that no unhandled runtime exceptions can occur. 
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+While this is obviously a rather simple example, it illustrates the two main advantages of Rust well:
+
+* Many of the concepts and rules that were implicit in the Rails example are explicitly communicated through types in the Rust code. Active and inactive users can be clearly distinguished, and for the date fields even the expected timezone is encoded in the types. That expressiveness makes the code much more approachable, in particular to newcomers to the codebase, thus resulting in improved maintainability.
+* Rust substantially improves reliability as well since whole classes of errors that are common in other languages (including typed languages such as Java or Go) will be detected at compile time instead of runtime. The compiler guarantees the `activation_date` argument of the `activate` function to have a value and any error that's potentially returned by the function to be handled.
+
+Overall, the improvements in reliability and maintainability that Rust comes with, are often overlooked while everyone is focussed on Rust's performance. However, in many cases these benefits might be more relevant for a project's long term success than sheer performance numbers.
```
LukeMathWalker:
```suggestion
Overall, the improvements in reliability and maintainability that Rust comes with are often overlooked while everyone is focussed on Rust's performance. However, in many cases these benefits might be more relevant for a project's long term success than sheer performance numbers.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+## How & Where?
+
+As Rust's main advantages are reliability, maintainability, efficiency, and performance, use cases for the language are obviously ones where these four aspects are particularly relevant. While one could say those aspects are always relevant and generally be correct saying that, the benefits come at a cost that needs to be factored into a decision for Rust.
```
LukeMathWalker:
```suggestion
As Rust's main advantages are reliability, maintainability, efficiency, and performance, use cases for the language are obviously ones where these four aspects are particularly relevant. While one could say those aspects are always relevant and generally be correct saying that, the benefits come at a cost that needs to be accounted for.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+Overall, Rust still requires a higher upfront investment than other technologies, in particular compared to other technologies commonly used in web projects:
```
LukeMathWalker:
```suggestion
Overall, Rust still requires a higher upfront investment than other technologies, in particular compared to those commonly used in web projects:
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+![Chart showing the initial effort for building on Rust compared to other languages](/assets/images/posts/2023-07-22-the-case-for-rust-on-the-web/initial-effort.png)
+
+While languages like JavaScript and Ruby are made for getting results fast, Rust deliberately doesn't leave as much freedom and requires a program to pass all of the compiler's checks before one gets to a working result. That requires higher initial effort for building an application with Rust compared to those languages. Additionally, there's a few mountains people need to climb before they're productive with Rust at all – mainly that is mastering Rust's infamously hard to learn ownership system.
```
LukeMathWalker:
```suggestion
While languages like JavaScript and Ruby are made for getting results fast, Rust deliberately doesn't leave as much freedom and requires a program to pass all of the compiler's checks before one gets to a working result. That requires higher initial effort for building an application with Rust compared to those languages. Additionally, there's a few mountains people need to climb before they're productive with Rust at all – mastering Rust's unique ownership system.
```

Following Google's article, I think we can avoid phrasing it as exceptionally difficult, but rather calling this out as a unique characteristic.

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+However, when looking beyond the initial stages of a project and expanding the view on a longer time horizon where aspects like maintainability, reliability, and stability become exceedingly important, the picture changes as the effort for working on Rust projects goes down over time:
+
+![Chart showing the effort required for building and maintaining Rust applications over time](/assets/images/posts/2023-07-22-the-case-for-rust-on-the-web/total-effort.png)
+
+Rust apps being more reliable and thus requiring less time invested into bug fixing, as well as being more maintainable and thus easier to work on efficiently with growing and changing teams, results in effort going down over time. That tends to be the other way round for other languages where effort goes up over time as the impact of reliability and maintainability challenges becomes bigger and more costly as the team grows. In a way, the additional investment made in the beginning when using Rust, pays dividends over time.
```
LukeMathWalker:
```suggestion
Rust apps being more reliable and thus requiring less time invested into bug fixing, as well as being more maintainable and thus easier to work on efficiently with growing and changing teams, results in effort going down over time. That tends to be the other way round for other languages where effort goes up over time as the impact of reliability and maintainability challenges becomes bigger and more costly as the team grows. The additional investment made in the beginning when using Rust pays dividends over time.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Use Cases & Adoption Paths
+
+Base on Rust's strengths and effort curve, whenever evaluating whether or not to choose Rust for a particular case, the main questions to answer are:
```
LukeMathWalker:
```suggestion
Based on Rust's strengths and effort curve, whenever evaluating whether or not to choose Rust for a particular case, the main questions to answer are:
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+* What are the requirements regarding reliability?
+* What are the plans for long term maintenance?
+* What scale are is the system being built for and thus how much money could Rust potentially save in terms of hosting?
```
LukeMathWalker:
```suggestion
* What scale is the system being built for and thus how much money could Rust potentially save in terms of hosting?
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+* Based on the answers to the above questions, is the additional initial investment worth it?
```
LukeMathWalker: I would add an extra question:

* How much Rust expertise do we have in the team?

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+While in some cases the conclusion will be the additional initial investment is not worth it, for some cases the assessment will be clearly in Rust's favor. Some typical use cases we're seeing include:
```
LukeMathWalker:
```suggestion
While in some cases the conclusion will be that the additional initial investment is not worth it, for some cases the assessment will be clearly in Rust's favor. Some typical use cases we're seeing include:
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+* For core business system that implement the core logic of a service, aspects like reliability and long-term maintainability are more relevant than for other parts of the system maybe (like e.g. a public HTTP API that might sit on top of the core business system where the flexibility of something like Rails might be valued higher and reliability requirements might not be as strict).
```
LukeMathWalker:
```suggestion
* For core business systems that implement the key business logic for a product, aspects like reliability and long-term maintainability are primary concerns.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+* For financial systems, there is generally little tolerance for bugs and the improved stability that Rust leads to can be a deciding factor. Plus, in many cases being able to make trades faster means making more money so Rust's performance advantage has direct financial value.
```
LukeMathWalker:
```suggestion
* For financial systems, there is generally little tolerance for bugs and the improved stability that Rust leads to can be a deciding factor. Plus, performance is a key requirement with a clear financial impact in specific scenarios (e.g. trading systems).
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
marcoow: I did not actually know that :)
I'd say let's just leave that sentence out then?

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
LukeMathWalker: Financial systems is a lot broader than trading, so reworded things slightly.

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
LukeMathWalker: I don't think the example quite fits, so I removed it in the suggestion, but we could use a different one.

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+* Any system that must be able to deliver high throughput and performance will clearly benefit from Rust. Systems like proxy servers that sit in front of a number of microservices must have minimal overhead and consistent performance. Garbage collected languages with their unreliable performance characteristics are typically not an option in these cases.
+* Finally, for any system that is run at large scale, there are big saving potentials in terms of hosting costs.
+
+Once a decision for Rust has been made, there's two main adoption paths – either an entire application is (re-)built from scratch or the decision is for an incremental adoption path.
```
LukeMathWalker:
```suggestion
Once a decision for Rust has been made, there's two main adoption paths – either an entire application is (re-)written from scratch in Rust or you are looking at incremental adoption alongside other technologies.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+One thing to keep in mind when starting a project from scratch, when the project domain might not be fully understood and the team might not have a clear idea of the final architecture they are going for, is that Rust isn't necessarily your friend for prototyping and explorative coding. While technologies like Ruby on Rails are specifically built for that kind of work but then show their shortcomings over time, Rust is the other way round as mentioned before. The same mechanisms that lead to improved reliability and maintainability make it harder to experiment since Rust requires the entire codebase to be consistent at all times. Because of that, it's often easier to replace an existing system with a Rust implementation instead of starting from a blank slate.
```
LukeMathWalker:
```suggestion
One thing to keep in mind when starting a project from scratch: Rust isn't necessarily your friend for prototyping and explorative coding, the kind taking place when the project domain might not be fully understood and the team might not have a clear idea of the final architecture they are going for. While technologies like Ruby on Rails are specifically built for that kind of work but then show their shortcomings over time, Rust is the other way round, as mentioned before. The same mechanisms that lead to improved reliability and maintainability make it harder to experiment since Rust requires the entire codebase to be consistent at all times. Because of that, it may be easier to replace an existing system with a Rust implementation instead of starting from a blank slate.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
LukeMathWalker: Yeah, let's avoid it. Or we can use C# as an example, that doesn't have checked exceptions.

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+#### Incremental Adoption
+
+While every engineer likes a greenfield project, incremental adoption paths are typically easier to manage. Instead of rewriting entire applications in Rust, it's usually much more straight forward to extract individual aspects of existing systems into Rust (micro) services. For these kinds of projects, the lack of full-featured macro frameworks also is less critical as much of the aspects that such frameworks would cover might not be relevant for these projects at all due to their limited scope.
```
LukeMathWalker:
```suggestion
While every engineer likes a greenfield project, incremental adoption paths are typically easier to manage. Instead of rewriting entire applications in Rust, it's usually much more straightforward to extract individual aspects of existing systems into Rust (micro) services. For these kinds of projects, the lack of full-featured macro frameworks is also less critical, as much of the aspects that such frameworks would cover might not be relevant for these projects at all due to their limited scope.
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+An even easier adoption path than building (micro) services is writing native extensions in Rust. Teams that have existing Node/Ruby/Elixir/Python/etc. applications that might suffer from performance or stability issues, might choose to rewrite parts of those applications in Rust and directly integrate the Rust code with their existing applications as native extensions. While that used to require writing C or C++ code in most languages and thus wasn't a feasible option for most teams, Rust opens up that possibility. Like (micro) services, native extensions are a great first step on an incremental adoption path while keeping risk at a minimum.
```
LukeMathWalker: We should link to our article on writing Rust native functions for Elixir here.

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
marcoow: let's just remove the sentence – I updated your suggestion above (one good thing about the crates.io drama: I learned that that's possible for the PR author :) )

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+In parallel to existing frameworks and libraries like actix and axum continue to improve, new frameworks will be built at higher abstraction levels as well. As more teams adopt Rust for web backends, the ecosystem will explore and evaluate approaches, learn from evaluating those in different use cases, and 
+collectively settle on a set of patterns and architectures. Eventually, we'll see full-featured macro-frameworks like a "Rust on Rails" that will significantly simplify building web applications in Rust.
+
+And the future in which Rust is easier to approach and adopting it doesn't come at an additional initial cost, isn't so far away even. Google, who have been adopting Rust significantly for years, recently shared they don't really see a productivity penalty for Rust compared to any other language they use:
```
LukeMathWalker: Let's add a link to the article here.

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,196 @@
...
+
+> "Overall, we’ve seen no data to indicate that there is any productivity penalty for Rust relative to any other language these developers previously used at Google. <author>Lars Bergstrom, Google</author>
+
+While it might be questionable how that experience translates to other teams and companies, it is a strong positive signal for sure.
+
+We at Mainmatter are highly optimistic that Rust will take off in the web and cloud space in the coming months and years and for that reason have [made a strategic bet on Rust](/blog/2022/10/12/making-a-strategic-bet-on-rust/). We believe that Rust is the first step towards a new era of web development where developers can levarage a technology that allows them to reach higher and previously unthinkable levels of efficiency, stability, reliability, and maintainability without giving up on developer experience and productivity.
+
+If you're looking to adopt Rust but aren't comfortable going on the journey alone, [reach out](/contact/). We can help with [mentoring, team augmentation, and custom development](/rust-consulting/).
```
LukeMathWalker:
```suggestion
If you're looking to adopt Rust but aren't comfortable going on the journey alone, [reach out](/contact/). We can help with [mentoring, training, team augmentation, and custom development](/rust-consulting/).
```

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
LukeMathWalker: 😂

LukeMathWalker:
Sorry for the cascade of comments 😅 
Many are style-related, so feel free to take them or leave them.

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
marcoow: I think it's a good point, in particular as I can say then many teams have expertise because so many devs love Rust and use it in the free time

LukeMathWalker: Yup, plenty of hobby users that are looking for a chance to use it at work

marcoow:
@LukeMathWalker I think those were all good – I had a hard time getting into a proper flow when writing this so there was definitely lots of room for improvement regarding the wording 🙌

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
marcoow: makes sense – added

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
marcoow: added

marcoow:
@davidhajba can you prepare an `og:image` and something for the title (something with the Rust logo and some kind of schematic internet illustration like a globe or whatever with servers/lines would work I think)

davidhajba:
> @davidhajba can you prepare an `og:image` and something for the title (something with the Rust logo and some kind of schematic internet illustration like a globe or whatever with servers/lines would work I think)

Sure - [HERE](https://www.notion.so/mainmatter/832aecb85d3747a2adab6a5fce452f5e?v=999496eec6c24b06818ad82a363a8d45&p=e24168f02e584991a4838e5e84f3f11b&pm=s) is the ticket with the designs

src/posts/2023-07-22-the-case-for-rust-on-the-web.md
```diff
@@ -0,0 +1,459 @@
...
+description: "TODO"
+og:
+  image: /assets/images/posts/2023-06-09-securing-the-ecosystems-investment-in-emberjs/og-image.jpg
+tagline:
+  '<p>Mainmatter <a href="/blog/2022/10/12/making-a-strategic-bet-on-rust/">made
+  a strategic bet on Rust for the web</a> in the last year. Our goal is to guide
+  teams during the process of adopting Rust in web environments and we''re on
+  our way towards reaching that goal: we started the EuroRust conference, we
+  joined the Rust Foundation, we hired Rust expert and author <a
+  href=\"http://twitter.com/algo_luca/\">Luca Palmieri</a> as our Principal
```
LukeMathWalker:
```suggestion
  href="http://twitter.com/algo_luca/">Luca Palmieri</a> as our Principal
```

The link is broken otherwise

marcoow: I think that was Prettier that broke it actually when it converted the outer `"` to `'`…

PR #2100, opened 2023-08-11
Add CTA before content in all posts tagged with Rust

LukeMathWalker:
My e11y is _extremely_ basic, so I'm sure there's a better way to do this 😅 
The key idea: we have a consistent CTA in all posts tagged as Rust, with associated tracking in Plausible to measure conversion. 

~If we like the idea, I can then go through all existing Rusts posts to make sure that we don't have redundant CTAs.~ Done ✅

### DIFF
--- a/src/assets/css/app.scss
+++ b/src/assets/css/app.scss
@@ -65,6 +65,7 @@
 @import "components/startups";
 @import "components/scroll-slides";
 @import "components/quotes-list";
+@import "components/blog-cta";
 
 @import "animations/text-animation";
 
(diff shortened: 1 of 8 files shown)


marcoow (commented):
I think the implementation is good – just not sure we should use a blockquote. I think some special paragraph in bold or so might be better. Maybe David can make something real quick? I should also be able to add the CSS then next Monday easily.

LukeMathWalker:
I'll ask David to see if he has ideas 👍🏻 

The blockquote is not _perfect_ but I think it's workable as a start while we find a more appropriate UI element.

marcoow:
sure but it should really be easy to get sth. done that's better until Monday or so so no significant delay even

marcoow (approved):

PR #2122, opened 2023-09-21
Twios: add proper cross compiled binary, always fetch latest ref

BobrImperator:
- Adds new properly cross-compiled binary
- Adds a new 'twios' label and allows to run `create-twios` job when added/removed.
this allows for testing the workflow.
- Removes unnecessary volta setup, runs the binary directly instead.
- Makes sure the latest branch changes are fetched when creating and editing post.

closes https://github.com/mainmatter/mainmatter.com/pull/2129

### DIFF
--- a/.github/workflows/twios.yaml
+++ b/.github/workflows/twios.yaml
@@ -20,7 +20,7 @@ permissions:
 jobs:
   create_twios:
     runs-on: ubuntu-latest
-    if: ${{!github.event.pull_request.body && github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' || github.event.label.name == 'twios'}}
+    if: ${{!github.event.pull_request.body && github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' || github.event.label.name == 'create-twios'}}
     steps:
       - uses: actions/checkout@v3
       - name: Fetch $TWIOS_BRANCH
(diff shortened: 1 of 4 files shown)


marcoow:
@BobrImperator doesn't seem to work? https://github.com/mainmatter/mainmatter.com/actions/runs/6265427251/job/17014303132?pr=2122

BobrImperator:
@marcoow It does work, it just attempts to read this Pull request's comment which doesn't have the info.

Here's an actual PR it would work for
https://github.com/mainmatter/mainmatter.com/pull/2125

marcoow:
ok but since the build for this PR fails, I'm not sure what I'm supposed to review – assume we don't want to merge with failing build?

BobrImperator:
@marcoow  The jobs are setup to run only on `twios` branches, unfortunately i made the mistake of naming this one with `twios` word in it.

BobrImperator:
@marcoow see the jobs ran for a renamed PR with the same changes https://github.com/mainmatter/mainmatter.com/pull/2129

"create_twios" and "edit_twios" jobs fail here only because they're not meant to be ran here.
In a real scenario "create_twios" would only run on automated, scheduled trigger or workflow_dispatch while the "create-twios" solution is only for testing new program versions.

Also the `edit_twios` job works correctly here: https://github.com/mainmatter/mainmatter.com/pull/2125

marcoow (approved):

PR #2157, opened 2023-10-19
Add og-images to case studies

### LINKED ISSUE #1926: Add `og:image`s to case studies
Since the case study pages are specifically relevant and we might be linking to them directly, we should have `og:image`s for them.

yaigrie:
closes #1926


yaigrie:
@marcoow Is adding the og: image: ... bit everything I need to add to make the og images work?

marcoow (commented):
I think these work and yes, adding `og:image:` in the metadata is the only think you need to do. We should use larger files though maybe? These seem to be much smaller than what we're typically using and relatively low quality?

marcoow (commented):
I only realize now the "y" in "Study" is cut off for all images

marcoow (approved):

PR #2175, opened 2023-10-30
The Case against Project Managers - blog post

emmasofiah2o:
- Add text
- Add og-imga
- Add visual

We are using the same background image for all biz dev posts - does it make sense to change it now or should we keep it that way?


src/posts/2023-11-03-the-case-against-project-managers.md
```diff
@@ -0,0 +1,92 @@
+---
+title: "The Case against Project Managers"
+authorHandle: sarahlorenzu
+tags: [startups]
+bio: "Sarah Lorenz"
+description:
+  "Sarah Lorenz showcases the reasons that make the classic project manager role
+  obsolete for teams."
```
marcoow:
```suggestion
  "Sarah Lorenz explains the reasons not to have a classic project manager role
  and introduces Mainmatter's approach to self-managed teams."
```

marcoow (commented):
let's not use the same background image everywhere

src/posts/2023-11-03-the-case-against-project-managers.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Improved Productivity and Morale
+
+We have seen this evolution of project leadership leading to improved
+productivity in many different projects. At the same time, it leads to improved
+morale and a less confrontational work environment and overall atmosphere in
+project teams.
+
+If you’re interested in establishing the “iteration lead” approach in your own
+team, check out our Playbook or reach out and we’ll be happy to talk.
```
SaeLLo: We should add the link that leads to the playbook:  https://mainmatter.com/playbook/

SaeLLo (approved):

src/workshops/telemetry-for-rust-apis.md
SaeLLo, line 1: This should not be here, right?

emmasofiah2o: There was a problem with linting, I rebased but it didn't solve it apparently. I'll ask someone to check

BobrImperator: This is fine. As Emma mentioned, someone pushed broken changes and it was failing our CI everywhere.

It's just an automated fix that wasn't somehow propagated through the `master` branch, it should be fine to keep it here.

emmasofiah2o: Bart said we can leave it as is

marcoow (approved):

PR #2190, opened 2023-11-08
change base layout to display pagination

### LINKED ISSUE #2186: Title on blog index page(s) is wrong
The title on the blog index page(s) is wrong – it seems like we're trying to show the correct page title but the statement doesn't actually get evaluated and renders as plain text instead:

Bildschirmfoto 2023-11-08 um 14 57 02

yaigrie:
closes #2186

### DIFF
--- a/src/author.njk
+++ b/src/author.njk
@@ -1,7 +1,7 @@
 ---
 layout: base
 eleventyComputed:
-  title: "All posts by {{ paged.author.data.name }}"
+  title: "All posts by {{ paged.author.data.name }} {% if paged.number > 1 %}- Page {{ paged.number }}{% endif %}"
 description: "Frontend and JavaScript, Svelte and Ember, Rust and WASM, Elixir and Phoenix, Product and Design: Stay on top of what's happening."
 pagination:
   data: collections.authorsPostsPaged
--- a/src/author.njk
+++ b/src/author.njk
@@ -1,7 +1,7 @@
 ---
 layout: base
 eleventyComputed:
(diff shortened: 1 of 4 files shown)


yaigrie:
@marcoow I found now way to stop the title if function being rendered as plain text in the blog file, so I put the condition in the base layout instead

marcoow (commented):
Nice! I recognized there's a mistake on the tags page (which might have existed before): https://deploy-preview-2190--objective-northcutt-37494c.netlify.app/blog/tag/rust/ (says "Page 9")

yaigrie:
On the live site the tags don't seem to have the pagination visible in the title
https://mainmatter.com/blog/tag/rust/page/3/#posts

if we want a quick fix for the blog title I added a new condition to the base template so it continues to be only visible for the main blog page, but I'll look further into why the tags got the pagination wrong

marcoow:
Hm, it's strange that in production we don't see the problem. With the extra property we'd also have to make sure we actually cover all the pages like the author pages which are paginated as well, e.g.: https://mainmatter.com/blog/author/tobiasbieniek/page/2/#posts

Maybe we can live without the extra prop for now.

yaigrie:
@marcoow I managed to put the if condition back into the blog file itself, by aligning the template information at the top to the ones of the tags and author file (so mainly removing the 'js', '{}' and '""'), so now it's able to interpret the condition and doesn't write it as plain text
Also I treated the tags and author the same way, so they would have the page number in their title, too

marcoow (approved):
🙌

PR #2221, opened 2023-11-15
Fix text reveal animation for tall screens

lolmaus:
This is a copy of the PR https://github.com/mainmatter/mainmatter.com/pull/2061 . It was closed by accident and could not be reopened.

***

# Before

The animation was starting and finishing when the element reached certain distance to the TOP of the screen. As a result, the animation would not complete on a portrait screen:

censorship_before

# After

The animation is now tied to the element emerging from the BOTTOM of the screen, so it now works the same regardless of screen height:

censorship_after


mansona (approved):
Looks good to me 👍

marcoow:
@yaigrie can you take this over and merge if it's fine?

PR #2240, opened 2023-11-24
Add :where() blog post

beerinho:
This PR adds a new blog post about using `data-selectors` and the `:where()` pseudo-function to help style your base components to make them easier to override and easier to access.

Three things I'd especially like feedback on:
- do you think the component images fit in? Would it be better to have them displayed in a different way?
- does it matter which image format I used? At the moment the images are PNG but I noticed they get transformed anyway,
- the "get in touch" section at the bottom feels repetitive as we already have 2 CTAs about Ember on the page


src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
```diff
@@ -0,0 +1,266 @@
...
+which would extend our base `Button` and give it some styles to reflect the
+aesthetic of our app. Finally, we could have a `Card` component that uses this
+`PrimaryButton`, but we actually want it to have slightly less padding and
+change the `border-radius`.
+
+So in this scenario, it wouldn’t be unlikely to have the padding for a single
+element set at three different levels, meaning that we would need to be very
+careful about the selectors we are using to make sure we are setting the styles
+without doing anything that might be impossible to override if we wanted to
+extend our `Card` component later on: I’m looking at you `!important`.
```
marcoow, line 51: I don't understand the problem tbh. The `Card` component would set the padding for a class it defines with in its own scope and sets on the `Button` component?

```
<style>
.button {
  padding: 5px;
}
</style>

<template>
  <!-- this is the Card component's template -->
  <div><Button class="button" /></div>
</template>
```

I'm not sure how we need to be careful which selectors we use as the `Card` component can only use the `.button` selector anyway as it cannot even know what the class names in the `<Button>` component are?

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
```diff
@@ -0,0 +1,266 @@
...
+
+`.button[type="button"]` gives as a specificity of (0,2,0) making it is slightly
+more difficult to override elsewhere. This is intentional as we shouldn’t really
+be overriding these as they are base styles for the button.
+
+![Button CSS with specificity of (0,2,0)](/assets/images/posts/2023-11-24-the-ultimate-way-to-style-base-components/button-css-specificity-0,2,0.png)
+
+> For anyone that didn’t know, if you open up the Inspector in Safari or Chrome
+> and hover over the CSS selector, it will tell you the specificity.
+> Unfortunately I couldn’t see how to do this on Firefox.
```
marcoow, line 136: I understand you want this to stand out but it's not a quote actually :)

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
```diff
@@ -0,0 +1,266 @@
...
+```html
+{% raw %} // primary-button.hbs
+<button
+  ...attributes
+  local-class="primary-button"
+  @onClick={{@onClick}}
+  @loading={{@loading}}
+>
+  {{yield}}
+</button>
```
marcoow, line 169:
```suggestion
<Button
  ...attributes
  local-class="primary-button"
  @onClick={{@onClick}}
  @loading={{@loading}}
>
  {{yield}}
</Button>
```

I guess? Otherwise this is the native button?

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
```diff
@@ -0,0 +1,266 @@
...
+}
+```
+
+`.card` simply adds styles to the component as you would expect,
+
+`.card [data-button]` allows us to access and style all of the `Button`
+components directly - that is both the `PrimaryButton` and `Button` because they
+are both using the same underlying element. This means that we can easily
+override the styles of the base component without needing to drill down through
+the whole component tree.
```
marcoow, line 250: But this means that the `Card` component is now implicitly coupled to an implementation detail in the `Button` component (the fact that it sets the `data-button` attribute on the `button` it renders). I'd argue this is an anti-pattern and if you want to change the buttons, you should set a class?

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
```diff
@@ -0,0 +1,266 @@
...
+
+`.button` is showcasing the difference between the specificity of the styles
+within the `PrimaryButton` component, as the `padding` is getting overridden but
+the `background-color` remains unchanged.
+
+As an app begins to grow, and multiple components start to be reused in
+different scenarios, minor changes can start to have a big impact across the app
```
marcoow, line 257: minor changes like not setting `data-button` anymore in the `Button` component :)

marcoow (commented):
The post reads well but I don't really get what problem it solves and am not sure whether this doesn't actually introduce a new problem anyway…

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
marcoow: If you wanted to set the padding for all `Button`s at any nesting level within the `Card` component, isn't that what CSS variables are for?

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
beerinho: I guess I need to find a better way to explain this, essentially the issue is that if the `Button` component has 2/3 selectors then the specificity will be higher and therefore we would need to be careful with the selectors in the `Card` component to make sure the base styles are being overridden

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
beerinho: Yeah I wasn't sure what to use here, do you have a suggestion for what it could be instead?

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
beerinho: Good spot, thank you

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
beerinho: I had the issue of trying to make this example as simple as possible so that the focus could be the thing that I wanted to showcase but I think it is too simple in this case. The main issue that I am failing to point out is that you might not have a way to add a class to that component directly; if it is 3+ components deep then you wouldn't want to have to pass a class all the way through those components. So in this scenario I can see why this doesn't make sense and I need to think of a better way to showcase it.

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
beerinho: In this scenario the `data-button` attribute would be one of the foundational blocks suggested, so it would have big consequences to remove it for sure. But you could put it on par with the class: the button component would definitely not look as intended without it

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
marcoow: You can make it bold or sth. - I don't think we have an element for that (you could ask Nora of course but there's some other things she needs to do with higher prio once she's back)

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
marcoow: But still, isn't setting the style for a component 3 levels down the tree based on a `[data-]` selector an anti-pattern since you're specifically using that technique to bypass CSS encapsulation? Isn't this what CSS variables are for so that the component can be explicit about what style properties are supposed to be overridden (and thus part of its public API)?

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
beerinho: I'll just make it bold for now then

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
marcoow: Sure, the button would look wrong without the `data-button` attribute but that's not the question (you can imagine a refactoring where the attribute is renamed and the component's own styles are changed accordingly). The issue here is that the `data-button` attribute becomes part of the component's public API and all of the component's own/internal styles become part of it's API as well in a way since you're allowing all of the styles to be changed via selectors based on the `data-button` attribute – whatever you change, you'll need to make sure it's consistent with other styles (e.g. aspect ratios or whatever; if you're exposing a CSS variable you can allow that to be changed and then define styles that depend on the value of the variable, thus enforcing consistency like width is always twice the height or whatever).

In short, what I'm saying is I have the feeling that this post recommends a pattern, I'd consider an anti-pattern actually – of course I might still be misunderstanding sth.

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
beerinho: We definitely could use a CSS variable here, and I think that's another flaw with this example, because overriding padding is easily done in that way, but if we wanted to override multiple things or if we needed to drill down deeper to get to the property then it wouldn't work as well. 

I'll add in the example with css variables and then explain the benefits and drawbacks

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
```diff
@@ -0,0 +1,266 @@
...
+```html
+{% raw %}
+// button.hbs
+<button
+	...attributes
+	data-button
+	data-loading={{@loading}}
+	local-class="button"
+	type="button"
+	{{on "click" this.onClick}}
```
BlueCutOfficial, line 79: I would change the indent with 2 spaces here.

marcoow:
@beerinho what's the status of this?

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
```diff
@@ -0,0 +1,266 @@
...
+  extended throughout your app."
+og:
+  image:
+tagline: |
+  <p>Styling your app can get difficult as the number of components grows and you begin to reuse components in different scenarios. This blog post goes through the ultimate way to style your base components to allow easy extensibility throughout your Ember app.</p>
+image:
+imageAlt:
+---
+
+Foreword: Should out to
```
beerinho, line 18:
```suggestion
Foreword: Shout out to
```

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
```diff
@@ -0,0 +1,266 @@
...
+[Florian Pichler](https://mainmatter.com/blog/author/pichfl) for introducing me
+to this pattern earlier this year.
```
beerinho, line 20:
```suggestion
to this pattern recently.
```

beerinho:
> @beerinho what's the status of this?

I haven't had a chance to revisit this in a meaningful way. And since the feedback you gave - the only feedback - seems to point towards it being an anti-pattern, I'm wondering if it should be a post at all. The other option is to reframe it and either remove the usage of `data-` attributes or make it clearer that the `data-` attribute can be used as an escape hatch rather than a useful way of styling; or perhaps using global classes instead of `data-` attributes would be better for this, but that still comes with the issue of it being part of the component's public API.

This post also makes extensive use of `ember-css-modules` which isn't `embroider` friendly, which already puts a shelf-life on it.

src/posts/2023-11-24-the-ultimate-way-to-style-base-components.md
beerinho: I think you touch on an interesting point about what should be considered the component's public API. I'm thinking I've just framed it wrong, perhaps this isn't a suggestion for how to build the foundation of your app, but rather a way to expose the ability to update styles in a mature app without having to refactor everything

marcoow:
> And since the feedback you gave - the only feedback - seems to point towards it being an anti-pattern

I mean I could totally be wrong about this of course :) – happy to exchange thoughts and come to a conclusion

marcoow:
closing this – @beerinho feel free to reopen if it is still planned to go out at some point

PR #2259, opened 2023-12-07
Shorter CTA for Rust

LukeMathWalker:
It feels like our Rust CTA is too verbose at the moment.
I reworded it to reduce the amount of screen real estate dedicated to the CTA: 

image

I've also changed the "contact us" to go straight to the Calendly link: lower friction than having to fill a form and we're not getting too many inbounds at the moment. If it gets spammy, we can adapt.


marcoow (approved):
👍

PR #2277, opened 2023-12-19
Experiment with the Rust workshop page

LukeMathWalker:
Before: https://mainmatter.com/services/workshops/rust/
After: https://deploy-preview-2277--objective-northcutt-37494c.netlify.app/services/workshops/rust/


LukeMathWalker:
I think this serves our purpose a bit better, and it packs more information than the existing version. 
Let me know what you think @marcoow @yaigrie

yaigrie:
I think not using the slides is a good idea, also generating some trust with the quotes.
I also think it's a great idea to promote the public workshops, I think that supports the credibility as well. 
But without the colourful slides we may lose some engagement, I will think about a solution in the overall design update :)

LukeMathWalker:
>  But without the colourful slides we may lose some engagement, I will think about a solution in the overall design update :)

Do we have a timeline for the redesign shipping?

yaigrie:
@LukeMathWalker The plan was getting it ready before christmas, but as I'm still in the design process and not starting to develop it I'm afraid it has to be delayed :(
But I'm fine with merging this one first and then changing the design afterwards

LukeMathWalker:
That'd be ideal for me, so that I have something relatively self-explanatory to share with potential prospects while we get the redesign over the line.

yaigrie (approved):

PR #2279, opened 2023-12-21
Ruby/RoR landing page

yaigrie:
```[tasklist]
### Tasks
- [x] change URL to include rails
- [x] replace client logos with only relevant ones
- [x] add this page in the footer
- [x] finalize texts
- [x] update `og:image` to say "We're veteran Ruby users and Ruby on Rails experts"
```


yaigrie:
This page has no content and isn't ready yet :)

marcoow:
Great – I'll look at it later or tomorrow!

marcoow:
@yaigrie let's use both the Ruby (the diamond) and the Ruby on Rails (the one that's on the page now) logo please

marcoow:
@yaigrie can you update the `og:image`?

src/ruby-rails-consulting.njk
```diff
@@ -0,0 +1,119 @@
+---
+layout: base
+title: We're veteran Ruby and Ruby on Rails Consultants
```
pablobm: Consultants or Experts? (Compare with the h1 text). Also: should the last word be capitalised when "veteran" is not? (Or vice-versa).

oliverbarnes (approved):
Do you intend to add Ruby is Not Dead, and the legacy stuff? Otherwise, it looks good to me 👍 

Posting my slack comment here, for reference:

> Ruby is not dead" is a slogan that pops up a lot these days, that I think resonates with ruby teams and companies. A resistance thing
> There's also lot of demand for help with big old legacy codebases. You already touch on that, with "upgrade legacy code and leave technical debt behind", "identify problems" etc. But perhaps elaborating could be good -  acknowledging teams and codebases that have gone through a lot of churn, where previous developers are gone and a lot of institutional knowledge has been lost. It's a particular situation where making sense of many layers of code by waves of developers with different styles is needed. An archeology of sorts. CTOs going crazy with that situation might be specially attracted by our specific experience with it (and we have loads).
> There's also the legacy scenario where they go through big migrations, extracting services, big refactors, phase outs, that sort of thing. They (CTOs) know they need senior devs for them, but senior devs with specific experience with big bang changes might be even more attractive

marcoow:
I think saying Ruby is not dead, what comes across is actually the opposite…

src/ruby-rails-consulting.njk
marcoow: I think we want to say both experts and consultants in different places on the site

marcoow: fixed capitalization

oliverbarnes:
> I think saying Ruby is not dead, what comes across is actually the opposite…

true, it'd need to be spun in a way where it didn't sound negative. perhaps not worth it then

marcoow:
@yaigrie sorry, can you change the `og:image` to say _"We're veteran Ruby users and Ruby on Rails
consultants"_?

PR #2287, opened 2024-01-08
testing redirect rules

yaigrie:


yaigrie:
Hey @marcoow 
I checked the Google index errors. I don't think there is any real error here, as the newest error, that the email was about was just duplicate links for the playbook, coming from the alternate URLs from the workshop. 

But looking into that topic lead me to finding that the redirect for those pretty URLs doesn't work? 
https://svelte-sveltekit-workshop.mainmatter.com/ should result in https://mainmatter.com/svelte-sveltekit-workshop/ imo, but it doesn't work for me. The other redirects like /services/digital-products to services/digital-products-and-design/ work fine. 

But that leads to https://svelte-sveltekit-workshop.mainmatter.com/ working for visitors of the page and results in links not working correctly (like the homepage link on the Mainmatter logo, which just stays on the same site or the links in the footer resulting in https://svelte-sveltekit-workshop.mainmatter.com/blog or https://svelte-sveltekit-workshop.mainmatter.com/services etc

So I replaced the "/" link for the logo with the explicit mainmatter.com link, as I couldn't find a more elegant solution, and tried to adjust the redirect rules. But I'm not sure if they'll work that way and how to test it, so I need some help :)

My thought process was something like if there's something added after https://svelte-sveltekit-workshop.mainmatter.com/ like blog (*), then it will redirect to mainmatter.com/something. 
I hope this isn't too confusing 

Also the replacement workshop for the web based rust workshop doesn't exist, should we replace it with another or is it okay that people following the old workshop link don't find a workshop anymore? 
https://mainmatter.com/services/workshops/web-based-services-in-rust/

marcoow:
> But looking into that topic lead me to finding that the redirect for those pretty URLs doesn't work?
https://svelte-sveltekit-workshop.mainmatter.com/ should result in https://mainmatter.com/svelte-sveltekit-workshop/ imo, but it doesn't work for me. The other redirects like /services/digital-products to services/digital-products-and-design/ work fine.

No, those should not redirect – we want the page to show under the subdomain. I think the problem is that we're not marking one as the canonical URL for the content (or maybe we're marking both?) – the subdomain should be the canonical URL for these pages I think.

yaigrie:
@marcoow Okay, then I misunderstood that! 
Currently the original URL (https://mainmatter.com/svelte-sveltekit-workshop/) is the only canonical URL
All those sites are displayed in the google search console as "Alternative page with correct canonical tag", which is correct I think? So it's not an error that we should fix? 

"All those sites" are a lot, because every site exists like three times due to the subdomains (like https://rust-telemetry-workshop.mainmatter.com/blog/author/levelbossmike/)

marcoow:
So the workshop sites for which we have landing pages (currently the only live one is https://svelte-sveltekit-workshop.mainmatter.com) should be the canonical URLs for that content. I'd say all links going out of those pages should point to mainmatter.com again – only the page itself should be on the subdomain.

marcoow:
I think we can also switch off all the other subdomains for the workshops we did in the past btw (e.g. Luca's telemetry workshop)

yaigrie:
@marcoow This topic is new to me, so please excuse my struggles :) 
I added the canonical link snippet to the remaining svelte workshop, I hope this is sufficient.
Also I need your help with switching off the subdomains (I assume you need netlify access?)

I kept the * -> splat snippet for the svelte workshop as I hope this will enable the outgoing links to redirect to mainmatter.com

marcoow:
No problem – this is a bit of a complex topic for sure. 

> I kept the * -> splat snippet for the svelte workshop as I hope this will enable the outgoing links to redirect to mainmatter.com

No, we'll actually have to use links starting with `https:/mainmatter.com/` in the footer etc.

marcoow:
I'm wondering whether we should redirect the other subdomains to the respective workshop pages under mainmatter.com/services/workshops – what do you think? In that case it might be good to have a banner or so though, saying the workshop ended

yaigrie:
@marcoow I think having a banner on the regular workshop pages could be confusing, as people who don't come from the redirect might assume they can't book this anymore? Tho I guess you can put a visibility rule on that section depending if the user is coming from such an old subdomain link, but that's a new topic for me too :D 
Maybe we could keep the individual workshop pages and say on those "Hey, this one workshop has ended but you can book it for your team (link to respective regular workshop) or look if you like our other workshop (link to workshop page)"

marcoow:
Yes, that idea makes sense 👍

PR #2301, opened 2024-01-18
Add rail europe featured case study

yaigrie:


yaigrie:
- [ ] add screenshots of the rail europe platform

src/cases/rail-europe.md
```diff
@@ -9,5 +11,80 @@ description:
...
+  image: /assets/images/cases/cs-rail-europe-og-image.jpeg
 ---
+
+{% from "quote.njk" import quote %}
+{% from "image-aspect-ratio.njk" import imageAspectRatio %}
+
+<div class="case-study__section">
+  <h2 class="case-study__heading h5">About Rail Europe</h2>
+  <div class="case-study__text">
+    <h3 class="h4">Rail Europe was founded in 1931 and is looking back at an impressive history of 90+ years in the travel economy. Over the years, the company established itself as the go-to brand for European train bookings and technology solutions for over 15,000 travel professionals in 70 countries.</h3>
```
SaeLLo: Wouldn't be "Industry" the right word (instead of "Economy")?

SaeLLo:
I was wondering if we should also add the tag "development"?

SaeLLo:
@marcoow Can you also review so we can merge this?

marcoow (commented):
Let's feature this at the top of the Work page and shift all the others down? I'd also put Jürgen's image next to the quote and place the quote somewhere inside of the text maybe vs. at the bottom under the image? We should also add Rail Europe to the "Work" list in the footer (and maybe remove Timify and Expedition from there) and we could feature the case study on the Svelte page as well. Finally, we **could** replace one of the featured case studies on the home page with the RE case (I'd suggest replacing mvb).

yaigrie:
@marcoow Are there two other case studies we could feature on the svelte page, so we can have three?
right now, there are only blog posts on the technology pages, can we find three case studies for each one?

marcoow:
We only have this one for Svelte for now. We need to have design options though that allow us to work with that as well. For the Rust and Elixir pages I think we also only have one each, for Ember it's no problem to find 3.

marcoow (changes requested):
Just one more thing I just recognized: now the small description of our work for RE is right below the teaser for this new case study. Let's switch the positions of the RE and Sage small descriptions on the /work page

src/index.njk
```diff
@@ -26,7 +26,7 @@ title: null
 {{ featuredServices(services['services-list'], content) }}
 {% set logoCompanies = ['Qonto', 'Trainline', 'Aleph Alpha', 'Timify', 'Experteer', 'Robin Hood', 'Rail Europe' ] %}
 {{ logoList(logoCompanies) }}
-{% set clients = ['qonto', 'experteer', 'trainline', 'mvb'] %}
+{% set clients = ['qonto', 'experteer', 'trainline', 'rail-europe'] %}
```
marcoow, line 29: @yaigrie let's replace Experteer with Aleph Alpha (in a separate PR)

marcoow (approved):

PR #2302, opened 2024-01-18
Pin version of sharp to fix installation on Linux

lolmaus:
On Linux, `pnpm i` fails with:

> sharp: Installation error: Request timed out

This is a [known problem](https://github.com/lovell/sharp/issues/3795). According to an [official recommendation](https://github.com/lovell/sharp/issues/3795#issuecomment-1849524010), it is supposed to be fixed by upgrading to `sharp` 0.33 (we are on 0.32).

`sharp` is a dependency of `@11ty/eleventy-img`.

Tested build and dev locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,5 +77,10 @@
   },
   "engines": {
     "node": ">=16"
+  },
+  "pnpm": {
+    "overrides": {
+      "sharp": "^0.33.2"
+    }
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,6 +4,9 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
+overrides:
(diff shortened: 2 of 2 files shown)


marcoow:
If it's a problem on Linux, why does it work on CI which runs Linux? I don't really know anything about this though – guess someone else would be better able to review this.

oliverbarnes:
> If it's a problem on Linux, why does it work on CI which runs Linux?

I guess it could vary according to the platform Linux is running on, for example.  I'm not familiar with sharp at all and I haven't dealt with npm build issues in a while (thank god), but skimming through the issue and the PR for the version update on the lib itself, this looks fine to me. Looks like it fixes networking problems when pulling in dependencies?

I'll defer to the ember/svelte folks, though

marcoow:
this seems to work fine

PR #2305, opened 2024-01-23
Embroider update 2

lolmaus:
Depends on #2302.


lolmaus:
So far, I have transcribed the entire Chris's report from the Ember Initiative video call. It requires heavy editing, probably can be shrunk down 2—4 times.

I've also added details on the part that I've been doing.

PR #2339, opened 2024-03-08
Fix Lighthouse CI

marcoow:
We haven't had the checks applied to PRs for quite a while which is a shame as the entire job isn't really useful without and it's very easy to miss regressions…

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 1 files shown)


BobrImperator (approved):
Isn't the current branch head a default for the checkout action?

> We haven't had the checks applied to PRs for quite a while which is a shame as the entire job isn't really useful without and it's very easy to miss regressions

It seems like older PRs did run it but didn't appear in checks for some reason.

https://github.com/mainmatter/mainmatter.com/actions/runs/8143837390/job/22256566918

image
image

It's weird :sweat_smile:

marcoow:
No by default you're in detached head state. That's why the check didn't show before – LH CI added it but it was added for a Sha that doesn't even exist in the repo history which is why it didn't show on the PR that shows checks for the latest revision in the PR's branch.

PR #2356, opened 2024-04-02
Fix animated GIFs

zeppelin:
Anim GIFs are currently broken on the entire site, such as inside @BobrImperator 's [post](https://mainmatter.com/blog/2021/12/08/validations-in-ember-apps/). This might not be the actual fix, because AFAIK WebP images can be animated, too, but for some reason I could't make it work. Maybe someone with more experience with Eleventy configs can?

Another issue: even if I managed to output anim GIFs _locally_, it refused to work when deployed to Netlify. Guessing it's a caching issue? Nevertheless, this is something to be aware of, so at the very least the solution should be found on a local setup first, before wasting time with Netlify (which I did).

cc @marcoow

### DIFF
--- a/utils/transforms/contentParser.js
+++ b/utils/transforms/contentParser.js
@@ -135,7 +135,7 @@ module.exports = function (value, outputPath) {
 
         let formats = ["webp", imageData.fileType];
         if (imageData.fileType === "gif") {
-          formats = ["webp"];
+          formats = ["gif"];
         }
 
         let url = "./static" + imageData.src;


zeppelin:
Looks like the GIFs in the deployment won't animate :/

BobrImperator:
We could also opt to telling eleventy, or markdown parser (whateve'r handling the attachments) to use a video element instead of an img [like here ](https://github.com/mainmatter/mainmatter.com/pull/2331/files#diff-eda4a1532cb0ecdc50048a9f7e93c6452a4826279a1616b95edd6a9d8ffff4e3R448).

marcoow:
I thought WebP supports animations?

zeppelin:
Pretty sure it does, but I couldn't find the right incantation for Eleventy. I can dig further, but at first I asked if anyone knows this from the top of their head

marcoow (approved):

PR #2378, opened 2024-04-25
Update nav.njk

### LINKED ISSUE #2374: Current nav item not marked in top menu
I realized the current item isn't marked in the top menu anymore. E.g. when I'm on https://mainmatter.com/work/, the top menu looks like this:

Bildschirmfoto 2024-04-23 um 14 04 52

but "Work" should actually be marked in some way as I'm on that page (and it should be marked active for any of the sub-pages like case studies as well) – same for blog and the services.

yaigrie:
closes #2374 

(this was just missing spaces between the class names that get added in the if condition)

### DIFF
--- a/src/components/global/nav.njk
+++ b/src/components/global/nav.njk
@@ -28,7 +28,7 @@
         <details class="nav-submenu" data-has-submenu>
           <summary class="nav-submenu__toggle nav__link" data-menu-toggle>
             <span
-              class="nav__link-text{% if page.url and '/services/' in page.url %}nav__link-text--active{% endif %}"
+              class="nav__link-text {% if page.url and '/services/' in page.url %} nav__link-text--active {% endif %}"
               >Services</span
             >
             <span class="nav__link-arrow">{% include 'svg/arrow.njk' %}</span>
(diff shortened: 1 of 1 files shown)


src/components/global/nav.njk
```diff
@@ -28,7 +28,7 @@
         <details class="nav-submenu" data-has-submenu>
           <summary class="nav-submenu__toggle nav__link" data-menu-toggle>
             <span
-              class="nav__link-text{% if page.url and '/services/' in page.url %}nav__link-text--active{% endif %}"
+              class="nav__link-text {% if page.url and '/services/' in page.url %} nav__link-text--active {% endif %}"
```
marcoow:
```suggestion
              class="nav__link-text{% if page.url and '/services/' in page.url %} nav__link-text--active {% endif %}"
```

src/components/global/nav.njk
```diff
@@ -107,7 +107,7 @@
             {% set parentChildCompare = entry.url %}
           {% endif %}
           <a
-            class="nav__link nav__link-text{% if entry.url == page.url or (page.url and parentChildCompare in page.url) %}nav__link-text--active{% endif %}"
+            class="nav__link nav__link-text {% if entry.url == page.url or (page.url and parentChildCompare in page.url) %} nav__link-text--active{% endif %}"
```
marcoow:
```suggestion
            class="nav__link nav__link-text{% if entry.url == page.url or (page.url and parentChildCompare in page.url) %} nav__link-text--active{% endif %}"
```

marcoow (approved):
Need to make 2 changes to fix linting

PR #2396, opened 2024-05-10
add newsletter cta

yaigrie:
This PR adds Newsletter CTA sections at the bottom of the Rust, Svelte and Ember Consulting Pages and to the Homepage. 

@marcoow Please feel free to change the text I wrote :)

### DIFF
--- a/src/assets/css/base/_buttons-links.scss
+++ b/src/assets/css/base/_buttons-links.scss
@@ -11,6 +11,7 @@ p a {
   color: var(--color-white);
   text-decoration: none;
   font-weight: 700;
+  border: none;
   border-radius: 0.125rem;
   transition: color ease 0.5s;
   overflow: hidden;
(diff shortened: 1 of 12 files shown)


marcoow (commented):
I think this is fine overall but let's include the input field for email right in the cta block on the website so users can fill it out there already

yaigrie:
@marcoow How do I send the email address further to the mailchimp form? Can you help me with that?

marcoow:
@yaigrie it's the same as what we're doing for emberfest.eu essentially – a form that posts directly to mailchimp. You can just put an empty `<form>` tag in for now – I can fill in the rest.

yaigrie:
@marcoow That would be great, thank you :)

marcoow (changes requested):
I guess the email field should be a little wider, at least on desktop? Also, let's reuse the same styles from the contact form with the placeholder that moves into a label when editing etc.

marcoow:
@yaigrie I fixed the forms but styling needs to be fixed

marcoow (approved):

PR #2403, opened 2024-05-10
replace 62 breakpoint with 64

### LINKED ISSUE #2389: Use consistent breakpoints
I recognized we seem to use similar, yet not identical breakpoints for the menu and header section (see how the menu and header do not change at the same time):

https://github.com/mainmatter/mainmatter.com/assets/1510/d67dc2de-5f17-480f-96ed-a6466417a90d

I think it would feel better if we used the same breakpoint (and might be easier to maintain as well)? I don't know whether there are other similar cases or this is the only – might be good to review all breakpoints.

yaigrie:
closes #2389

### DIFF
--- a/src/assets/css/components/_author.scss
+++ b/src/assets/css/components/_author.scss
@@ -78,7 +78,7 @@
   color: var(--color-aqua);
 }
 
-@media (min-width: 62em) {
+@media (min-width: 64em) {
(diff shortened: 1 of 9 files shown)


marcoow (commented):
Should we use variables for the breakpoints?

PR #2408, opened 2024-05-14
move cta from post to layout

yaigrie:
Luca noticed that there's something broken with Chris' git blog post:

Screenshot_2024-05-14-17-25-51-24_e4424258c8b8649f6e67d283a50a2cbc

This seems to occur when a global cta is placed inside of the blog content file, so I added a condition for git and design to have those two topic specific ctas in the post layout instead of the post itself

### DIFF
--- a/src/posts/2021-06-02-how-to-create-an-interface-inventory.md
+++ b/src/posts/2021-06-02-how-to-create-an-interface-inventory.md
@@ -226,4 +226,3 @@ and audit on your own, you are never alone. Consider joining the
 place to ask questions or get feedback regarding this process), or
 [hire a facilitator](/services/workshops/design-system-kickoff-interface-inventory/)
 to help make this project a success.
-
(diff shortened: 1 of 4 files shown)


src/posts/2021-06-02-how-to-create-an-interface-inventory.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: "How to create an interface inventory"
 authorHandle: msmarhigh
-tags: process
+tags: process design
```
marcoow:
```suggestion
tags: [process, design]
```

marcoow: otherwise this is used as one string

marcoow (changes requested):

marcoow (approved):

PR #2410, opened 2024-05-15
add KISTERS mini-case

marcoow:
…to Rust page and work page


marcoow:
holding off from merging this until we can release the BMW mini case as well

PR #2438, opened 2024-06-06
unregister all service workers that might be there from the past

### LINKED ISSUE #2437: Remove service worker
Apparently we have a service worker, probably a leftover from the past. We should add code that deletes that as we don't need it and it can cause problems in the worst case.

marcoow:
closes #2437

### DIFF
--- a/src/components/layouts/base.njk
+++ b/src/components/layouts/base.njk
@@ -56,6 +56,12 @@ og:
     </div>
     <script>
       document.querySelector("body").setAttribute("data-js-enabled", "true");
+
+      navigator.serviceWorker.getRegistrations().then(registrations => {
+        for (const registration of registrations) {
+          registration.unregister();
+        }
+      });
     </script>
   </body>
 </html>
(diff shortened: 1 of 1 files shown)


src/components/layouts/base.njk
```diff
@@ -56,6 +56,12 @@ og:
     </div>
     <script>
       document.querySelector("body").setAttribute("data-js-enabled", "true");
+
+      navigator.serviceWorker.getRegistrations().then(registrations => {
+        for (const registration of registrations) {
+          registration.unregister();
+        }
+      });
```
BobrImperator: `navigator.serviceWorker` might be undefined on incognito/private mode.
as per [docs](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/serviceWorker)

```suggestion
     if (navigator.serviceWorker) {
         navigator.serviceWorker.getRegistrations().then(registrations => {
           for (const registration of registrations) {
            registration.unregister();
           }
       });  
     }
```

marcoow: fixed

BobrImperator (approved):

PR #2445, opened 2024-06-13
Turn on incremental builds

nchan0154:


marcoow:
sent you an invite as a collaborator which (I think) should make required env vars available for GitHub actions jobs (the Gravity and Percy jobs fail due to the env vars not being available). If that doesn't work, please work on branches in this repo instead of a fork which should definitely make the env vars available.

PR #2457, opened 2024-06-18
Improve richtext styles

### LINKED ISSUE #2455: Fix formatting in blog posts
- [ ] `<code>` tags aren't styled
- [ ] italic text isn't styled
- [ ] bold text isn't styled
- [ ] lists aren't styles
- [ ] footnotes don't work
- [ ] strikethrough doesn't work

nchan0154:
Closes #2455

- Code and pre tags inherit from the prism styles, of course, if you want syntax highlighting, you need to pass in a language, but this will now catch all the languageless tags
- Inherited other styles based on other areas of the site
- Installed the markdown-it-footnotes plugin for footnotes support

Preview:
image


marcoow (commented):
This seems to break some things – see the Percy check (I'll send you an invite in a minute)

nchan0154:
@marcoow Fixed by adding markdown-it options. Do you know if there is a way to run percy locally in a way that you can see the images or does it only trigger with a PR?

marcoow:
I don't think you can run it locally that way. Normally it works well to just check a few pages manually locally and then have Percy raise any things you missed locally (which usually are a few only if any)

marcoow (commented):
The list style doesn't seem to work in Safari:

Bildschirmfoto 2024-06-18 um 10 56 14

Also it seems there's a bunch of additional `<br>`s that might get added by the Markdown parser?

Bildschirmfoto 2024-06-18 um 10 56 40

nchan0154:
Switched to using `::before` instead of `::marker` for better support. Percy looks a little odd but this solution appears to be more cross-browser compatible. Also removed the markdown-it option for adding a `<br>` tag for new lines.

marcoow (commented):
🙌

PR #2460, opened 2024-06-19
Standardize icon animations

### LINKED ISSUE #2386: Fix broken arrow animations
Some of the arrow animations aren't working as expected:

* the one for the "Services" item in the main menu is only animated forwards but reverts instantly:

https://github.com/mainmatter/mainmatter.com/assets/1510/f8f9ab66-edc5-41a0-a65c-f9c7fc745261

* the one for the playbook download button on the main landing page is a bit wobbly (at least in Safari):

https://github.com/mainmatter/mainmatter.com/assets/1510/5bb71c79-dac4-47ce-a1bc-76024fc7869f

Can you check all of the animations across the entire site? I assume we should also be using the same animation everywhere anyway so that either all of them work or none?

nchan0154:
closes #2386

Pulled out all the icon arrows to use CSS custom properties and draw form the same animation source so future maintenance is easier. 

I tried to reproduce the issue Marco experienced in #2386 with several different Safari devices and couldn't seem to replicate the wobblyness, I suspect it has to do with pixel rounding, judging by the way the arrow seems to move down half a pixel, transform, then move up half a pixel. I've applied a 3d transform and the css `will-change` property which has [fixed animation issues](https://www.nicchan.me/blog/a-use-case-for-will-change/) in the past. Fingers crossed that it is an improvement!

### DIFF
--- a/src/assets/css/app.scss
+++ b/src/assets/css/app.scss
@@ -9,6 +9,7 @@
 @import "base/forms";
 @import "base/prism-theme";
 
+@import "components/icon-arrow";
 @import "components/nav";
 @import "components/nav-submenu";
 @import "components/mobile-menu";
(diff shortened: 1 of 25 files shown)


marcoow (approved):
Yep, this looks much cleaner now!

PR #2471, opened 2024-06-26
Add Chris's talk from EmberConf 2024

emmasofiah2o:
- Add talk to appearances
- Add channel
- Add new thumbnail design made by Julia

Screenshot 2024-07-02 at 11 14 39


marcoow:
let's find a different thumbnail design – I don't think this works well with the rest our branding/design

emmasofiah2o:
> let's find a different thumbnail design – I don't think this works well with the rest our branding/design

Do you have any more details I should forward to Julia?

marcoow:
let's discuss in Sketch

PR #2476, opened 2024-07-04
add BMW Car IT mini case

marcoow:
…on the Rust page:

Bildschirmfoto 2024-07-04 um 12 51 39

as well as on the Work overview page:

 
Bildschirmfoto 2024-07-04 um 12 52 16


src/cases/bmw-car-it.md
```diff
@@ -0,0 +1,19 @@
...
+  We ran a 4-day mixed on-site and remote workshop for them, covering all the
+  topics they needed to know.
+tags: Workshops
+description:
+  <p>BMW Car IT develops software for infotainment and autonomous
+  driving.</p><p>When they wanted to get a team of developers up to speed with
+  Rust, they reached out to Mainmatter. We ran a workshop for them, giving them
+  a smooth entry to Rust.</p>
+hero:
+  tags: "Rust / Training"
```
SaeLLo: training without capital letter

src/cases/bmw-car-it.md
```diff
@@ -0,0 +1,19 @@
+---
+layout: case-study
+company: BMW Car IT
+problem:
+  BMW Car IT was looking to get a group of developers up to speed with Rust
```
SaeLLo: I'd suggest this instead: "BMW Car IT was looking for support to familiarize a group of developers with Rust quickly."

SaeLLo (changes requested):

src/cases/bmw-car-it.md
marcoow: isn't that just a more complicated way to say the same?

SaeLLo: I think it sounds more professional. At least I would delete the "fast".

marcoow: ok, removed fast – I think things sounding more boring is often confused with sounding "professional" and we don't really want that kind of professional jargon, see the brand styleguide as well

PR #2492, opened 2024-07-15
turn off proseWrap

mansona:
Previously we have had prosewrap turned on to make sure there is consistency in the line width for markdown files. This changes the prosewrap to have a value of `never` to bring the consistency in the opposite way. 

Reasons for this change:

- wrapping behaviour is now customisable in your own editor
- the cognitive load of a diff is **significantly** reduced when working on blog posts e.g. when rephrasing things sometimes it can change every line in a paragraph because all the lines get shuffled around
- improve the workflow of accepting changes on github without the need to run `lint:fix` locally afterwards

Personal reasons for this change: 

In most cases, my dyslexia doesn't affect my work; this is mainly because I can configure the tools I use to work most effectively for me. When line breaks are forced in markdown documents I can't edit my local settings to correct this, thus forcing any processes to be significantly harder for me. That plus the extra cognitive load of [diffs like this](https://github.com/mainmatter/mainmatter.com/pull/2489/commits/58aa4b630094aed7c37487462db5f80f59719bf2#diff-b099d1234ee0b29df4754ec0c7c836e877f7de8cf2cc6b73e42ef7a6f05f0820L175-L184) make blog writing/reviewing a thoroughly unpleasant experience for me.

I understand that a PR that changes every file in the repo is not easiest thing to consider but I would like us not to dismiss it outright because it's difficult. I don't mind a PR like this being rejected as long as it's seriously considered and we **actively** decide to reject it 👍

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -10,8 +10,8 @@ module.exports = {
     {
       files: "*.md",
       options: {
+        proseWrap: "never",
         printWidth: 80,
-        proseWrap: "always",
       },
     },
     {
--- a/README.md
+++ b/README.md
@@ -4,6 +4,4 @@ The source code for [https://mainmatter.com](https://mainmatter.com).
 
 ## Copyright
 
-Copyright &copy; 2019-2022 Mainmatter GmbH (https://mainmatter.com), released
-under the
-[Creative Commons Attribution-NonCommercial 4.0 International license](https://creativecommons.org/licenses/by-nc/4.0/).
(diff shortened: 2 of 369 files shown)


BlueCutOfficial (approved):
I am in favor of this change because it happened to me several times to have the linter failing after accepting a suggestion directly on GitHub. Overall I would say the current config doesn't help me in any way but occasionally gets me lose some time.

oliverbarnes (approved):
Seems reasonable to me, I do find the blog posts' diffs easier to read like this as well.

And changing past blog posts will keep consistency in the site design overall.

nickschot (approved):

mansona:
In case anyone is wondering why I'm continuously force pushing: as this changes every Markdown in the website I need to rebase it every time anything is merged. I'm rebasing any time I notice anyone has pushed because I'm eager to keep this in a mergeable state 👍

PR #2489, opened 2024-07-09
Blog post/ last embroider update

BlueCutOfficial:
Preview link: https://deploy-preview-2489--objective-northcutt-37494c.netlify.app/blog/2024/07/16/embroider-update/

PR #2493, opened 2024-07-15
Apply changes directly

mansona:
Instead of working on the #2489 PR via comments I'm proposing that we merge this PR essentially addressing all of the comments and once we have merged this PR we can merge #2489 as our release 👍

### DIFF
--- a/src/posts/2024-07-16-embroider-update.md
+++ b/src/posts/2024-07-16-embroider-update.md
@@ -3,7 +3,9 @@ title: "Embroider Initiative: Conclusion"
 authorHandle: academierenards
 tags: [ember, embroider]
 bio: "Marine Dunstetter, Senior Software Engineer"
-description: "A summary of what Mainmatter achieved with the Embroider Initiative and what are the next steps for the community."
+description:
+  "A summary of what Mainmatter achieved with the Embroider Initiative and what
+  are the next steps for the community."
 og:
   image: /assets/images/posts/2024-05-29-embroider-update/og-image.jpg
 tagline: |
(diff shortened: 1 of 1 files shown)


src/posts/2024-07-16-embroider-update.md
```diff
@@ -189,42 +187,46 @@ modern Ember apps, the `babel.config.cjs` file will be responsible for the Babel
...
 
-## 4. Handover
-
-The Embroider Initiative now comes to a - successful - end; We have achieved so
-many important pieces and we are proud to see the first Ember apps in the
-community making their way to Vite. It's now up to the Ember community to pursue
-the implementation of Embroider: achieve the topics that remain open, challenge
-the blueprint to make Ember as intuitive as possible, support the classic
-features that are still wanted, enhance the documentation, make sure all the
-most popular addons work with Vite... To ease the handover, the public GitHub
```
BlueCutOfficial, line 215: I listed a few examples of the remaining tasks to highlight the fact there's still a lot to do. I am not under the impression it reduces the value of what we achieved, and to me, it was rather a way to introduce why we are still needed on the project. wdyt?

src/posts/2024-07-16-embroider-update.md
```diff
@@ -20,45 +20,43 @@ imageAlt:
...
 Embroider. The
 [lastest blog post from May](https://mainmatter.com/blog/2024/05/29/embroider-update/)
 is a good summary of the different steps of the implementation. In even shorter
 words:
 
-- Stage 1 (achieved in the 1st half of the initiative): To achieve backward
-  compatibility, each classic Ember Addon package is rebuilt into a new v2
-  package format stored in a `rewritten-packages` folder, more static and
-  analyzable.
+- To achieve backward compatibility, each classic Ember Addon package is rebuilt
```
BlueCutOfficial, line 39: What's the value of removing the references to the stages? To me, it helps visualize the global roadmap and the timeline a little better.

BlueCutOfficial:
@mansona I am not opposed to the PR approach to earn time, but considering it removes the possibility of selecting what changes to apply and it kind of "imposes" the whole thing, I think this approach would benefit from a few auto-review comments to explain why you think these changes are valuable (at least the not-that-obvious ones).

src/posts/2024-07-16-embroider-update.md
mansona: I agree that it's useful to consider the embroider stages to be aware of what it is doing under the hood, but every time I tried to read this section with the context of each of the sages (and even thinking which "half" of the initiative they were completed in) just made it hard to reason about it. In reality the 3 stages aren't as distinct in Vite any more, there is ember prebuild then there is vite build 🤷 and some of the time references weren't even correct, we didn't actually do the stage 1 work as part of the initiative because it was done before we came along

src/posts/2024-07-16-embroider-update.md
mansona: what i did with this section was I merged your work with Marco's suggestions in the comments to try and come up with something that kept the original feeling

PR #2494, opened 2024-07-16
Add new travel industry landing page

nchan0154:
Also switches one component to use margins instead of padding so that the margin collapses, but happy to tweak.

## TODO

* [x] finalize text
* [x] link whitepaper (below "…and adapt flexibly to a changing landscape.")
* [x] add preview image of whitepaper to CTA
* [x] update preview image of whitepaper with final screenshot
* [x] put whitepaper PDF in Google Drive (_"Mainmatter – Travel Transformation Whitepaper.pdf"_)
* [x] reduce file size of travel industry brochure


marcoow (commented):
Thanks, looks good! I'll take over and finalize the content

marcoow:
@SaeLLo can you have a look already? I'd say this is ready for now (although missing the correct download link)

SaeLLo:
looks good to me.

marcoow:
@SaeLLo can you test this please at https://deploy-preview-2494--objective-northcutt-37494c.netlify.app/travel/ ? This should all work now, including the mail delivery, Pipedrive integration, etc.

SaeLLo:
For the whitepaper download: Shall we add the field "Your biggest challenge you are facing"?

marcoow:
Hm, it would have to be optional for sure and I wouldn't be sure how many people would actually fill sth. in than. We could make it a select box but not sure what the options would be?

SaeLLo:
Yes, maybe this is too complicated. The idea was to get insights that we can use later on.

umbrellarocket:
@marcoow I added a commit with the whitepaper preview to this branch, pls check it out.

marcoow:
@umbrellarocket can we make the image much larger? Also, it looks a bit strange on mobile tbh with the image being placed above the form but seemingly sized more or less randomly?

umbrellarocket:
@marcoow Thanks for the feedback! I made the image larger; placed it differently on wider screens; and left-justified it on mobile devices (so its placement is in line with the rest of the elements)

.gitignore
```diff
@@ -26,3 +26,4 @@ doc/
 _site
 .DS_Store
 .sass-cache
+package.json
```
BobrImperator: This should be reverted

PR #2545, opened 2024-10-18
twios(update): use v0.8.0

BobrImperator:
Updates TWIOS to use 0.8.1 version.

This should help with avoiding Github rate limits that are making it impossible to generate twios file.

https://github.com/mainmatter/this-week-in-open-source/commits/v0.8.1


mansona (approved):
I'm approving this even though it's literally impossible to review because we're just using a binary file 🙈 🫠 

should we consider deploying twios as an npm package so that we can at least have some knowledge of which version we're consuming?

PR #2551, opened 2024-10-22
Improve workshops booking flow

marcoow:
This makes it easier for people to reach out to us for workshops by putting a contact form right on the page:

Bildschirmfoto 2024-10-22 um 17 58 09

### DIFF
--- a/src/assets/css/components/_cta-banner.scss
+++ b/src/assets/css/components/_cta-banner.scss
@@ -1,5 +1,4 @@
 .cta-banner {
-  position: relative;
   z-index: 50;
 
   &--purple {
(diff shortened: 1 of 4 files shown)


umbrellarocket:
One suggested change here @marcoow if it's easily doable: can we make the text field for  "Tell us about your training needs" start out one line tall and expand if users type in more? This change would improve the UX on mobile. Reasoning being that the "Get in touch button" might be too far below the bounds of a mobile users's screen.

image

marcoow:
Hm, I see what you mean but it might be non-trivial. Can you create an issue for that so that we do that later as an improvement?

IdeliaRisella (approved):

PR #2610, opened 2025-01-24
Workshop fixes

marcoow:
* Remove the mention of our focus being backend from the Rust workshops page intro. While that's true, we don't want to send people away who are looking for a Rust workshop for their C team that's working on embedded topics or so (think BMW).
* Filter public workshops by start date, not end date – once the workshop has started and it's too late to join, it should disappear.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -126,12 +126,10 @@ module.exports = function (eleventyConfig) {
     return Boolean(collection?.find(({ page }) => page.url === url));
   });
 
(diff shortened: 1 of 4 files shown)


umbrellarocket:
Do we also want to display upcoming workshops first? For example, this order would be flipped if we ordered by start date

image

marcoow:
@umbrellarocket good point – added

umbrellarocket (approved):

LukeMathWalker (approved):

PR #2611, opened 2025-01-24
feat: new blog post `Runes and Global state: do's and don'ts`

paoloricciuti:
Initial draft of the blog post in title...it's not finished but i wanted to get it out there to get initial feedback...is missing the part where i solve the gotchas about global state in an isomorphic application (which basically means `event.locals` for the load functions side and `setContext` for the svelte components.

[Rendered version](https://deploy-preview-2611--objective-northcutt-37494c.netlify.app/blog/2025/02/01/global-state-in-svelte-5/)


src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+tagline: |
+  <p>Are you using SvelteKit? Are you using Storybook with it? Well, i have good news for you because the SvelteKit integration just got a whole lot better!</p>
+
+image: "/assets/images/posts/2025-02-01-global-state-in-svelte-5/header.jpg"
+imageAlt: "The Svelte logo on a gray background picture"
+---
+
+# Runes and Global state: do's and don'ts
+
+On the 20th of September of 2023 Runes were introduced to the Svelte ecosystem. This brand new paradigm was a complete rewrite of the underlying structure of Svelte, moving from a compile only to a compile enhanced reactivity system, with a very small and extremely performant runtime footprint.
```
oscard0m:
```suggestion
[On the 20th of September of 2023 Runes were introduced to the Svelte ecosystem](https://svelte.dev/blog/runes). This brand new paradigm was a complete rewrite of the underlying structure of Svelte, moving from a _compile-only_ to a _compile-enhanced_ reactivity system, with a very small and extremely performant runtime footprint.
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
+---
+title: "Runes and Global state: do's and don'ts"
+authorHandle: paoloricciuti
+tags: [svelte]
+bio: "Paolo Ricciuti, Senior Software Engineer"
+description: "Paolo Ricciuti (paoloricciuti) shows how (but most importantly how not) handle global state in Svelte 5"
+og:
+  image: /assets/images/posts/2023-11-28-sveltekit-storybook/og-image.jpg
+tagline: |
+  <p>Are you using SvelteKit? Are you using Storybook with it? Well, i have good news for you because the SvelteKit integration just got a whole lot better!</p>
```
oscard0m: We need a tagline aligned to the blogpost:

```suggestion
tagline: |
  <p></p>
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+image: "/assets/images/posts/2025-02-01-global-state-in-svelte-5/header.jpg"
+imageAlt: "The Svelte logo on a gray background picture"
+---
+
+# Runes and Global state: do's and don'ts
+
+On the 20th of September of 2023 Runes were introduced to the Svelte ecosystem. This brand new paradigm was a complete rewrite of the underlying structure of Svelte, moving from a compile only to a compile enhanced reactivity system, with a very small and extremely performant runtime footprint.
+
+One of the greatest advantages of this brand new paradigm is what was introduced as "Universal Reactivity".
```
oscard0m:
```suggestion
One of the greatest advantages of this brand new paradigm is what was introduced as ["Universal Reactivity"](https://svelte.dev/tutorial/svelte/universal-reactivity).
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Previously on Svelte 4
+
+Before the release of runes Svelte lived in two worlds: the magical and fantastical world inside a Svelte component and the harsher reality of a Javascript file. Inside a `.svelte` file variables were automatically reactive and the labeled statement allowed you to react to changes. In a `.js` file however Svelte had no power and you had to use stores to reactively interact with data.
```
oscard0m:
```suggestion
Before the release of runes, Svelte lived in two worlds: the magical and fantastical world inside a Svelte component and the harsher reality of a Javascript file. Inside a `.svelte` file, variables were automatically reactive and the [labeled statement](https://svelte.dev/docs/svelte/legacy-reactive-assignments) (`$:`) allowed you to react to changes. In a `.js` file however Svelte had no power and you had to use stores to interact reactively with data.
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Stores
+
+A store in Svelte was an observable: an object with a `subscribe` method (that returned an `unsubscribe` function) and potentially a `set` and an `update` methods. You could `subscribe` to a store passing as an argument a function that would be invoked whenever a new value was `set` on the store.
```
oscard0m:
```suggestion
A store in Svelte was an observable: an object with a `subscribe()` method (that returned an `unsubscribe()` function) and potentially a `set()` and an `update()` methods. You could `subscribe()` to a store passing as an argument a callback function that would be invoked whenever a new value was `set()` on the store.
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+inside a svelte component or
+
+```ts
+import { count } from "./my-count";
+
+count.update($count => $count + 1);
+```
+
+inside a javascript file.
+
```
oscard0m, line 60: I would cook more this block with the upcoming one: What are the implications of doing this the old way? Maybe an external link where that is explained (like this one? https://svelte.dev/blog/runes#Signal-boost)?

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+## What about Svelte 5?
+
+Well, firstly let's make it clear: stores are not deprecated and you continue to use them in Svelte 5 just like we just explained.
+
+That said there's a reason why you are here: you want to know how to do this with the new and performant signal based reactivity system.
+
+### A quick look at runes
+
+Before we begin let's take a brief look at how runes work: runes are magical symbols that instruct the compiler that we want something special. They work inside `.svelte` files but also inside `.svelte.js` files. The simplest runes is `$state` which, as you might have guessed, is used to declare a stateful variable.
```
oscard0m:
```suggestion
Before we begin let's take a brief look at how runes work: runes are magical symbols that instruct the compiler that we want something special. They work inside `.svelte` files but also inside `.svelte.js` files. The simplest rune is `$state` which, as you might have guessed, is used to declare a stateful variable.
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+```ts
+let count = $state(0);
+
+// use the variable in the template
+
+count++; // this will update the template
+```
+
+as you can see you don't need to `import` anything: runes are just part of Svelte, the language.
```
oscard0m:
```suggestion
as you can see, you don't need to `import` anything: runes are just part of Svelte, the language.
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+let double = $derived(count * 2);
+let triple = $derived.by(() => count * 3);
+
+count++;
+console.log(double, triple); // this will log 2 and 3
+```
+
+the third piece of the puzzle is `$effect`...this is used to sync your state with something that is not reactive by nature (Svelte for example uses `effects` to sync your state with the DOM).
+
+**N.b. using `$effect` is considered an escape hatch...most of the time you don't need to use it**
```
oscard0m, line 94: Do we have something better to write this in eleventy? (a note box?) Maybe using quote instead?

> N.b. using `$effect` is considered an escape hatch...most of the time you don't need to use it

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+```ts
+let count = $state(0);
+$effect(() => {
+  console.log(count);
+});
+
+count++; // this will log 1 within the effect
+```
+
+One very relevant piece of information is how reactivity can cross module or function boundaries. Why? Because for the stateful variable to be "live" and react to changes cross module/function it needs to be enclosed in a closure. I know, these are a lot of complicated words but the gist of it is that if you export a stateful variable as is from a module the value will be "frozen" at the moment you import it.
```
oscard0m, line 107:
```suggestion
One very relevant piece of information is how reactivity can cross module or function boundaries. Why? Because for the stateful variable to be "alive" and react to changes cross module/function it needs to be enclosed in a closure. I know, these are a lot of complicated words but the gist of it is that if you export a stateful variable as is from a module the value will be "frozen" at the moment you import it.
```

oscard0m: Is it possible to link this to a REPL with this example (even an iframe?) so people can see this explanation in action?

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+If you do this
+
+```ts
+let count = $state(0);
+
+export { count };
+```
+
+the moment you import `count` in another file only the current value of count will be imported (this is a JS limitation unfortunately).
```
oscard0m, line 117: Is it possible to link this to a REPL with this example (even an iframe?) so people can see this explanation in action?

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+And i'm ready to bet that this is exactly the reason you searched for this article. So let's dive right into it...
```
oscard0m:
```suggestion
And I'm ready to bet that this is exactly the reason you searched for this article. So let's dive right into it...
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Runic Global State
+
+As i've said
```
oscard0m:
```suggestion
As I've said
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+> for the stateful variable to be "live" and react to changes cross module/function it needs to be enclosed in a closure
```
oscard0m, line 125:
```suggestion
> for the stateful variable to be "alive" and react to changes cross module/function it needs to be enclosed in a closure
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+but what does it means? It's very simple: you need a function!
```
oscard0m:
```suggestion
but what does this mean? It's very simple: you need a function!
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+export function get() {
+  return count;
+}
+
+export function set(new_count) {
+  count = new_count;
+}
+```
+
+just by doing this our count variable will now be fully reactive
```
oscard0m:
```suggestion
just by doing this, our `count` variable will now be fully reactive
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+}}>
+	{get()}
+</button>
+```
+
+Thanks for reading, bye!
+
+...
+
+I know, I know, this is not the best api. Let's try to do better: instead of exporting a function that return our `count` we can export an object with a getter and a setter!
```
oscard0m:
```suggestion
I know, I know, this is not the best API. Let's try to do it better: instead of exporting a function that return our `count` we can export an object with a _getter_ and a _setter_!
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+</script>
+
+<button onclick={()=>{
+	counter.value++;
+}}>
+	{counter.value}
+</button>
+```
+
+Much nicer right? We can do even better...when you pass an object to `$state` svelte cleverly wrap that object in a Proxy (you can read more about proxies on the official [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)...they are a fascinating feature of the language). To make it short: with proxies we wrap every property of the object with a getter and a setter for you. So our declaration file become even shorter.
```
oscard0m:
```suggestion
Much nicer right? We can do even better...when you pass an object to `$state` svelte cleverly wrap that object in a Proxy (you can read more about proxies on the official [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)...they are a fascinating feature of the language). To make it short: with proxies, we wrap every property of the object with a _getter_ and a _setter_ for you. So our declaration file becomes even shorter.
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+Here's a global counter implementation with classes:
+
+```ts
+class Counter {
+  value = $state(0);
+}
+
+export const counter = new Counter();
+```
```
oscard0m, line 208: What do you think about this suggestion?
```suggestion
```ts
class Counter {
  constructor(initialValue: number) {
     this.value = $state(initialValue)  
  }
}

export const counter = new Counter(0);
```
```

File: src/posts/2025-02-01-global-state-in-svelte-5.md [hunk 20]
```diff
@@ -0,0 +1,231 @@
...
+
+and yes, you can use this in the same way.
+
+## The perils of Global State
+
+Ok, you have your answer and i know you might be tempted to close this page and run to your editor of choice to implement it but please bear with me just another second because what I'm about to explain is **VERY IMPORTANT**.
```
oscard0m:
```suggestion
Ok, you have your answer, and I know you might be tempted to close this page and run to your editor of choice to implement it, but please bear with me just another second because what I'm about to explain is **VERY IMPORTANT**.
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+Global state in general may seem pretty innocent but it's generally discouraged even for medium sized apps. Things get even more dangerous when your app is what we call an Isomorphic Application. Isomorphic is just a fancy word to say that your app runs in different environment: most app nowadays run in two phases: a server side and a client side.
```
oscard0m:
```suggestion
Global state in general may seem pretty innocent but it's generally discouraged even for medium sized apps. Things get even more dangerous when your app is what we call an Isomorphic Application. Isomorphic is just a fancy word to say that your app runs in different environments: most apps nowadays run in two phases: a _server-side_ and a _client-side_.
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+When a request hit our server a Node process accept the request and proceed to render the right page, this means importing svelte components, executing them, agglomerate all the results and craft a valid HTML page that includes a script tag that, when on the client, will "hydrate" our application by attaching listeners, effects and such.
```
oscard0m:
```suggestion
When a request hits our server, a Node process accepts the request and proceeds to render the right page, this means importing svelte components, executing them, agglomerate all the results and craft a valid HTML page that includes a script tag that, when on the client, will "hydrate" our application by attaching listeners, effects and such.
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+This is very important because our global state is technically alive in both environments. On the server there's a global `count` variable that get's recreated on the client. But the twos lives in two separate computers!! The server one lives on our server which serve all our requests. The client one lives in our users browsers.
```
oscard0m:
```suggestion
This is very important because our global state is technically alive in both environments. On the server there's a global `count` variable that gets recreated on the client. But both live on two separate computers! The server one lives on our server which serves all our requests. The client one lives in our users' browsers.
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+What does this means? It means that if we have a true global state we need to be very attentive of where we update it: if in any moment we update `count` in an imported JS module or in the script tag of a rendered component we will increment the value of `count` on the server which means that the next request the value will not be `0` anymore but `1` (and so on for the next requests).
```
oscard0m:
```suggestion
What does this mean? It means that if we have an actual global state we need to be very attentive to where we update it: if at any moment we update `count` in an imported JS module or in the `<script>` tag of a rendered component we will increment the value of `count` on the server which means that the subsequent request the value will not be `0` anymore but `1` (and so on for the next requests).
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+This is probably fine for a counter but imagine if instead this was our user profile?
+
+_"But if i reassign it every time it's fine right?"_
+
+Well, still no: it might be good as long as you don't have asynchronous code but as soon as you do a `fetch` request (which is almost a guarantee) here's what could happen:
```
oscard0m:
```suggestion
Well, still not: it might be good as long as you don't have asynchronous code, but as soon as you do a `fetch` request (which is almost a guarantee), here's what could happen:
```

src/posts/2025-02-01-global-state-in-svelte-5.md
```diff
@@ -0,0 +1,231 @@
...
+
+![a diagram showing how multiple async request could mutate global state causing the read of the wrong value](/assets/images/posts/2025-02-01-global-state-in-svelte-5/global-state-diagram.png)
```
oscard0m, line 230: 👍🏽

oscard0m (commented):
LGTM! I would like to read the missing part to give an overall overview. It may be a bit too long once the article is complete?

src/posts/2025-02-01-global-state-in-svelte-5.md
paoloricciuti: Mmm i'm not sure this needs much more...is meant to be just a light introduction of stores to refresh everyones ideas. The implication of global state imho should come later.

src/posts/2025-02-01-global-state-in-svelte-5.md
paoloricciuti: Tbf i don't know...i don't think i've see much in other articles and a quote block feels wrong.

src/posts/2025-02-01-global-state-in-svelte-5.md
paoloricciuti: Good idea with the repl (I think the iframe would look bad tho)

paoloricciuti: Also i like live more since live binding is a proper javascript concept

src/posts/2025-02-01-global-state-in-svelte-5.md
paoloricciuti: You can add a constructor if you want to make it more dynamic but in this case counter is basically just a vessel. I don't think it's worth making the code more complex for the sake of reusability...also the example is wrong, you would still need to declare value about as `$state` and assign it in the constructor

KevinBongart:
@IdeliaRisella Could you make an og:image for @paoloricciuti please? Ideally by Friday or Monday

IdeliaRisella:
og-image added

oscard0m (approved):

PR #2627, opened 2025-02-06
Capture more error data in Sentry

marcoow:
This captures more data in the browser so we got more context when errors happen on the website. All errors on the website are high priority for us because they typically happen when submitting forms which is the way that potential clients reach out to us – if anything goes wrong with that, we obviously want to know it happened and why (and ideally to who so we can reach out to them). Specifically, this adds:

* https://docs.sentry.io/platforms/javascript/configuration/integrations/httpclient/
* https://docs.sentry.io/platforms/javascript/configuration/integrations/captureconsole/

### DIFF
--- a/src/assets/js/app.js
+++ b/src/assets/js/app.js
@@ -8,6 +8,8 @@ import bindSelectDropdowns from "./select";
 if (window.location.host === "mainmatter.com") {
   Sentry.init({
     dsn: "https://43f7627909d94dc4a769340ad730f1a2@o68744.ingest.sentry.io/4504039028817920",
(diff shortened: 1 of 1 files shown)


BobrImperator (commented):
Looks good, but I'm wondering wheter `browserTracing` would be enough.

That's what we do over on WW and we seem to be getting both console logs and instrumented `fetch` requests.

https://github.com/mainmatter/whirlwind-chat/blob/main/web-client/src/hooks.client.ts#L13

marcoow:
I don't think we need tracing though since mainmatter.com is 99% a static page anyway

BobrImperator (approved):

PR #2647, opened 2025-02-28
Improve contact form

marcoow:
This adds some missing services to the contact form (e.g. general Rust and Svelte) and prefills the service with the correct one depending on where the user came from.

### DIFF
--- a/src/rust-consulting.njk
+++ b/src/rust-consulting.njk
@@ -147,7 +147,7 @@ og:
         {%
           set link = {
           "label": 'Book a 1:1 call with us',
-          "url": '/contact/?service=rust'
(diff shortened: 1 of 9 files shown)


src/rust-consulting.njk
```diff
@@ -147,7 +147,7 @@ og:
         {%
           set link = {
           "label": 'Book a 1:1 call with us',
-          "url": 'https://mainmatter.com/contact/'
+          "url": '/contact/?service=rust'
```
marcoow: @LukeMathWalker we had changed this when you went on parental leave. I suggest we add a link to your calendar again since that had worked well in a few cases iirc. Can you give me a link?

LukeMathWalker: https://calendly.com/luca-palmieri/talk-to-a-rust-expert

PR #2652, opened 2025-03-06
Migrate scss files to @use syntax

pichfl:
Removes deprecations in dart-sass in preparation for future CSS updates.


mansona (approved):
It's a shame we don't have percy to double check this one 😞 but it looks good to me, and the fact that gravity didn't notice any change at all in the CSS size likely means that this is a 1-1 upgrade 👍

PR #2654, opened 2025-03-06
Lint scss files

pichfl:
Extends linting commands to cover our scss files


mansona (approved):
No functional changes in this PR 👍

PR #2656, opened 2025-03-06
render HTML inside rss/atom feeds

### LINKED ISSUE #2582: RSS feed looks broken
The Mainmatter blog feed has always looked broken to me, and is the only feed that looks like this in the many RSS feeds I read:

Screenshot 2024-12-09 at 3 48 43 PM

Screenshot from [Feedbin](https://feedbin.com), but I see the same with the Reeder iOS app which uses the Feedbin backend.

Looks like HTML tags aren't sanitized properly?

Could be related: [the feed is invalid](https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fmainmatter.com%2Ffeed.xml) because:
> XML parsing error: <unknown>:3:0: XML or text declaration not at start of entity

Indeed, https://github.com/mainmatter/mainmatter.com/blob/master/src/feed.njk#L15 shows that `<!-- prettier-ignore-start -->` gets rendered even though the [XML declaration](https://developer.mozilla.org/en-US/docs/Web/XML/XML_introduction#xml_declaration) should be the very first line in the response. Maybe that breaks the parsers on my end?

pichfl:
Closes #2582

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,3 +2,6 @@ lib/generate-blog/lib/templates/**/*
 lib/generate-resources/lib/templates/**/*
 lib/global-css/css/vendor/normalize.css
 src/ui/components/Header/template.hbs
+
+src/feed.njk
+src/atom.njk
--- a/src/atom.njk
+++ b/src/atom.njk
@@ -11,7 +11,6 @@ metadata:
     name: "Mainmatter GmbH"
     email: "info@mainmatter.com"
 ---
-
 <?xml version="1.0" encoding="utf-8"?>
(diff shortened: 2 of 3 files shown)


KevinBongart:
"Congratulations! This is a valid RSS feed."

❤️

KevinBongart (approved):
I'm approving with a superficial understanding of the fix, but it's close enough to what I know of rendering templates: marking the rendered content as safe HTML.

Feel free to get an additional approval from an actual Mainmatter dev

marcoow (approved):

PR #2658, opened 2025-03-07
Update eleventy

mansona:
This was mostly a drop-in update, the only issue that needed to be fixed was a problem with the sitemap plugin. From what I could see online this isn't actually needed any more 🤔 So I removed it 👍 

Edit: In fact this fixes our sitemap 🙈 if you go to https://deploy-preview-2658--objective-northcutt-37494c.netlify.app/sitemap.xml vs https://mainmatter.com/sitemap.xml you'll see the deployed version has an XML error 😞

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -11,7 +11,6 @@ const syntaxHighlightPlugin = require("@11ty/eleventy-plugin-syntaxhighlight");
 const Image = require("@11ty/eleventy-img");
 const pluginRss = require("@11ty/eleventy-plugin-rss");
 const eleventyNavigationPlugin = require("@11ty/eleventy-navigation");
-const sitemap = require("@quasibit/eleventy-plugin-sitemap");
 
 const contentParser = require("./utils/transforms/contentParser.js");
 const htmlMinTransform = require("./utils/transforms/htmlmin.js");
@@ -156,11 +155,6 @@ module.exports = function (eleventyConfig) {
     init,
   });
   eleventyConfig.addPlugin(eleventyNavigationPlugin);
-  eleventyConfig.addPlugin(sitemap, {
-    sitemap: {
-      hostname: config.url,
-    },
-  });
 
   /**
    * Add Shortcodes
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -20,7 +20,6 @@ const { init } = require("./utils/svelteSyntaxHighlight");
 /**
  * Import site configuration
  */
-const config = require("./src/_data/config.js");
 const pathConfig = require("./src/_data/paths.json");
 
 module.exports = function (eleventyConfig) {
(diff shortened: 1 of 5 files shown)


marcoow:
They are both broken actually – I broke it with 7f22a53e67d91acd9d5d8d317585e6c895936c5f. The problem is the extra newline – fixed in #2659

marcoow (approved):

marcoow:
The sitemap isn't quite right still – there's a bunch of entries like:

```xml
<url>
  <loc>.</loc>
  <lastmod>2025-04-02T00:00:00.000Z</lastmod>
</url>
```

I found this with Gravity which showed the sitemap growing by 12KB with this PR which I found strange and looked more closely… 🎉

mansona:
So I've fixed the issue that you described, the issue was that the plugin we were using was filtering on items in the all collection that have a URL (eleventy is so strange 🙈) 

Gravity is reporting the sitemap to be smaller because the new basic way to do it doesn't walk through paginated collections and we don't get each of the paginated URLs in the sitemap. here is an example of what is being removed: 

```
<loc>https://mainmatter.com/blog/page/2/</loc>
<lastmod>2025-03-10T07:06:09.574Z</lastmod>
</url>
<url>
<loc>https://mainmatter.com/blog/page/3/</loc>
<lastmod>2025-03-10T07:06:09.574Z</lastmod>
</url>
<url>
<loc>https://mainmatter.com/blog/page/4/</loc>
<lastmod>2025-03-10T07:06:09.574Z</lastmod>
</url>
<url>
<loc>https://mainmatter.com/blog/page/5/</loc>
<lastmod>2025-03-10T07:06:09.574Z</lastmod>
</url>
<url>
<loc>https://mainmatter.com/blog/page/6/</loc>
<lastmod>2025-03-10T07:06:09.574Z</lastmod>
</url>
```

This is all fine because there is no **content** on these pages so we don't actually need to tell Google about them 👍 

This isn't something that we can fix because the plugin we were using throws an error about "using templateContent too early". This was reported here (with no response): https://github.com/quasibit/eleventy-plugin-sitemap/issues/82

marcoow:
> Gravity is reporting the sitemap to be smaller

It's reporting the sitemap to be **larger** though?

Bildschirmfoto 2025-03-10 um 13 54 26

I'd like to understand what changes there before merging this. Your explanation suggests we don't really know why it's growing?

mansona:
ah, that's because every instance of `https://mainmatter.com` has been replaced with `https://deploy-preview-2658--objective-northcutt-37494c.netlify.app/` because of the netlify deploy URL and those extra characters are adding up past the bits that have been removed 👍

marcoow:
ah, that makes sense

marcoow:
@mansona please review and approve in Gravity yourself, I sent you an invite – good way to test the invitation flow.

marcoow (approved):

PR #2662, opened 2025-03-10
Ad functionality to generate og images automatically

mansona:
This is currently opt-in, you would need to add `autoOg: true` to your blog post and it will create one for you. If we're happy with it after a while we can move everything over to it 👍 

I applied this to a post locally to compare and here is a before and after: 

Before: 

og-image

After: 

2a3911b1


marcoow (approved):
Nice 🙌 This is a big help indeed – didn't realize it was that easy (seemingly) to build

PR #2666, opened 2025-03-11
fix: correct publication date for blog post

paoloricciuti:
Duh, forgot to change the publication date


marcoow (commented):
Let's please rename the folder that contains the images accordingly. Also, we should add a redirect rule from the old date

paoloricciuti:
> Let's please rename the folder that contains the images accordingly. Also, we should add a redirect rule from the old date

@marcoow done

marcoow (approved):

marcoow:
@paoloricciuti good to go – please review in Gravity and approve there (I want everyone to use that – the more eyes on it we get, the more likely we are to catch any errors that are left) before merge

paoloricciuti:
> @paoloricciuti good to go – please review in Gravity and approve there (I want everyone to use that – the more eyes on it we get, the more likely we are to catch any errors that are left) before merge

Done...i was getting file changes for file i didn't touch in this pr...is that expected?

marcoow:
That's not really expected, no. Although: if your PR branch is not up to date and it compares to latest master, it will show seemingly unrelated differences (just like any visual regression tool will). Can you check that's the case perhaps? We should make sure there are no substantial bugs before we launch next week

paoloricciuti:
> That's not really expected, no. Although: if your PR branch is not up to date and it compares to latest master, it will show seemingly unrelated differences (just like any visual regression tool will). Can you check that's the case perhaps? We should make sure there are no substantial bugs before we launch next week

image

Yup this is what i got...let's see if updating the branch fixes it

paoloricciuti:
Nope still got the same result...gonna report on the channel

marcoow:
nvm, I think it's expected – on all of those pages above, the link to the post changes which results in tiny changes

paoloricciuti:
> nvm, I think it's expected – on all of those pages above, the link to the post changes which results in tiny changes

Ohhhhhh i would have never guessed 😄

However this can still be confusing...what about adding an option to show the actual diff in gravity?

marcoow:
Yes, I think that would be good in general. Anyway, not before launch next week :) Can you create an issue for it though (describing this scenario and why it would be useful to have)?

paoloricciuti:
> Yes, I think that would be good in general. Anyway, not before launch next week :) Can you create an issue for it though (describing this scenario and why it would be useful to have)?

Cool, will do

PR #2672, opened 2025-03-14
Improve Rust landing page

marcoow:
* make the text more concrete about what we're helping with with Rust. We aren't specific enough about this currently.
* add Redis case


src/cases/redis.md
umbrellarocket, line 1: How about
> Our team collaborated closely with their engineers throughout the process, providing real-time Rust training along the way.

marcoow: improved – needs to be present tense

umbrellarocket (approved):

PR #2712, opened 2025-05-12
disable contat form until initialized

marcoow:
We sometimes receive form submission without any data which of course leads to an error. This does a few things to make that less likely to happen:

* forms are disabled until JS has initialized the form
* on submission, we explicitly trigger validation (and displaying the results) of all fields
* enable Sentry tracing so we get better insights into what happens

### DIFF
--- a/src/assets/js/contact-form.js
+++ b/src/assets/js/contact-form.js
@@ -13,15 +13,21 @@ export class ContactForm {
     this.prefillService();
 
     this.bindEvents();
+
+    this.enableForm();
   }
 
   bindEvents() {
     this.form.addEventListener("submit", event => {
-      event.preventDefault();
-      this.updateFormState("loading", "Your message is being sent...");
+      if (this.form.reportValidity()) {
+        event.preventDefault();
+        this.updateFormState("loading", "Your message is being sent...");
 
-      const formData = new FormData(this.form);
-      this.sendMessage(Object.fromEntries(formData.entries()));
+        const formData = new FormData(this.form);
+        this.sendMessage(Object.fromEntries(formData.entries()));
+      } else {
+        event.preventDefault();
+      }
     });
 
     this.reset.forEach(reset => {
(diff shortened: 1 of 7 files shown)


src/components/contact-form.njk
```diff
@@ -17,8 +17,9 @@
         data-action="https://contact.mainmatter.dev/send"
         data-method="post"
         data-goal="Contact"
+        disabled
```
BobrImperator: This attribute is unnecessary and is also never cleaned up.

Some other things:
1. When fieldset is disabled, there's no visual indication that it's in a disabled state.
2. Is it necessary to disable the inputs instead of just the submit button?

marcoow: fixed – I'd address the other things when Xenia is bad, really want to get a fix out since every submission that doesn't work is potentially lost revenue.

marcoow: We could disable just the submit button but I think this way it's better – letting people put in stuff and then never letting them submit is only frustrating.

src/assets/js/contact-form.js
```diff
@@ -13,15 +13,21 @@ export class ContactForm {
...
-      this.sendMessage(Object.fromEntries(formData.entries()));
+      if (this.form.reportValidity()) {
+        event.preventDefault();
+        this.updateFormState("loading", "Your message is being sent...");
+
+        const formData = new FormData(this.form);
+        this.sendMessage(Object.fromEntries(formData.entries()));
+      } else {
+        event.preventDefault();
+      }
```
BobrImperator: It really calls `preventDefault` always here so it can be hoisted but overall looks good
```suggestion
      event.preventDefault();
      if (this.form.reportValidity()) {
        this.updateFormState("loading", "Your message is being sent...");

        const formData = new FormData(this.form);
        this.sendMessage(Object.fromEntries(formData.entries()));
      }
```

BobrImperator (approved):
Looks good to me, checked /contact, /travel, /svelte-subscription and it seems like forms are becoming interactive once JS loads.

PR #2715, opened 2025-05-16
Fix forms

marcoow:
I presume we might be handling GET requests wrongly. This fixes those by

* making the checking for get requests more solid
* instead of manually serializing the query string, using stable APIs

### DIFF
--- a/src/assets/js/contact-form.js
+++ b/src/assets/js/contact-form.js
@@ -90,7 +90,7 @@ export class ContactForm {
       method,
       mode: "cors",
     };
-    if (method.toLowerCase().match(/get/)) {
+    if (method.toLowerCase().includes("get")) {
       action.search = new URLSearchParams(formData);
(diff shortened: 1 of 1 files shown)


src/assets/js/contact-form.js
```diff
@@ -78,15 +83,15 @@ export class ContactForm {
...
 
     let params = {
       cache: "no-cache",
       method,
       mode: "cors",
     };
-    if (method.toLowerCase() === "get") {
-      const queryString = new URLSearchParams(formData).toString();
-      action = `${action}?${queryString}`;
+    if (method.toLowerCase().match(/get/)) {
```
BobrImperator: I think `includes` would be a bit more appropriate.
```suggestion
    if (method.toLowerCase().includes("get")) {
```

BobrImperator (approved):

PR #2719, opened 2025-05-20
feat(contact-form): add request instrumentation

BobrImperator:
- Adds ContactForm.makeRequest method that wraps around `fetch` and instruments the request manually.
- Refactors the code so the http call itself is infallible and is always reported.


marcoow (approved):
simplified a little bit; @BobrImperator: please merge when you think it's good

BobrImperator:
@marcoow We're getting the network requests correctly now :+1:

PR #2728, opened 2025-06-06
feat(blog-post): Introducing Whirlwind Chat

BobrImperator:
- [x] replace og image
  - We're using an autoOg
- [x] replace imageAlt

related to #2731


src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,68 @@
+---
+title: "Whirlwind: Meet everyone in 2-minute, 1-on-1 video chats"
```
KevinBongart: I still think it should say "in the browser" or something. Finding a title for this announcement is quite tricky :/

KevinBongart:
Is the product called Whirlwind Chat? I tend to write Whirlwind, I see also "whirlwind.chat" in the blog post. Let's pick one (I believe it's "Whirlwind Chat") and make sure it's the same everywhere in the post, plus a link to https://whirlwind.chat on the first 1 or 2 mentions and the last mention

marcoow:
I think the blog post should go more into technical detail really.

* We can leave out the sales-y intro and just link to the to-be-written case study for describing the product
* Instead of the diagrams (I think we can leave out the detailed WebRTC diagram completely anyway), I'd focus on the technical overview of what we're building first: peer-to-peer for the calls but to establish connections a mediator is needed; plus, we need a server to manage the pairs
* then, how we built this which I assume can focus on the Rust part (guess the Svelte part is rather standard/boring), plus the fact that since it's Rust and the calls are peer-to-peer, we're able to handle large numbers of users with minimal resources

marcoow:
also, btw: we'll replace the Rust and Svelte CTAs with one specific for the product, similar to what we have for Gravity: https://mainmatter.com/blog/2025/03/19/gravity/

KevinBongart:
Thanks @marcoow, good feedback. @BobrImperator here's my attempt, feel free to edit/reuse/discard as you see fit.

```md
# Whirlwind: peer-to-peer video chat in the browser for hybrid events

Spontaneous one-on-one conversations are still hard to replicate at online events. We built [Whirlwind](https://whirlwind.chat) to make that easier. It's a simple app for short, peer-to-peer video chats. You join a group then get matched with others for 2-minute conversations.

If you're more interested in how Whirlwind works as a product for events, check out this [case study on using Whirlwind at hybrid meetups](TODO). This post focuses on how we built it and how we kept the infrastructure simple and scalable.

We wanted it to work for a large number of users with minimal infrastructure and no app to install, which meant serving Whirlwind as a web app and using peer-to-peer video connections.

## Built with Rust, SvelteKit, and WebRTC

Whirlwind has two parts: a lightweight frontend web app written in Svelte, and a backend server written in Rust. The frontend runs entirely in the browser and handles video calls using WebRTC. The backend coordinates users, manages sessions, and helps peers connect.

The actual video and audio data never touch our servers. Everything flows directly between browsers using peer-to-peer connections. This approach improves privacy (no server sits in the middle watching calls) and it makes Whirlwind scalable to a large number of users with minimal infrastructure.

## The backend

The web server is written in Rust using [Axum](https://docs.rs/axum/latest/axum/) for HTTP/WebSocket routing and [Tungstenite](https://github.com/snapview/tungstenite-rs) for WebSocket communication. Whirlwind uses [Tokio](https://tokio.rs) tasks spawned on demand for managing sessions, allowing each session to be configured independently.

The backend handles:

- user sessions and group coordination
- readiness states
- matching pairs
- helping clients exchange the necessary information to set up a direct video connection <- note: is this correct that the backend coordinates all this?

Each session gets its own task (a session server) that manages its participants and handles signaling logic. These tasks are short-lived and isolated from each other, which simplifies error handling and helps with scaling.

Since the backend never processes the video calls, and thanks to [Rust efficiency](https://mainmatter.com/rust-consulting), its workload stays light even with many concurrent users.

## The frontend

We used [SvelteKit](https://svelte.dev/docs/kit/introduction) for the frontend. It's a good fit for small, reactive UIs with tight performance requirements. It also kept bundle size down and helped us ship quickly without needing a complex build setup. (At Mainmatter, [we like Svelte and SvelteKit](https://mainmatter.com/svelte-consulting/) because they strike the right balance between developer productivity and building lightweight, performant web apps.)

The hard part wasn't building the interface, it was making it work reliably across all the different browsers, devices, and hardware users bring. Some users join from phones, others from dual-screen desktops. Microphones and cameras vary. Permission prompts behave differently across OS/browser combinations.

We also had to handle stream negotiation, dynamic device selection, and failure cases where the camera or microphone is missing, in use elsewhere, or blocked. Optional background blur added one more layer of complexity, requiring custom video processing in the browser. (<- note from Kevin: is this correct?)

## WebRTC

Whirlwind uses [WebRTC](https://webrtc.org) to handle video and audio calls. It's built into all modern browsers and allows two users to stream media directly to each other, without passing through a central server. This fits Whirlwind's goals well of keeping infrastructure lightweight and ensuring that no server sees or stores the contents of a conversation.

To establish a connection, browsers first need to exchange network and media details, a process called Interactive Connectivity Establishment (ICE). The Rust backend coordinates this exchange. Once both sides complete it, the video call becomes fully peer-to-peer.

While powerful, WebRTC brings a few challenges:

- Browsers differ in how they handle signaling and connection timing
- Network conditions can change during or between calls
- Devices may appear, disappear, or fail to provide identifying labels
- Codec mismatches or renegotiation quirks can break streams unexpectedly

We've handled these cases to keep conversations stable and smooth. Once connected, WebRTC gives us exactly what we need: fast, secure, direct video calls with no installs and no central bottlenecks.

## Give it a try

Whether you're organizing a meetup or just curious to see how it works, you can try Whirlwind right now at [whirlwind.chat](https://whirlwind.chat). No sign-up, no install.
```

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,134 @@
...
+autoOg: true
+tagline: <p>Lessons learned behind Whirlwind and its structure.</p>
+
+image: "/assets/images/posts/2025-05-31-introducing-whirlwind/header.png"
+imageAlt: "Smiling and waving geometric shape folks with Whirlwind in between."
+---
+
+# Whirlwind: peer-to-peer video chat in the browser for hybrid events
+
+The goal of this blog post is to provide a technical overview for a video chat application [Whirlwind](https://whirlwind.chat/) which we’ve made, as well as give an insight into some of the more interesting parts and the intentions behind them.
```
KevinBongart: I like that you annonce the intent clearly, but as an opener it feels quite flat. Something like "Hybrid events often miss the informal chats that happen in person. With [Whirlwind], we wanted to bring that experience online. In this blog post, we'll provide a technical overview of how it works and explain the technical decisions behind it."

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,134 @@
...
+
+## The core: Rust, SvelteKit, and WebRTC
+
+Whirlwind has two parts: a web app written in Svelte, and a backend server written in Rust. The frontend runs entirely in the browser and handles video calls using WebRTC. The backend coordinates users, manages sessions, and helps peers connect.
+
+The actual video and audio data never touch our servers. Everything flows directly between browsers using peer-to-peer connections. This approach improves privacy (no server sits in the middle watching calls) and it makes Whirlwind scalable to a large number of users with minimal infrastructure.
+
+## The backend
+
+The backend is written in Rust using [Axum](https://docs.rs/axum/latest/axum/) which is somewhat split into two parts: The web server and a Supervisor that takes care of spawning session servers as they’re requested.
```
KevinBongart: The backend is written in Rust using [Axum](https://docs.rs/axum/latest/axum/). It's split into two parts: a web server and a supervisor that spawns session servers on demand.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+Whirlwind Chat has two parts: a web app written in Svelte, and a backend server written in Rust. The frontend runs entirely in the browser and handles video calls using WebRTC. The backend coordinates users, manages sessions, and helps peers connect.
+
+The actual video and audio data never touch our servers. Everything flows directly between browsers using peer-to-peer connections. This approach improves privacy (no server sits in the middle watching calls) and it makes Whirlwind Chat scalable to a large number of users with minimal infrastructure.
+
+## The backend
+
+The backend is written in Rust using [Axum](https://docs.rs/axum/latest/axum/) which is somewhat split into two parts: The web server and a Supervisor that takes care of spawning session servers as they’re requested.
+
+It also relies on PostgreSQL as a persistence layer for user records and the Cloudflare Realtime service which provides TURN servers for WebRTC.
```
KevinBongart: I think the post is missing an explanation for what TURN and STUN are, something like: "It uses Cloudflare Real-time API to request STUN and TURN server configuration. Together, STUN/TURN servers help clients discover each others and establish direct connections."

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+Session servers are responsible for:
+
+- matchmaking
+- managing real-time user states like “readiness”
+- exchanging messages between users in that group
+
+### Supervisor
+
+One of the great concerns of building the server was resilience; making sure that an unexpected failure in a single lobby doesn’t bring other lobbies with it.
```
KevinBongart: "One of our biggest concerns when building the server was resilience. An unexpected failure in a single lobby should not affect others."

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+As a result we’ve created an `ApplicationSupervisor` that takes care of spawning and isolating `LobbyServer` structs which in turn spawn their own tasks (`matchmaking`, `activity monitor`). That way a lobby is free to crash and at worst users of that particular lobby will be disconnected without spoiling the fun for others.
```
KevinBongart: "To address this, we created an `ApplicationSupervisor` that spawns and isolates `LobbyServer` structs. In turn, each lobby server spawns its own tasks (such as `matchmaking` and `activity monitor`). If a lobby crashes, only users in that lobby are affected."

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+Lobby servers don’t run always, instead they’re spawned on-demand as users join a lobby for the first time and later they’re shutdown after some period of time when a lobby is considered inactive.
```
KevinBongart: "Lobby servers don’t run continuously. They are spawned on demand as users join a lobby for the first time, and shut down after a period of inactivity."

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+Additionally we take advantage of that fact to make a request to fetch STUN/TURN server configuration to [Cloudflare Realtime API](https://developers.cloudflare.com/realtime/) before a Lobby wakes up. That is because STUN/TURN server configuration must be shipped to the end users, but making a hardcoded, long lived configuration could easily allow somebody to use our Cloudflare service. So instead the credentials are generated per-server.
```
KevinBongart: I don't quite get what fact we take advantage of here. The paragraph immediately before is describing the automatic lifecycling of lobby servers, is that related? In which case I suggest rephrasing to say something like "As part of the 
startup process, the lobby server fetches STUN and TURN configuration…"

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+Another function of a Supervisor is to provide an access to the internal state of a given lobby. We rely on this mechanism internally for owner actions which are regular HTTP calls instead and not WebSocket messages. This helps with avoiding re-implementing request/response and authentication mechanisms in a WebSocket connection, ultimately making things simpler by relying on the already coined practices around HTTP authentication.
```
KevinBongart: already coined -> reusing well-established HTTP practices

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+The Supervisor isn’t very sophisticated though and doesn’t attempt to restart a Lobby once it crashes. Instead if a Lobby or any of its child tasks encounter an unrecoverable error, the Lobby is shutdown immediately, while not great, it’s still an annoyance at best.
```
KevinBongart: Slightly tweaked: "Instead, if a lobby or any of its child tasks fail with an unrecoverable error, the lobby shuts down immediately. This isn't ideal, but in practice it's just a minor annoyance for users."

Although I'm not sure what the value is for this section. If a lobby crashes and all its users have to find out after a minute of two then reload the page to rejoin, that's pretty bad for a product so I wouldn't call it a minor annoyance. If they actually can't rejoin the lobby because it crashed and the organizer has to create a new one entirely (and thus reinvite users to a different URL), that's even worse and far from a minor annoyance.

In that case I'd emphasize that crashes are contained within each lobby (one event doesn't take down other events) and that we are working on automatic restart of lobbies that crashes, even though it's rare

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+![Server structure quick overview](/assets/images/posts/2025-05-31-introducing-whirlwind/server-structure.png)
+
+### InMemory
+
+`InMemory` is a module that stores and implements methods to add or change the `Lobby` data and it's deeply connected to a specific `Lobby`.
+
+It stores all the information needed for a `Lobby` to function i.e. matches, user readiness and their `mailboxes`, it also emits events and talks to users connected to a given `Lobby`. Essentially it facilitates all of the `User` → `User` communication and `Lobby` → `User` through a `mailbox`.
```
KevinBongart: `InMemory` is a module that stores and modifies the data associated with a specific lobby. It keeps track of matches, user readiness, and mailboxes. It also emits events and communicates with users in that lobby.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+A `mailbox` is an `mpsc` channel of each users’ Websocket connection when they connect to a `Lobby`. Without it users wouldn’t be able to talk to each other or get notified by the `Lobby`.
```
KevinBongart: A `mailbox` is a message channel (specifically, an `mpsc` channel, or multiple producer, single consumer) linked to each user's WebSocket connection. When users join a lobby, they receive a mailbox so that the lobby or other users can send them messages directly.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Websocket
```
KevinBongart: WebSocket

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+A great chunk of Whirlwind Chat functionalities are a part of Websocket connections.
+
+It serves a few purposes:
+
+- Server must be able to notify users of state changes and new matches in real-time.
+- Users must be able to send a dozen messages as fast as possible to each other when negotiating a WebRTC connection.
+- Keep track of whether users are still participating in the session or if they’ve quit or lost connection.
```
KevinBongart: A large part of Whirlwind’s functionality relies on WebSocket connections. These connections are used to notify users of real-time changes (such as state or matches), fast exchange of messages during the WebRTC negotiation, and track whether users are still connected.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+The Websocket is pretty simple, its role is to receive and react to messages from a user. A Websocket connection is a `tokio::task` that `axum` spawns for us and provides us with a `stream`. Additionally when the Websocket connection is already upgraded, when "connecting" to a `Lobby`, the handler takes a hold of an `InMemoryHandle` that is the de-facto interface of `InMemory` that give us read and write access to `InMemory` via `Actor` model messaging using `oneshot` channels.
```
KevinBongart: As a rule of thumb, it's better to avoid "just do x", "simply connect to y" or "is pretty simple" (it didn't seem simple to me as a reader, unless you meant that its design and scope are limited, maybe "focused" to something more straightforward?)

"The WebSocket connection is managed as a `tokio::task` spawned by Axum, which provides a stream of messages. When a user connects to a lobby, the handler also takes ownership of an `InMemoryHandle`, a message-passing interface for reading and writing lobby state via an actor-style model using `oneshot` channels."

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+By itself it can't interact with the rest of the system or users however. In order to handle that the Websocket spawns more tasks and channels. It creates a `mailbox` and forwards the writer side of it to the `InMemory` so the `InMemory` module and other users have the ability to push messages to that user who owns the receiving side of `mailbox`. It also creates a `queue` channel - once a message reaches the channel it's read by the `sender` task which pushes that message to the client. The `queue` collects messages from other sources like the `mailbox` and `receiver`.
```
KevinBongart: The WebSocket task can't interact with the rest of the system on its own. To do that, it spawns additional tasks and channels. It creates a mailbox (the sending side of an `mpsc` channel) and registers it with `InMemory`, allowing the lobby and other users to send messages to this user. It also sets up a queue channel that collects messages from multiple sources and forwards them to the client.

KevinBongart: The WebSocket task can't interact with the rest of the system on its own. To do that, it spawns additional tasks and channels. It creates a mailbox (the sending side of an `mpsc` channel) and registers it with `InMemory`, allowing the lobby and other users to send messages to this user. It also sets up a queue channel that collects messages from multiple sources and forwards them to the client.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+Messages sent to a user are generally either a result of their direct actions like sending a `Ready` message or indirect like somebody joining to the Lobby, in that case user will receive a `LobbyStatus` message originating from the session server.
```
KevinBongart: Messages sent to a user can be triggered by their own actions (such as sending a `Ready` message) or by external events (like another user joining). For example, when someone joins the lobby, all connected users receive a `LobbyStatus` message from the session server.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+![Websocket communication flow](/assets/images/posts/2025-05-31-introducing-whirlwind/websocket-overview.png)
+
+### Testing
+
+Whirlwind Chat is an application where 90% of the work and functionality happens in the Websockets.
+
+As such it requires a different testing approach that’s more similar to testing `Evented systems` rather than an `HTTP API`.
```
KevinBongart: Most of Whirlwind Chat's functionality runs through WebSockets. As a result, testing it requires a different approach that is closer to testing event-driven systems than traditional HTTP APIs.

(No need for blockquotes here IMO)

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+First of all, the test setup actually runs our Server by binding it to a random port with its own database and configuration on a test by test basis - making sure everything is isolated, concurrent and fast. It also works great when testing different configurations of various `matchmaking` task settings.
```
KevinBongart: Each test runs a full server instance, bound to a random port with its own database and configuration. This keeps tests isolated, allows them to run in parallel, and supports custom matchmaking configurations.

(Also no need for blockquote around matchmaking, unless its a specific Rust keyword?)

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+The test suite essentially uses `Black Box testing` approach. Our tests define and make use of a stateful `Interactions` module. The `Interactions` module is a test helper that defines methods that make _real_ HTTP calls (not mocked), and uses `tokio_tungstenite` to make a real Websocket connection to the server, so it can act as a real user using their browser. `Interactions` also stores received messages for inspection, which helps with keeping tests fast and accurate by making sure it’s OK to receive different messages while we’re waiting for a specific one; since messages are unordered by nature and the client might receive a `UserStatus` message in between awaiting an `IceAnswer` message.
```
KevinBongart: The test suite follows a black-box approach. Tests use an `Interactions` module, which makes real HTTP requests (not mocks) and connects to the server using `tokio_tungstenite` (note: can this be a link to a library?) to simulate a real user session.

The `Interactions` module stores all received messages for later inspection. This improves reliability, since messages can arrive out of order. For example, a `UserStatus` message might appear while we are waiting for an `IceAnswer`. By collecting all messages, the test can verify outcomes without depending on timing.

KevinBongart: - What do we want?
- Now!
- A way to process asynchronous messages chronologically without having to account for clock drift!
- And when do we want it?!

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+The hard part wasn't building the interface, it was making it work reliably across all the different browsers, devices, and hardware users bring. Some users join from phones, others from dual-screen desktops. Microphones and cameras vary. Permission prompts behave differently across OS/browser combinations.
+
+We also had to handle stream negotiation, dynamic device selection, and failure cases where the camera or microphone is missing, in use elsewhere, or blocked. An optional background blur feature added one more layer of complexity by having to juggle multiple video streams and elements, which is more tricky than it sounds.
+
+![WebRTC overview](/assets/images/posts/2025-05-31-introducing-whirlwind/webrtc-overview.png)
+
+### Background blur with Tensorflow
+
+While integrating Tensorflow (BodyPix) is pretty easy, there are some tricky parts in context of a video call and WebRTC though.
```
KevinBongart: To power the background blur feature, we use [TensorFlow.js](https://www.tensorflow.org/js) together with its [BodyPix](https://github.com/tensorflow/tfjs-models/tree/master/body-pix) model. BodyPix is a machine learning model that runs entirely in the browser. It performs real-time person segmentation, which means it can identify which parts of the video are the person and which are the background. This makes it possible to blur the background while keeping the speaker in focus.

While integration was relatively straightforward, applying it to a real-time video call presented a few challenges in the context of WebRTC.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+- The person segmentation model is quite heavy and ideally is sideloaded
+- Video processing is pretty heavy on CPU and so we needed a way to limit how often segmentation runs using `requestAnimationFrame` and then limit it further by pegging it to the video stream fps by counting how much time has elapsed between draws. Current frequency is `1000ms / 30fps` .
+- It requires swapping out a `video` element with a `canvas` element when blur is toggled on. Video element must be kept in the background and overlayed with `canvas` because it’s still the video camera and audio source.
+- Video processing sometimes throws errors that’d typically stop blur from functioning. When that happens, we’re restarting the process.
```
KevinBongart: - The person segmentation model is large, so we sideload it instead of bundling it directly into the app.  
- Video processing uses a lot of CPU, so we limit how often segmentation runs using `requestAnimationFrame`. We also throttle it further by aligning it with the frame rate of the video stream. Right now, it runs at most once every 33 milliseconds, which matches 30 frames per second.  
- When blur is enabled, we overlay a canvas element on top of the video feed. The video element stays in the background because it still provides the camera and microphone stream.  
- Sometimes, video processing fails and breaks the blur feature. When that happens, we restart the process to recover automatically.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+The last tricky part we’ve had to solve was making sure which video track needs to be sent to the Peer without overly complicating things by having to send additional metadata. While WebRTC has a `replaceTrack` API, we learned quickly that when called too often, WebRTC would stop transmitting any video. The fix was to debounce the blur functionality, but finding the answer wasn’t so obvious at the time.
```
KevinBongart: One of the hardest problems was deciding which video track to send to the peer. We wanted to avoid adding extra metadata to describe the current stream. WebRTC provides a `replaceTrack` API, but calling it too frequently can cause the connection to stop transmitting video. To avoid that, we debounce the blur toggle (i.e. wait briefly before applying the change) so that track switching only happens once the user has made a final decision.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Detecting whether video is being sent
+
+This was probably our most favorite piece of the puzzle. WebRTC doesn’t provide any off the shelf API that could quickly tell that there’s likely a problem with the connection and a video stream is stuck for our Peer. We can see our camera feed OK but the person on the other end might not see us.
```
KevinBongart: This was one of our favorite challenges. WebRTC does not provide a simple API to tell whether the connection is working and video is actually being delivered to the peer. You can see your own camera feed just fine, but the person on the other end might not be receiving anything.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+Luckily, WebRTC does provide statistics for the transceiver that takes care of receiving and sending the data. It’s possible to call `getStats` on `RTCPeerConnection` instance, select a “report” for `video` channel and using `report.framesReceived` property, begin to count how many times we’ve received less than a certain amount of frames in that second. If it keeps dropping frames for roughly 10 seconds, we’ll call `restartIce` on the WebRTC connection to make clients renegotiate their connection.
+
+This mechanism works great for situations where:
+
+- Peers are not sending data to one another
+- Peers are using incompatible codecs
+- The internet connection quality has dropped or switched from wifi to mobile data.
```
KevinBongart: Luckily, WebRTC does provide connection statistics through the `getStats` method on the `RTCPeerConnection` object. We use this to monitor the video channel and look at the `framesReceived` count in each report. If the number of frames received stays low for several seconds, we assume the connection is stalled and we call `restartIce` to force renegotiation between peers. This often fixes problems caused by codec mismatches, connection drops, or switching networks during a call.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+- The person segmentation model is quite heavy and ideally is sideloaded
+- Video processing is pretty heavy on CPU and so we needed a way to limit how often segmentation runs using `requestAnimationFrame` and then limit it further by pegging it to the video stream fps by counting how much time has elapsed between draws. Current frequency is `1000ms / 30fps` .
+- It requires swapping out a `video` element with a `canvas` element when blur is toggled on. Video element must be kept in the background and overlayed with `canvas` because it’s still the video camera and audio source.
+- Video processing sometimes throws errors that’d typically stop blur from functioning. When that happens, we’re restarting the process.
+
+The last tricky part we’ve had to solve was making sure which video track needs to be sent to the Peer without overly complicating things by having to send additional metadata. While WebRTC has a `replaceTrack` API, we learned quickly that when called too often, WebRTC would stop transmitting any video. The fix was to debounce the blur functionality, but finding the answer wasn’t so obvious at the time.
+
+### Detecting whether video is being sent
```
KevinBongart: "Detecting when a video stream stops working" feels a bit clearer and intriguing to me, but up to you!

src/posts/2025-05-31-introducing-whirlwind.md
marcoow, line 1: Let's update the date to today

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
+---
+title: "Whirlwind Chat: Learnings behind building a P2P video chat in browser"
+authorHandle: BobrImperator
+tags: [rust, svelte, webrtc]
+bio: "Bartlomiej Dudzik, Software Developer"
+description: "A technical overview of Whirlwind Chat: P2P browser video chat, learnings behind it and a look at the more interesting bits."
```
marcoow:
```suggestion
description: "A technical overview of Whirlwind Chat: learnings from building a browser-based P2P video chat."
```

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
+---
+title: "Whirlwind Chat: Learnings behind building a P2P video chat in browser"
```
marcoow:
```suggestion
title: "Whirlwind Chat: Learnings from building a browser-based P2P video chat"
```

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+authorHandle: BobrImperator
+tags: [rust, svelte, webrtc]
+bio: "Bartlomiej Dudzik, Software Developer"
+description: "A technical overview of Whirlwind Chat: P2P browser video chat, learnings behind it and a look at the more interesting bits."
+autoOg: true
+tagline: <p>Lessons learned and decisions made behind Whirlwind Chat and its structure.</p>
```
marcoow: This should really be the first paragraph of the post. I'd explain what Whirlwind is in 1-2 sentences to give people context as well.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+imageAlt: "Smiling and waving geometric shape folks swept by a whirlwind."
+---
+
+# Whirlwind Chat: peer-to-peer video chat in the browser for hybrid events
+
+The goal of this blog post is to provide a technical overview for a video chat application [Whirlwind Chat](https://whirlwind.chat/) which we've made, as well as give an insight into some of the more interesting parts and the intentions behind them.
+
+## Why Whirlwind Chat exists?
+
+Spontaneous one-on-one conversations are still hard to replicate at online events. We built [Whirlwind Chat](https://whirlwind.chat/) to make that easier. It's a simple app for short, peer-to-peer video chats. You join a group then get matched with others for 2-minute conversations.
```
marcoow: I think this should be the start of the post.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+If you're more interested in how Whirlwind Chat works as a product for events, check out our [case study on using Whirlwind Chat at hybrid meetups](/cases/whirlwind).
```
marcoow: can use the above:

```suggestion
This post is to provide a technical overview for a video chat application [Whirlwind Chat](https://whirlwind.chat/) which we've made, as well as give an insight into some of the more interesting parts and the intentions behind them.
```

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+The web server handles:
+
+- HTTP API and Websocket connections
+- Session servers spawning
+
+Session servers are responsible for:
+
+- matchmaking
+- managing real-time user states like “readiness”
+- exchanging messages between users in that group
```
marcoow: I'd add these messages are for establishing the WebRTC connections, otherwise it can be confusing

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+It also relies on PostgreSQL as a persistence layer for user records and the Cloudflare Realtime service which provides TURN servers for WebRTC.
+
+The web server handles:
+
+- HTTP API and Websocket connections
+- Session servers spawning
+
+Session servers are responsible for:
+
+- matchmaking
```
marcoow:
```suggestion
- matchmaking and keeping track of previous matches
```

src/posts/2025-05-31-introducing-whirlwind.md
marcoow: I guess this could be left out – it's essentially just saying there's an open TODO but we can easily not mention it at all.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+Another function of a Supervisor is to provide an access to the internal state of a given lobby. We rely on this mechanism internally for owner actions which are regular HTTP calls instead and not WebSocket messages. This helps with avoiding re-implementing request/response and authentication mechanisms in a WebSocket connection, ultimately making things simpler by relying on the already coined practices around HTTP authentication.
+
+The Supervisor isn’t very sophisticated though and doesn’t attempt to restart a Lobby once it crashes. Instead if a Lobby or any of its child tasks encounter an unrecoverable error, the Lobby is shutdown immediately, while not great, it’s still an annoyance at best.
+
+![Server structure quick overview](/assets/images/posts/2025-05-31-introducing-whirlwind/server-structure.png)
+
+### InMemory
```
marcoow: This is coming a bit out of the blue. Without showing code, nobody knows what `InMemory` is at all. I'd rephrase this maybe to not talk about particular code elements but say we're storing session state in memory, etc.

src/posts/2025-05-31-introducing-whirlwind.md
marcoow: The backticks are confusing here. You want code formatting for code, e.g. module or function names, not technical terms in general. The patter continues below.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+The test suite essentially uses `Black Box testing` approach. Our tests define and make use of a stateful `Interactions` module. The `Interactions` module is a test helper that defines methods that make _real_ HTTP calls (not mocked), and uses `tokio_tungstenite` to make a real Websocket connection to the server, so it can act as a real user using their browser. `Interactions` also stores received messages for inspection, which helps with keeping tests fast and accurate by making sure it’s OK to receive different messages while we’re waiting for a specific one; since messages are unordered by nature and the client might receive a `UserStatus` message in between awaiting an `IceAnswer` message.
+
+## The frontend
+
+We used [SvelteKit](https://svelte.dev/docs/kit/introduction) for the frontend. It's a good fit for reactive UIs while keeping bundle size to minimum. (At Mainmatter, [we like Svelte and SvelteKit](https://mainmatter.com/svelte-consulting/) because they strike the right balance between developer productivity and building lightweight, performant web apps.)
```
marcoow:
```suggestion
We used [SvelteKit](https://svelte.dev/docs/kit/introduction) for the frontend. It's a good fit for reactive UIs while keeping bundle size to minimum. (At Mainmatter, [we like Svelte and SvelteKit](/svelte-consulting/) because they strike the right balance between developer productivity and building lightweight, performant web apps.)
```

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+The hard part wasn't building the interface, it was making it work reliably across all the different browsers, devices, and hardware users bring. Some users join from phones, others from dual-screen desktops. Microphones and cameras vary. Permission prompts behave differently across OS/browser combinations.
+
+We also had to handle stream negotiation, dynamic device selection, and failure cases where the camera or microphone is missing, in use elsewhere, or blocked. An optional background blur feature added one more layer of complexity by having to juggle multiple video streams and elements, which is more tricky than it sounds.
```
marcoow:
```suggestion
We also had to handle stream negotiation, dynamic device selection, and failure cases where the camera or microphone is missing, is in use elsewhere, or blocked. An optional background blur feature added one more layer of complexity by having to juggle multiple video streams and elements, which is more tricky than it sounds.
```

src/posts/2025-05-31-introducing-whirlwind.md
marcoow: I'd add an image or so to make clear what this is about. This comes out of the blue with no real context.

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+Luckily, WebRTC does provide statistics for the transceiver that takes care of receiving and sending the data. It’s possible to call `getStats` on `RTCPeerConnection` instance, select a “report” for `video` channel and using `report.framesReceived` property, begin to count how many times we’ve received less than a certain amount of frames in that second. If it keeps dropping frames for roughly 10 seconds, we’ll call `restartIce` on the WebRTC connection to make clients renegotiate their connection.
+
+This mechanism works great for situations where:
+
+- Peers are not sending data to one another
+- Peers are using incompatible codecs
+- The internet connection quality has dropped or switched from wifi to mobile data.
+
+## Give it a try
```
marcoow: This will be in the CTA

src/posts/2025-05-31-introducing-whirlwind.md
```diff
@@ -0,0 +1,133 @@
...
+
+## The core: Rust, SvelteKit, and WebRTC
+
+Whirlwind Chat has two parts: a web app written in Svelte, and a backend server written in Rust. The frontend runs entirely in the browser and handles video calls using WebRTC. The backend coordinates users, manages sessions, and helps peers connect.
+
+The actual video and audio data never touch our servers. Everything flows directly between browsers using peer-to-peer connections. This approach improves privacy (no server sits in the middle watching calls) and it makes Whirlwind Chat scalable to a large number of users with minimal infrastructure.
+
+## The backend
+
+The backend is written in Rust using [Axum](https://docs.rs/axum/latest/axum/) which is somewhat split into two parts: The web server and a Supervisor that takes care of spawning session servers as they’re requested.
```
marcoow:
```suggestion
The backend is written in [Rust](/rust-consulting/) using [Axum](https://docs.rs/axum/latest/axum/) which is somewhat split into two parts: The web server and a Supervisor that takes care of spawning session servers as they’re requested.
```

src/posts/2025-05-31-introducing-whirlwind.md
marcoow: I'd also add a screenshot here so people can imagine what this app looks like and how it works.

marcoow (commented):
These should be replaced with the special Whirlwind CTA introduced with the case study: #2731

Bildschirmfoto 2025-06-23 um 08 32 16

src/posts/2025-05-31-introducing-whirlwind.md
BobrImperator: This is now changed and the suggestion doesn't work anymore

src/posts/2025-05-31-introducing-whirlwind.md
BobrImperator: I'm not exactly sure what do you mean here, but the first paragraph has changed since and Whirlwind is briefly explained now.

marcoow: That content should go here. The `tagline` is in fact used for the first paragraph of the post (it's weirdly named)

BobrImperator:
@marcoow @KevinBongart  please have a look again. I've updated pretty much all of the suggestionsa and those I'm unsure about are left unresolved.

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+## The core: Rust, SvelteKit, and WebRTC
+
+Whirlwind Chat has two parts: a web app written in Svelte, and a backend server written in Rust. The frontend runs entirely in the browser and handles video calls using WebRTC. The backend coordinates users, manages sessions, and helps peers connect.
+
+The actual video and audio data never touch our servers. Everything flows directly between browsers using peer-to-peer connections. This approach improves privacy (no server sits in the middle watching calls) and it makes Whirlwind Chat scalable to a large number of users with minimal infrastructure.
+
+## The backend
+
+The backend is written in [Rust](/rust-consulting/) using [Axum](https://docs.rs/axum/latest/axum/). It's into two parts: a web server and a Supervisor that spawns session servers on demand.
```
KevinBongart:
```suggestion
The backend is written in [Rust](/rust-consulting/) using [Axum](https://docs.rs/axum/latest/axum/). It's split into two parts: a web server and a Supervisor that spawns session servers on demand.
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+It relies on PostgreSQL as a persistence layer for user records and the Cloudflare Realtime service. The Cloudflare Realtime service provides STUN/TURN servers for WebRTC, which are what allows devices to discover each other and establish a direct connection. Most connections will work well by utilizing only the STUN server, which essentially help finding a "path" to the other device and once it does the devices can use that information to connect. This doesn't work always however, when one of the devices is behind a hardened network that uses a hardened firewall or a more restrictive NAT, then a direct connection likely won't be possible - and here's where TURN servers come in. TURN servers are relays that transmit data between users.
```
KevinBongart:
```suggestion
It relies on PostgreSQL as a persistence layer for user records and the Cloudflare Realtime service. The Cloudflare Realtime service provides STUN and TURN servers for WebRTC, which allow devices to discover each other and establish a direct connection. Most connections will work well by utilizing only the STUN server, which essentially help finding a "path" to the other device and once it does the devices can use that information to connect. This doesn't always work. If one of the devices is behind a firewall or a restrictive NAT, then a direct connection likely won't be possible - and that's when TURN servers come in. TURN servers are relays that transmit data between users.
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+![WebRTC overview](/assets/images/posts/2025-06-23-introducing-whirlwind/webrtc-overview.png)
+
+### Background blur with Tensorflow
+
+To power the background blur feature, we use TensorFlow.js together with BodyPix. BodyPix is a machine learning model that runs entirely in the browser. It performs real-time person segmentation, which means it can identify which parts of the video are the person and which are the background. This makes it possible to blur the background while keeping the speaker in focus.
+
+While integration was relatively straightforward, applying it to a real-time video call presented a few challenges in the context of WebRTC.
+
+- The person segmentation model is quite heavy and ideally is sideloaded
```
KevinBongart:
```suggestion
- The person segmentation model is quite large and should ideally be sideloaded rather than bundled.
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+- Video processing is pretty heavy on CPU and so we needed a way to limit how often segmentation runs using `requestAnimationFrame` and then limit it further by pegging it to the video stream fps by counting how much time has elapsed between draws. Current frequency is `1000ms / 30fps` .
```
KevinBongart:
```suggestion
- Video processing is pretty heavy on CPU. We needed a way to limit how often segmentation runs using `requestAnimationFrame` and then limit it further by pegging it to the video frame rate by counting how much time has elapsed between draws. The current frequency is `1000ms / 30fps`.
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+- It requires swapping out a `video` element with a `canvas` element when blur is toggled on. Video element must be kept in the background and overlayed with `canvas` because it’s still the video camera and audio source.
```
KevinBongart:
```suggestion
- It requires swapping out a `video` element with a `canvas` element when blur is toggled on. The video element must be kept in the background and overlayed with `canvas` because it remains the video camera and audio source.
```

src/posts/2025-06-23-introducing-whirlwind.md
BobrImperator: :bow:

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+customCta: "global/whirlwind-cta.njk"
+tagline: <p>Spontaneous one-on-one conversations are still hard to replicate at online events. We built <a href="https://whirlwind.chat/">Whirlwind Chat</a> to make that easier. It's a simple app for short, peer-to-peer video chats. You join a group then get matched with others for 2-minute conversations.</p>
+image: "/assets/images/posts/2025-06-23-introducing-whirlwind/whirlwind-visual.jpg"
+imageAlt: "Smiling and waving geometric shape folks swept by a whirlwind."
+---
+
+## What is Whirlwind Chat
+
+Spontaneous one-on-one conversations are still hard to replicate at online events. We built [Whirlwind Chat](https://whirlwind.chat/) to make that easier. It's a simple app for short, peer-to-peer video chats. You join a group then get matched with others for 2-minute conversations.
+
```
marcoow:
```suggestion
```

this is duplicated from above

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
+---
+title: "Whirlwind Chat: Learnings from building a browser-based P2P video chat"
+authorHandle: BobrImperator
+tags: [rust, svelte, webrtc]
+bio: "Bartlomiej Dudzik, Software Developer"
+description: "A technical overview of Whirlwind Chat: learnings from building a browser-based P2P video chat."
+autoOg: true
+customCta: "global/whirlwind-cta.njk"
+tagline: <p>Spontaneous one-on-one conversations are still hard to replicate at online events. We built <a href="https://whirlwind.chat/">Whirlwind Chat</a> to make that easier. It's a simple app for short, peer-to-peer video chats. You join a group then get matched with others for 2-minute conversations.</p>
```
marcoow:
```suggestion
tagline: <p>Spontaneous one-on-one conversations are still hard to replicate at online events. We built <a href="https://whirlwind.chat/">Whirlwind Chat</a> to make that easier. It's a simple app for short, peer-to-peer video chats. You join a group, then get matched with others for 2-minute conversations.</p>
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+imageAlt: "Smiling and waving geometric shape folks swept by a whirlwind."
+---
+
+## What is Whirlwind Chat
+
+Spontaneous one-on-one conversations are still hard to replicate at online events. We built [Whirlwind Chat](https://whirlwind.chat/) to make that easier. It's a simple app for short, peer-to-peer video chats. You join a group then get matched with others for 2-minute conversations.
+
+This post is to provide a technical overview for a video chat application [Whirlwind Chat](https://whirlwind.chat/) which we've made, as well as give an insight into some of the more interesting parts and the intentions behind them.
+
+![A screenshot showcasing how Whirlwind Chat looks like on mobile](/assets/images/posts/2025-06-23-introducing-whirlwind/mobile-screenshot.png)
```
marcoow: Let's use a desktop screenshot with 2 people with actual video enabled. If we don't have that, I guess we can just sth. from Figma.

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+tagline: <p>Spontaneous one-on-one conversations are still hard to replicate at online events. We built <a href="https://whirlwind.chat/">Whirlwind Chat</a> to make that easier. It's a simple app for short, peer-to-peer video chats. You join a group then get matched with others for 2-minute conversations.</p>
+image: "/assets/images/posts/2025-06-23-introducing-whirlwind/whirlwind-visual.jpg"
+imageAlt: "Smiling and waving geometric shape folks swept by a whirlwind."
+---
+
+## What is Whirlwind Chat
+
+Spontaneous one-on-one conversations are still hard to replicate at online events. We built [Whirlwind Chat](https://whirlwind.chat/) to make that easier. It's a simple app for short, peer-to-peer video chats. You join a group then get matched with others for 2-minute conversations.
+
+This post is to provide a technical overview for a video chat application [Whirlwind Chat](https://whirlwind.chat/) which we've made, as well as give an insight into some of the more interesting parts and the intentions behind them.
```
marcoow:
```suggestion
This post is to provide a technical overview of Whirlwind, as well as give an insight into some of the more interesting parts and the intentions behind them.
```

we just said what it is and that we built it in the previous sentence

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+## The backend
+
+The backend is written in [Rust](/rust-consulting/) using [Axum](https://docs.rs/axum/latest/axum/). It's split into two parts: a web server and a Supervisor that spawns session servers on demand.
+
+It relies on PostgreSQL as a persistence layer for user records and the Cloudflare Realtime service. The Cloudflare Realtime service provides STUN and TURN servers for WebRTC, which allow devices to discover each other and establish a direct connection. Most connections will work well by utilizing only the STUN server, which essentially help finding a "path" to the other device and once it does the devices can use that information to connect. This doesn't always work. If one of the devices is behind a firewall or a restrictive NAT, then a direct connection likely won't be possible - and that's when TURN servers come in. TURN servers are relays that transmit data between users.
```
marcoow:
```suggestion
It relies on PostgreSQL as a persistence layer for user records and the Cloudflare Realtime service. The Cloudflare Realtime service provides STUN and TURN servers for WebRTC, which allow devices to discover each other and establish a direct connection. Most connections will work well by utilizing only the STUN server, which essentially help finding a "path" to the other device and once it did, the devices can use that information to connect. This doesn't always work. If one of the devices is behind a firewall or a restrictive NAT, then a direct connection likely won't be possible - and that's when TURN servers come in. TURN servers are relays that transmit data between users.
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+- matchmaking and keeping track of previous matches
+- managing real-time user states like “readiness”
+- exchanging WebRTC messages between users in that group
+
+### Supervisor
+
+One of our biggest concerns when building the server was resilience. An unexpected failure in a single lobby shouldn't affect others.
+
+To address this, we created an `ApplicationSupervisor` that spawns and isolates `LobbyServer` structs. In turn, each lobby server spawns its own tasks (such as `matchmaking` and `activity monitor`). If a lobby crashes, only users in that lobby are affected.
```
marcoow:
```suggestion
To address this, we created an `ApplicationSupervisor` that spawns and isolates `LobbyServer` structs. In turn, each lobby server spawns its own tasks (such as `matchmaking` and `activity monitor`). If a lobby crashes (which is rather unlikely anyway thanks to Rust), only users in that lobby are affected.
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+Lobby servers don’t run continuously. They are spawned on demand as users join a lobby for the first time, and shut down after a period of inactivity.
+
+Another function of a Supervisor is to provide an access to the internal state of a given lobby. We rely on this mechanism internally for owner actions which are regular HTTP calls instead and not just WebSocket messages. This helps with avoiding re-implementing request/response and authentication mechanisms in a WebSocket connection, ultimately making things simpler by reusing well established HTTP practices.
```
marcoow:
```suggestion
Another function of a Supervisor is to provide an access to the internal state of a given lobby. We rely on this mechanism internally for owner actions which are regular HTTP calls instead of WebSocket messages. This helps with avoiding re-implementing request/response and authentication mechanisms in a WebSocket connection, ultimately making things simpler by reusing well established HTTP practices.
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+![Server structure quick overview](/assets/images/posts/2025-06-23-introducing-whirlwind/server-structure.png)
+
+### WebSocket
+
+A large part of Whirlwind’s functionality relies on WebSocket connections. These connections are used to notify users of real-time changes (such as state or matches), fast exchange of messages during the WebRTC negotiation, and track whether users are still connected.
+
+The WebSocket connection is managed as a `tokio::task` spawned by Axum. When a user connects to a `Lobby`, the handler also takes ownership of an `InMemoryHandle`, a message-passing interface for reading and writing lobby state via an actor-style model using `oneshot` channels.
+
+The WebSocket task can't interact with the rest of the system on its own. To do that, it spawns additional tasks and channels. It creates a `mailbox` (the sending side of an `mpsc` channel) and registers it with `InMemory`, allowing the lobby and other users to send messages to this user. It also sets up a `queue` channel that collects messages from multiple sources and forwards them to the client.
```
marcoow: > It creates a `mailbox` (the sending side of an `mpsc` channel)

A mailbox should be for **receiving**? Otherwise it's a really misleading name…

marcoow: > and registers it with `InMemory`

What's `InMemory`? The same as `InMemoryHandle`?

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+Messages sent to a user can be triggered by their own actions (such as sending a Ready message) or by external events (like another user joining). For example, when someone joins the lobby, all connected users receive a LobbyStatus message from the session server.
+
+![WebSocket communication flow](/assets/images/posts/2025-06-23-introducing-whirlwind/websocket-overview.png)
```
marcoow: User 1 and user 2 should be in 2 separate browsers?

marcoow: I don't understand the arrows from one user's `receiver` to the other user's `mailbox` – should the arrow not be from one user's `sender` to the other user's `mailbox`?

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Testing
+
+Whirlwind Chat is an application where 90% of the work and functionality happens in the WebSockets. As such it requires a different testing approach that’s more similar to testing Evented systems rather than an HTTP API.
```
marcoow:
```suggestion
Whirlwind Chat is an application where 90% of the functionality is in the messages sent over the WebSockets. As such it requires a different testing approach that’s more similar to testing Evented systems rather than an HTTP API.
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+## What is Whirlwind Chat
+
+Spontaneous one-on-one conversations are still hard to replicate at online events. We built [Whirlwind Chat](https://whirlwind.chat/) to make that easier. It's a simple app for short, peer-to-peer video chats. You join a group then get matched with others for 2-minute conversations.
+
+This post is to provide a technical overview for a video chat application [Whirlwind Chat](https://whirlwind.chat/) which we've made, as well as give an insight into some of the more interesting parts and the intentions behind them.
+
+![A screenshot showcasing how Whirlwind Chat looks like on mobile](/assets/images/posts/2025-06-23-introducing-whirlwind/mobile-screenshot.png)
+
+## The core: Rust, SvelteKit, and WebRTC
```
marcoow:
```suggestion
## The Core: Rust, SvelteKit, and WebRTC
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+Whirlwind Chat has two parts: a web app written in Svelte, and a backend server written in Rust. The frontend runs entirely in the browser and handles video calls using WebRTC. The backend coordinates users, manages sessions, and helps peers connect.
+
+The actual video and audio data never touch our servers. Everything flows directly between browsers using peer-to-peer connections. This approach improves privacy (no server sits in the middle watching calls) and it makes Whirlwind Chat scalable to a large number of users with minimal infrastructure.
+
+## The backend
```
marcoow:
```suggestion
## The Backend
```

title casing

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+To address this, we created an `ApplicationSupervisor` that spawns and isolates `LobbyServer` structs. In turn, each lobby server spawns its own tasks (such as `matchmaking` and `activity monitor`). If a lobby crashes, only users in that lobby are affected.
+
+Lobby servers don’t run continuously. They are spawned on demand as users join a lobby for the first time, and shut down after a period of inactivity.
+
+Another function of a Supervisor is to provide an access to the internal state of a given lobby. We rely on this mechanism internally for owner actions which are regular HTTP calls instead and not just WebSocket messages. This helps with avoiding re-implementing request/response and authentication mechanisms in a WebSocket connection, ultimately making things simpler by reusing well established HTTP practices.
+
+![Server structure quick overview](/assets/images/posts/2025-06-23-introducing-whirlwind/server-structure.png)
+
+### WebSocket
```
marcoow:
```suggestion
### WebSockets
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+Whirlwind Chat is an application where 90% of the work and functionality happens in the WebSockets. As such it requires a different testing approach that’s more similar to testing Evented systems rather than an HTTP API.
+
+Each test runs a full server instance, bound to a random port with its own database and configuration. This keeps tests isolated, allows them to run in parallel, and supports custom matchmaking configurations.
+
+The test suite follows a black-box approach. Tests use an `Interactions` module, which makes real HTTP requests (not mocks) and connects to the server using `tokio_tungstenite` to simulate a real user session.
+
+The Interactions module stores all received messages for later inspection. This improves reliability, since messages can arrive out of order. For example, a `UserStatus` message might appear while we are waiting for an `IceAnswer`. By collecting all messages, the test can verify outcomes without depending on timing.
+
+## The frontend
```
marcoow:
```suggestion
## The Frontend
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+We used [SvelteKit](https://svelte.dev/docs/kit/introduction) for the frontend. It's a good fit for reactive UIs while keeping bundle size to minimum. (At Mainmatter, [we like Svelte and SvelteKit](/svelte-consulting/) because they strike the right balance between developer productivity and building lightweight, performant web apps.)
+
+The hard part wasn't building the interface, it was making it work reliably across all the different browsers, devices, and hardware users bring. Some users join from phones, others from dual-screen desktops. Microphones and cameras vary. Permission prompts behave differently across OS/browser combinations.
+
+We also had to handle stream negotiation, dynamic device selection, and failure cases where the camera or microphone is missing, is in use elsewhere, or blocked. An optional background blur feature added one more layer of complexity by having to juggle multiple video streams and elements, which is more tricky than it sounds.
+
+![WebRTC overview](/assets/images/posts/2025-06-23-introducing-whirlwind/webrtc-overview.png)
+
+### Background blur with Tensorflow
```
marcoow:
```suggestion
### Background Blur with Tensorflow
```

If we have a screenshot, I think that would come in handy here.

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+To power the background blur feature, we use TensorFlow.js together with BodyPix. BodyPix is a machine learning model that runs entirely in the browser. It performs real-time person segmentation, which means it can identify which parts of the video are the person and which are the background. This makes it possible to blur the background while keeping the speaker in focus.
```
marcoow:
```suggestion
To power the background blur feature, we use [TensorFlow.js](https://www.tensorflow.org/js) together with [BodyPix](https://github.com/tensorflow/tfjs-models/tree/master/body-pix). BodyPix is a machine learning model that runs entirely in the browser. It performs real-time person segmentation, which means it can identify which parts of the video are the person and which are the background. This makes it possible to blur the background while keeping the speaker in focus.
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+While integration was relatively straightforward, applying it to a real-time video call presented a few challenges in the context of WebRTC.
+
+- The person segmentation model is quite large and should ideally be sideloaded rather than bundled.
+- Video processing is pretty heavy on CPU. We needed a way to limit how often segmentation runs using `requestAnimationFrame` and then limit it further by pegging it to the video frame rate by counting how much time has elapsed between draws. The current frequency is `1000ms / 30fps`.
```
marcoow: I don't think I understand what 1 second divided by 30 frames per second means. Is the frequency just 30fps?

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+- It requires swapping out a `video` element with a `canvas` element when blur is toggled on. The video element must be kept in the background and overlayed with `canvas` because it remains the video camera and audio source.
+- Video processing sometimes throws errors that’d typically stop blur from functioning. When that happens, we’re restarting the process.
+
+One of the trickiest problems was deciding which video track to send to the peer. We wanted to avoid adding extra metadata to describe the current stream. WebRTC provides a `replaceTrack` API, but calling it too frequently can cause the connection to stop transmitting video. To avoid that, we debounce the blur toggle (i.e. wait briefly before applying the change) so that track switching only happens once the user has made a final decision.
+
+### Detecting when a video stream stops working
```
marcoow:
```suggestion
### Detecting When a Video Stream Stops Working
```

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,106 @@
...
+
+This was one of our favorite challenges. WebRTC does not provide a simple API to tell whether the connection is working and video is actually being delivered to the peer. You can see your own camera feed just fine, but the person on the other end might not be receiving anything.
+
+Luckily, WebRTC does provide connection statistics through the getStats method on the RTCPeerConnection object. We use this to monitor the video channel and look at the framesReceived count in each report. If the number of frames received stays low for several seconds, we assume the connection is stalled and we call restartIce to force renegotiation between peers. This often fixes problems caused by codec mismatches, connection drops, or switching networks during a call.
```
marcoow:
```suggestion
Luckily, WebRTC does provide connection statistics through the `getStats` method on the `RTCPeerConnection object`. We use this to monitor the video channel and look at the `framesReceived` count in each report. If the number of frames received stays low for several seconds, we assume the connection is stalled and we call restartIce to force renegotiation between peers. This often fixes problems caused by codec mismatches, connection drops, or switching networks during a call.
```

src/posts/2025-06-23-introducing-whirlwind.md
marcoow, line 1: We should have some kind of closing sentences – maybe sth. like

* it seemed easy when we started
* but we encountered a bunch of interesting challenges along the way
* however, solving those we learned a lot

src/posts/2025-06-23-introducing-whirlwind.md
BobrImperator: Nicely spotted, I think the change should be as below however. I mean to say "helps" there originally.
```suggestion
It relies on PostgreSQL as a persistence layer for user records and the Cloudflare Realtime service. The Cloudflare Realtime service provides STUN and TURN servers for WebRTC, which allow devices to discover each other and establish a direct connection. Most connections will work well by utilizing only the STUN server, which essentially helps find a "path" to the other device and once it does, the devices can use that information to connect. This doesn't always work. If one of the devices is behind a firewall or a restrictive NAT, then a direct connection likely won't be possible - and that's when TURN servers come in. TURN servers are relays that transmit data between users.
```

src/posts/2025-06-23-introducing-whirlwind.md
BobrImperator: I'd stay away from such claims. Rust only makes null and error handling explicit, it's still up to the developers whether the program is resilient or not.

src/posts/2025-06-23-introducing-whirlwind.md
BobrImperator: `mpsc` stands for "Multi Producer Single Consumer".

Consumer here is a WebSocket connection that creates it, and if anyone wants to send messages there, the connection must provide the "sending" side of the channel for others to use.

`InMemory` is the module that holds data and implements all the methods. `InMemoryHandle` is "just" a tiny wrapper that allows communicating with `InMemory` through messages.

marcoow: > InMemory is the module that holds data and implements all the methods. InMemoryHandle is "just" a tiny wrapper that allows communicating with InMemory through messages.

I'd add sth. like this to the post

marcoow: > mpsc stands for "Multi Producer Single Consumer".

> Consumer here is a WebSocket connection that creates it, and if anyone wants to send messages there, the connection must provide the "sending" side of the channel for others to use.

ok, sure but I can't imagine the `mailbox` is used for sending (as opposed to receiving) as the text says atm?

src/posts/2025-06-23-introducing-whirlwind.md
BobrImperator: I guess so, yes. We could say "we're running the blur effect at 30fps" instead.

KevinBongart:
```suggestion
- Video processing is pretty heavy on the CPU. We needed a way to throttle segmentation using `requestAnimationFrame`, then we limited it further by skipping updates when the interval between frames was too short. This keeps segmentation close to a frame rate of 30 FPS.
```

src/posts/2025-06-23-introducing-whirlwind.md
BobrImperator: No, `sender` only ever pushes messages to the user who owns a connection, it doesn't do anything else.

src/posts/2025-06-23-introducing-whirlwind.md
BobrImperator: I don't understand the suggestion here, it sounds to me like the only thing that WW does is relaying messages.

src/posts/2025-06-23-introducing-whirlwind.md
BobrImperator: It's "receiving" only for the websocket connection that owns it, "sending" side is for everybody else.

>...the connection must provide the "sending" side of the channel for others to use.

This doesn't sound wrong to me

src/posts/2025-06-23-introducing-whirlwind.md
```diff
@@ -0,0 +1,110 @@
...
+
+Luckily, WebRTC does provide connection statistics through the `getStats` method on the `RTCPeerConnection object`. We use this to monitor the video channel and look at the `framesReceived` count in each report. If the number of frames received stays low for several seconds, we assume the connection is stalled and we call restartIce to force renegotiation between peers. This often fixes problems caused by codec mismatches, connection drops, or switching networks during a call.
+
+## Conclusion
+
+A proof-of-concept we'd built before making Whirlwind Chat made it seem like building the actual thing would be effortless, after all, we'd already figured out how to connect two devices, right? That was the case until we discovered cross-platform and cross-browser issues with things like codecs and device reporting. Even though WebRTC is great, supporting multiple operating systems and browsers can still be challenging.
+
+Things like designing an `ApplicationSupervisor` architecture and managing WebSocket connections with multiple message sources weren't immediately obvious either.
+
+Ultimately, solving all those challenges was great fun and a great learning experience!
```
KevinBongart:
```suggestion
The proof of concept we initially built made it seem like building the full Whirlwind Chat app would be straightforward. After all, we had already figured out how to connect two devices, right? That held up until we ran into cross-platform and cross-browser issues, including codec differences and inconsistent device reporting. Even though WebRTC is great, supporting multiple operating systems and browsers can still be challenging.

Designing the `ApplicationSupervisor` architecture and managing WebSocket connections with multiple message sources also presented challenges that weren't immediately obvious.

Ultimately, solving these challenges was fun and pushed us to solve tough problems across Rust, SvelteKit, and WebRTC. If you're building in any of those areas, [we’d love to help](https://mainmatter.com/contact/). In the meantime, [give Whirlwind Chat a try](https://whirlwind.chat)!
```

src/posts/2025-06-23-introducing-whirlwind.md
KevinBongart:
```suggestion
Whirlwind Chat is an application where most of the functionality takes place in the WebSocket connections. As such it requires a different testing approach that’s more similar to testing evented systems rather than an HTTP API.
```

BobrImperator:
@marcoow all conversations should be resolved now.

The only one would be https://github.com/mainmatter/mainmatter.com/pull/2728/files#r2162319700

marcoow (approved):
It's good but let's replace the second screenshot with one that only shows a video view with a blurred background – ideally not one that appears on the screenshot at the top already. Atm every reader will think it's the same screenshot twice and nobody will understand the backgrounds are blurred on the one at the bottom

PR #2742, opened 2025-06-23
stop duplicating author data for each workshop

mansona:
There is some serious duplication happening on our website 🙈 we have author data that we're not using in a lot of places and we're just duplicating the same information everywhere.

This simplifies things a bit when it comes to titles and bios that show up on the workshops pages, there can be some further unification between title, bio, and longBio fields but that can happen later 👍

### DIFF
--- a/src/components/layouts/workshop.njk
+++ b/src/components/layouts/workshop.njk
@@ -74,7 +74,7 @@ layout: "base"
           <div class="author__details mt-3 author__details--full-width">
             <p>
               {% if lead.bio %}
-                {{ lead.bio }}
+                {{ lead.bio | safe }}
               {% else %}
                 {{ author.data.longBio | safe }}
               {% endif %}
--- a/src/components/layouts/workshop.njk
+++ b/src/components/layouts/workshop.njk
@@ -54,24 +54,31 @@ layout: "base"
         Your mentor{%- if (leads | default([])).length > 1 -%}s{%- endif -%}
       </h2>
       {%- for lead in leads -%}
+
+        {% set author = collections.authors | getAuthor(lead.handle) %}
+
         <div class="author__wrapper">
           <div class="author">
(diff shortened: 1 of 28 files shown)


src/components/layouts/workshop.njk
```diff
@@ -54,24 +54,31 @@ layout: "base"
...
               </div>
 
-              <p class="author__title large">{{ lead.title }}</p>
+              <p class="author__title large">{{ author.data.title }}</p>
             </div>
           </div>
           <div class="author__details mt-3 author__details--full-width">
-            <p>{{ lead.bio | safe }}</p>
+            <p>
+              {% if lead.bio %}
```
pichfl, line 76: Why not safe?
```suggestion
              {% if lead.bio | safe %}
```

also, given that all authors should have a bio, would we ever see the longBio?

pichfl (changes requested):

src/components/layouts/workshop.njk
mansona: so the difference here is that a lead is not an author, the "lead" is the person leading the workshop and it's the data at the bottom of the workshop data.

You can see this working on the effective git workshop: https://github.com/mainmatter/mainmatter.com/blob/0884742c07babcea0cdc7369ba84580ceb4de0f0/src/workshops/effective-git.md?plain=1#L45-L51

Essentially my inline bio takes precedence over my "author" longBio, and since Marco doesn't have one defined it will fall back to his author longBio 👍

mansona: Does that make sense? And you were right about passing the inline bio through safe btw 👍 I committed your suggestion

pichfl: It does help. I'd have expected the opposite (short bio as fallback for long), but that's just juggling semantics. shall approve.

pichfl (approved):

Issue #2768, opened 2025-08-19
Add Auditboard and Skedda logos to the initiative sponsors collage


marcoow:
In the big collage with the current initiative sponsors on the [initiative page](https://mainmatter.com/ember-initiative/), we need to add [Auditboard](https://auditboard.com) and [Skedda](https://www.skedda.com):

<img width="2238" height="1068" alt="Image" src="https://github.com/user-attachments/assets/e755e708-700a-435c-9550-a4471ac46cd8" />

Auditboard needs to be added at the same size as Discourse while Skedda should be small like Productive.

PR #2769, opened 2025-08-19
Improve Ember landing pages

marcoow:
This

* removes the old Embroider & Vite upgrade service which we never sold
* improves the Ember initiative landing page to make it more centered around member benefits


umbrellarocket:
Closes #2768

umbrellarocket (approved):

PR #2770, opened 2025-08-19
feat: "Mock database in Svelte e2e tests with drizzle and playwright fixtures" blog post

paoloricciuti:
[RENDERED VERSION](https://deploy-preview-2770--objective-northcutt-37494c.netlify.app/blog/2025/08/21/mock-database-in-svelte-tests/)


src/posts/2025-08-21-mock-database-in-svelte-tests.md
```diff
@@ -0,0 +1,486 @@
+---
+title: "Mock database in Svelte e2e tests with drizzle and playwright fixtures"
+authorHandle: paoloricciuti
+tags: [svelte]
+customCta: ""
+bio: "Paolo Ricciuti, Senior Software Engineer"
+description: "Writing tests when you are loading your data from a database is very annoying but we can make it better"
```
marcoow:
```suggestion
description: "Writing tests when you are loading your data from a database is very annoying but we can make it better."
```

src/posts/2025-08-21-mock-database-in-svelte-tests.md
```diff
@@ -0,0 +1,486 @@
...
+
+Even with this very simple example we cut the total time in half because we don't need to request additional data: once the page is on the client, that's it.
+
+## The problem
+
+As we hinted before, this approach sounds like the most reasonable...there's a problem however: testability! Doing end-to-end tests when a database is involved is not an easy task and while, to be frank, this mostly boils down to the lack of mocking ability of the database drivers, it's definitely something to keep in mind.
+
+Now one of the solutions could be to simply add an API layer in front of our db and mock that, or we could contain all our db access into a single module and mock that module during testing, but both are not ideal: the first one adds an unnecessary network jump, the second one forces us to structure our code in a certain way and we are one new hire away from messing up that structure (and we would also need to reimplement all the logic in the mocking module).
+
+What we really want is a way to write our code just like we want with the ability to interact with the database from our tests.
```
marcoow: There seems to be something wrong with this sentence

src/posts/2025-08-21-mock-database-in-svelte-tests.md
```diff
@@ -0,0 +1,486 @@
...
+test("home page has the right first user", async ({ page, db, schema }) => {
+  const first_user = await db.select().from(schema.user).limit(1).get();
+  await page.goto("/");
+  await expect(page.locator("li").first()).toHaveText(
+    `${first_user?.id} - ${first_user?.name}`
+  );
+});
+```
+
+If we launch our Playwright tests with UI using `pnpm test:e2e -- --ui` we can see that the test passes and we have 10 users in our db!
```
marcoow, line 368: I don't understand where those users come from – they don't seem to be created anywhere above? Should the above show some kind of seed data file maybe that we're relying on here?

src/posts/2025-08-21-mock-database-in-svelte-tests.md
```diff
@@ -0,0 +1,486 @@
...
+      ],
+    },
+  });
+
+  test("home page has a single user", async ({ page }) => {
+    await page.goto("/");
+    await expect(page.locator("li")).toHaveText(`1 - Paolo Ricciuti`);
+  });
+});
+```
```
marcoow, line 482: IMO this approach should be the only allowed approach. If the data that a test relies on is not part of the test itself but hidden somewhere else (in the worst case some kind of global seed data file), that's going to be a nightmare to maintain over time.

marcoow (commented):
Nice post! The only caveat I see with this approach is that it can't work when tests are executed in parallel as all tests share the same database? Approaches to fix this are:
* run each test in a transaction and rollback the transaction at the end – as long as the test and the server (or the data loading in SvelteKit in this case) use the same connection, they will see the same data and changes "within" the transaction but that data and changes will be invisible "outside" of the transaction.
* give each test its own database – see e.g. https://gerust.rs/docs/architecture/the-web-crate#testing

paoloricciuti:
Oh yeah i meant to add that caveat but forgot...let me fix it

src/posts/2025-08-21-mock-database-in-svelte-tests.md
paoloricciuti: They are seeded with the `seed` function which creates 10 elements by default...it's created with random seeded values so it's consistent between runs.

src/posts/2025-08-21-mock-database-in-svelte-tests.md
paoloricciuti: What specifically?

paoloricciuti:
> give each test its own database

This is something i wanted to explore too but there's an issue with this: playwright launches one sveltekit instance and that's it...so you can't change the database completely...the connection still needs to be the same and you can wipe and re-create the data but you can't have a different db for each test

src/posts/2025-08-21-mock-database-in-svelte-tests.md
paoloricciuti: I think the random seed could be good if you do want some data but don't want to manually specify all of it

src/posts/2025-08-21-mock-database-in-svelte-tests.md
marcoow: ah, I see – would be good to say that somewhere

src/posts/2025-08-21-mock-database-in-svelte-tests.md
marcoow: it's not grammatically correct I think and I'm not sure what you want to say. I assume you want to say sth. like this?

```suggestion
What we really want is a way to write our code naturally while also having the ability to interact with the database from our tests.
```

?

src/posts/2025-08-21-mock-database-in-svelte-tests.md
paoloricciuti: Noted, will update that

marcoow:
Oh, I see. I think you need to add something though that mentions that and says you must make sure tests don't run in parallel because there's only one SvelteKit instance and thus one DB

src/posts/2025-08-21-mock-database-in-svelte-tests.md
paoloricciuti: Oh gotcha...i think there was a missing colon but your suggestion is better anyway

paoloricciuti:
> Oh, I see. I think you need to add something though that mentions that and says you must make sure tests don't run in parallel because there's only one SvelteKit instance and thus one DB

Yeah i had a note about adding it but forgot before pushing 😄

src/posts/2025-08-21-mock-database-in-svelte-tests.md
marcoow: **maybe** for smoke tests like _"it loads when there's data"_ but for anything that makes assertions on particular results or so I'd say you want the data the assertions are based on in the test – see https://mainmatter.com/blog/2017/09/17/magic-test-data/

paoloricciuti: Yeah if you want to assert on the data you need to use something like this...but what if you don't care about the data but just about the fact that there's something there?

marcoow: like I said – smoke tests

paoloricciuti:
@marcoow did some correction

src/posts/2025-08-21-mock-database-in-svelte-tests.md
paoloricciuti: Not necessarily smoke tests: maybe i want to test that clicking on a user opens a dialog...that's not a smoke test but i also don't care about the data.

src/posts/2025-08-21-mock-database-in-svelte-tests.md
```diff
@@ -0,0 +1,515 @@
...
+    // your cleanup code here
+  },
+});
+```
+
+This is already a huge step forward but we can also do better...as we said, we want to seed our db before every test and reset it after...let's see how it works with `drizzle-seed`
+
+#### Putting it all together
+
+Now that we understand how fixtures work we can use the `seed` and `reset` functions from `drizzle-seed`...the `seed` function will generate 10 random (but seeded, so consistent) users in out DB...we can do this before calling `use` to add the users to our database before the test start. After `use` we also invoke `reset` that takes care of wiping our db completely so it will be ready for the next iteration.
```
marcoow:
```suggestion
Now that we understand how fixtures work we can use the `seed` and `reset` functions from `drizzle-seed`...the `seed` function will generate 10 random (but seeded, so consistent) users in our DB...we can do this before calling `use` to add the users to our database before the test start. After `use` we also invoke `reset` that takes care of wiping our db completely so it will be ready for the next iteration.
```

src/posts/2025-08-21-mock-database-in-svelte-tests.md
```diff
@@ -0,0 +1,515 @@
...
+});
+```
+
+Personally I would suggest to default to this strategy if you need to make assertions on the data and use the `seed` function only for the cases where you need "some" data to be there but you don't care about which data is it.
+
+And that's it! All of this can be further improved using the `refine` function of `drizzle-seed` (more documentation on the package [here](https://orm.drizzle.team/docs/seed-overview)) but now that you know the basics the world is your playground!
+
+#### A small caveat
+
+There's a small caveat here that I've kept hidden from you this whole time: given we only have one SvelteKit application running we can only have one db. This means that every test needs to run in series otherwise two tests will update the data of the db at the same time. This is pretty straightforward to do in your `playwright.config.ts` by setting the number of workers to 1
```
marcoow:
```suggestion
There's a small caveat here that I've kept hidden from you this whole time: given we only have one SvelteKit application running we can only have one db. This means that tests cannot run in parallel but have to run in sequence. Otherwise two tests could update the data of the db at the same time. This is pretty straightforward to do in your `playwright.config.ts` by setting the number of workers to 1
```

marcoow (commented):
approving already, just 2 small suggestions left. Also, I'd remove the header – it doesn't add anything and only makes accessing the content harder.

marcoow (approved):

marcoow:
let's also change the date to today please unless Linn asked to release this tomorrow?

paoloricciuti:
> let's also change the date to today please unless Linn asked to release this tomorrow?

I already set the date to today from the beginning

PR #2773, opened 2025-08-20
fix: lowercase bart as an author

paoloricciuti:
Apparently the issue seems to be that @BobrImperator handle was uppercase 🤯


mansona (approved):
🙈

PR #2774, opened 2025-08-21
update(svelte-workshops-cta)

umbrellarocket:


marcoow (commented):
I'm not too sure this is the best way to present this. Can we

* replace the title and text paragraph with just a title (use placeholder for now, I think we can find sth. that's short and precise like "Join our remote workshops in September" or so)
* for each workshop, add the dates and price, and a button that say sth. like "sign up" – I'm thinking sth. like

  ```
  <date>: <title>
  <1 sentence>
  <button: sign up for x€>
  ```
* add the Typescript workshop as well?

umbrellarocket:
@marcoow Thanks, that's a good way of presenting the info. I have made the appropriate changes: https://deploy-preview-2774--objective-northcutt-37494c.netlify.app/blog/2025/06/25/web-components-with-svelte/

marcoow:
let's keep the times in please, otherwise it looks like 2 full days; essentially the idea was the content was final as I had it

marcoow (approved):

PR #2775, opened 2025-08-22
remove autoOg

marcoow:
…as it seems to cause perf problems.

## TODO

- [ ] add `og:image`s for all of the posts that used `autoOg: true`


marcoow:
#2777 is better

PR #2781, opened 2025-09-04
Blog: Useful accessibility tools and browser extensions

pichfl:
Originally part of our travel efforts, but it's useful for anyone.

### DIFF
--- a/src/posts/2025-09-09-accessibility-.md
+++ b/src/posts/2025-09-09-accessibility-.md
@@ -1,0 +1,41 @@
+---
+title: "Useful accessibility tools and browser extensions"
+authorHandle: pichfl
+tags: [a11y, accessibility, frontend]
+bio: "Florian Pichler, Software Developer"
+description: "The EAA is here, use smart tools to act and improve your web applications."
+autoOg: true
+tagline: <p>Add new tools to your toolbelt to validate and improve a11y in your your web application.</p>
+---
+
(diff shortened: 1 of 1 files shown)


src/posts/2025-09-09-accessibility-.md
```diff
@@ -0,0 +1,41 @@
+---
+title: "Useful accessibility tools and browser extensions"
+authorHandle: pichfl
+tags: [a11y, accessibility, frontend]
+bio: "Florian Pichler, Software Developer"
+description: "The EAA is here, use smart tools to act and improve your web applications."
+autoOg: true
+tagline: <p>Add new tools to your toolbelt to validate and improve a11y in your your web application.</p>
```
BlueCutOfficial, line 8: Since the title is quite general and doesn't seem to imply any prerequisites, the tagline could use a link to explain what "a11y" means, or just use "accessibility" for now since you introduce a11y later.

src/posts/2025-09-09-accessibility-.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Extensions
+
+[Google Lighthouse](https://developer.chrome.com/docs/lighthouse?hl=de)
+
+Built directly into Chrome, Lighthouse offers a solid foundation for accessibility investigations. While its scoring system doesn't cover every accessibility concern, it provides valuable direction for achieving fundamental compliance standards.
+
+[axe DevTools](https://www.deque.com/axe/devtools/web-accessibility) - [Chrome](https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd) [Firefox](https://addons.mozilla.org/de/firefox/addon/axe-devtools/)
+
+The full page scan feature gives a good overview and can be used for free without creating an account. The pro features make a big difference and help you detect and analyze more complex problems.
```
BlueCutOfficial, line 31: I let you decide if it's worth it, but it could be nice to add an example of common "more complex problems", especially if there are common ones.

BlueCutOfficial (commented):
Short and efficient. I don't know if that would fit in some sort of conclusion (maybe not) since it's a bit off topic, but for people who are also interested in green coding, there are evaluation tools like Ecograder that also include accessibility fixing instructions in their scoring. Is it something you had the chance to look into?

PR #2782, opened 2025-09-11
chore: show all blog posts when doing `pnpm serve`

paoloricciuti:
This is something that always bother me: when creating a new blog post is not shown locally because it's in the future.

This changes that by only applying the filter when we are in build mode and not in serve.

### DIFF
--- a/collections/posts.js
+++ b/collections/posts.js
@@ -1,7 +1,10 @@
 const pathConfig = require("../src/_data/paths.json");
 const sortByDate = require("../utils/sortByDate");
 const now = new Date();
-const livePosts = post => post.date <= now && !post.data.draft;
+// show all the posts that are not drafts and have a date before now unless we are in serve mode
+// so it will show a blog post you are working on locally
+const livePosts = post =>
+  (post.date <= now && !post.data.draft) || process.env.ELEVENTY_RUN_MODE !== "build";
 
 module.exports = collection => {
   return [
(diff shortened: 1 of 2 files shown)


collections/posts.js
```diff
@@ -1,7 +1,10 @@
 const pathConfig = require("../src/_data/paths.json");
 const sortByDate = require("../utils/sortByDate");
 const now = new Date();
-const livePosts = post => post.date <= now && !post.data.draft;
+// show all the posts that are not drafts and have a date before now unless we are in serve mode
+// so it will show a blog post you are working on locally
+const livePosts = post =>
+  (post.date <= now && !post.data.draft) || process.env.ELEVENTY_RUN_MODE !== "build";
```
pichfl, line 7: Combined Booleans continue to give me headaches. I think I'd prefer it this way:

```suggestion
// show all the posts that are not drafts and have a date before now unless we are in serve mode
// so it will show a blog post you are working on locally
const livePosts = post => 
  !post.data.draft && (post.date <= now || process.env.ELEVENTY_RUN_MODE !== "build");
```

pichfl (changes requested):
Great idea, tiny nitpick in terms of readability of the code.

collections/posts.js
paoloricciuti: But this is a different logic no? A `draft` post will not show up in `serve` mode (which maybe it's what we want but I'd say we can show them all in `serve` mode)

pichfl: This is why that stuff gives me headaches 😄

pichfl (approved):

PR #2786, opened 2025-09-27
add llms.txt

marcoow:
…with company description, list of core technologies, services, clients, cases, and blog posts


paoloricciuti (approved):
LGTM...I also saw Bun is serving the llms.txt even on their normal domain based on the user agent (but maybe that's too much for us)

marcoow:
Sure, that's better but not sure Netlify supports that even. Let's start with this and improve as we go.

PR #2788, opened 2025-10-15
Fix contact form

marcoow:
This changes the contact form so it calls Zapier directly instead of sending an email via the Cloudflare worker which then sends an email to Zapier – that's just an unnecessary indirection

### DIFF
--- a/src/components/contact-form.njk
+++ b/src/components/contact-form.njk
@@ -14,8 +14,8 @@
         id="contact-form"
         data-contact-form
         data-status="initial"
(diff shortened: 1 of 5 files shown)


src/components/contact-form.njk
```diff
@@ -15,7 +15,7 @@
         data-contact-form
         data-status="initial"
         data-action="https://contact.mainmatter.dev/send"
```
pichfl: Should this go to Zapier too?

marcoow: yes

PR #2828, opened 2026-03-03
Homepage redesign

### LINKED ISSUE #2827: Contact list: update the list of services
Yes but we should also change the list of services as it's not really any longer in line with the pages we have:

Image

I'd say we should only list:

* Rust
* Svelte
* Ember
* Training
* Strategic Advice
* Travel

_Originally posted by @marcoow in https://github.com/mainmatter/mainmatter.com/pull/2822#discussion_r2852488710_

umbrellarocket:
This PR adds the redesign homepage. Note: nav bar redesign will be a separate PR.

Closes #2827


umbrellarocket:
Pls review: https://deploy-preview-2828--objective-northcutt-37494c.netlify.app/

src/assets/css/components/_image-list.scss
marcoow, line 1: Does this not exist already?

src/index.njk
```diff
@@ -48,129 +65,68 @@ eleventyImport:
...
+    </p>
+    <p class="medium">
+      We invest heavily in the communities we inhabit by running leading conferences such as
+      <a href="https://eurorust.eu" class="link-black">EuroRust</a>,
+      <a href="https://emberfest.eu" class="link-black">EmberFest</a>, and
+      <a href="https://sveltesummit.com/" class="link-black">Svelte Summit</a>, supporting
+      foundations like the Rust Foundation, contributing to open source, and creating widely used
+      learning resources. We discover and establish new patterns and practices that advance the
+      state of the art, through continuous experimentation and real-world application.
+    </p>
```
marcoow, line 99: Let's make this a bit tighter:


```suggestion
    <p class="medium">
      We focus on a core set of technologies and commit to them fully. For each, we build teams around industry leaders, maintainers, authors, and long-time contributors — people who don't just use the technology, but help shape it.
    </p>
    <p class="medium">
      We invest in the communities we're part of by running conferences like <a href="https://eurorust.eu">EuroRust</a>, <a href="https://emberfest.eu">EmberFest</a>, and <a href="https://www.sveltesummit.com">Svelte Summit</a>, supporting foundations like the Rust Foundation, and contributing to open source and learning resources. Through real-world application and continuous experimentation, we help advance the state of the art.
    </p>
```

marcoow (commented):
are these supposed to have a black background on mobile?

Image

Also would a slightly faded color work better for those logos and a white arrow instead of a teal one so we're mixing fewer colors here (see what we have on the current website)?

Image

src/assets/css/components/_image-list.scss
umbrellarocket: The existing `logo-list.njk `can only handle SVGs and pulls them from `/logos/monochrome`. The logos here are different file formats: a mix of PNG, WEBP, and SVG, and they don't belong in the same folder. That, combined with very different dimensions for the logos in this section, meant that I either had to modify logo-list to handle multiple different configurations or use it as a base for a new element. I chose the latter but happy to switch to former if necessary.

marcoow: Can we not reuse the CSS at least instead of duplicating it?

umbrellarocket:
@marcoow 

> are these supposed to have a black background on mobile?
> 
> Image

I did this on purpose because the hero image is cropped differently on mobile, so a slightly darker section of the background is visible behind the tech boxes.

> Also would a slightly faded color work better for those logos and a white arrow instead of a teal one so we're mixing fewer colors here (see what we have on the current website)?

It becomes a bit too monochrome for my liking:
image

mainmatterlinnbp:
> are these supposed to have a black background on mobile?
> 
> Image
> Also would a slightly faded color work better for those logos and a white arrow instead of a teal one so we're mixing fewer colors here (see what we have on the current website)?
> 
> Image

I agree that a more faded teal color would work well for the logos. Also, the arrows should be white, the same color as the copy next to them. I assume it's not possible to move the arrows closer to the Rust/Ember/Svelte names, or the copy slightly to the center? They're currently very far away.

umbrellarocket:
I pushed an update that addresses your comments, pls check again @marcoow @mainmatterlinnbp 

- removed duplicate CSS from image-list
- changed technology boxes' logo and arrow colors
- changed tech boxes' height

marcoow (approved):

mainmatterlinnbp (approved):

marcoow:
Let me know before this is merged – I need to change some of the sales automation to work with the new services list in the contact form

pichfl:
I think the logos in the big three boxes on the landing page need to be revisited. 

They lack in color contrast against the background which creates a very muddy look. They also need balancing. Rust and Svelte are overpowering. Both in balance against the Ember logo  and the overall Typographic content rhythm.

umbrellarocket:
I strongly advocate that we:

1. Revert to using our established aqua color for the tech logos
2. Replace the ember logo with a square variant. It's one of the technology's approved logo variants. Additionally, it's what we'll be using in #2835 

Edit: new Deploy Preview reflects the suggested changes
Ping @mainmatterlinnbp @marcoow

marcoow:
The logos are way too dominant.

umbrellarocket:
@marcoow There are a few avenues we can pursue to make them less so

- smaller size
- switch to a more neutral color (aka a secondary aqua value)
- do both

Here are some screenshots, pls pick which one works best and we can implement it. FYI, my preference is (1).

**1, smaller size**
image


**2, less contrast through color**
image

**3, both approaches**
image


For context, current look:
image

marcoow:
Can we not use the same kind of semi-transparency we have on the current website? If that's too muddy with teal, maybe we just use white instead?

umbrellarocket:
@marcoow We can, but we do run into a similar issue two transparencies on top of each other being odd. 

Why do we not want these technologies to stand out? I'm thinking they are quite important, so it makes natural sense to place something important and of high contrast right above the fold. This will encourage users to scroll down

image

marcoow:
To me the last screenshot you shared looks pretty good. I'd say the logos are just there to support, the text in the boxes is more important

mainmatterlinnbp:
Looks good! 👍

mainmatterlinnbp (approved):

PR #2830, opened 2026-03-09
Improve run script output

pichfl:
While `npm-run-all2` is a maintained fork of `npm-run-all`, it still has the same issues in dev UX. Turns out pnpm provides this out of the box now. 

I renamed `lint:fix` to `fix` and prefix the `watch:*` related commands to keep the regex simple.

While on it, I noticed that our use of `cross-env` was a bit odd, so I fixed that too and upgraded to the latest version.


marcoow (approved):

pichfl:
@marcoow can you check whats wrong with Percy?

marcoow:
flaky

PR #2834, opened 2026-03-10
Add ember-mirage blog post

nickschot:
This is a blog post version of my ember-europe talk.


marcoow:
https://deploy-preview-2834--objective-northcutt-37494c.netlify.app/blog/2026/03/13/ember-mirage/

src/posts/2026-03-13-ember-mirage.md
```diff
@@ -0,0 +1,195 @@
...
+tagline: "If you've been developing Ember applications with API mocking, you're likely familiar with ember-cli-mirage. ember-cli-mirage has been a corner-stone for writing representative tests that rely on a backend API. It allows you to create focused test scenarios with exactly the data you need for that test. Additionally it can be used to develop frontend features before the API even exists. But there's a problem: ember-cli-mirage doesn't work with Vite! As the Ember ecosystem continues to move toward modern build tooling with Embroider and Vite, we need a solution that brings MirageJS along for the journey. This blog post outlines a path forward to setup MirageJS directly in modern Ember applications."
+autoOg: true
+---
+
+# MirageJS & ember-mirage
+
+MirageJS is the core library. At some point in the past this was extracted from ember-cli-mirage to its own library for framework independent use. Ember-mirage is a set of utilities that brings some of the benefits ember-cli-mirage provided. This means that for basic setup, you may not actually need it, but it provides some features that ember-cli-mirage used to provide that might make migrating a little easier.
```
marcoow:
```suggestion
[MirageJS](https://miragejs.com) is the core library. At some point in the past this was extracted from ember-cli-mirage to its own library for framework independent use. Ember-mirage is a set of utilities that brings some of the benefits ember-cli-mirage provided. This means that for basic setup, you may not actually need it, but it provides some features that ember-cli-mirage used to provide that might make migrating a little easier.
```

src/posts/2026-03-13-ember-mirage.md
```diff
@@ -0,0 +1,195 @@
...
+
+```javascript
+// mirage/servers/default.js
+import { importEmberDataModels } from "ember-mirage/ember-data";
+const emberDataModels = import.meta.glob("../../app/models/**/*");
+
+export async function makeServer(config, _store) {
+  // Don't attempt to import from test-helpers when running in the app
+  let store =
+    _store ??
```
marcoow, line 80: what's `_store`?

src/posts/2026-03-13-ember-mirage.md
marcoow:
```suggestion
MirageJS is the core library. At some point in the past this was extracted from ember-cli-mirage to its own library for framework independent use. [ember-mirage](https://github.com/bgantzler/ember-mirage) is a set of utilities that brings some of the benefits ember-cli-mirage provided. This means that for basic setup, you may not actually need it, but it provides some features that ember-cli-mirage used to provide that might make migrating a little easier.
```

src/posts/2026-03-13-ember-mirage.md
```diff
@@ -0,0 +1,195 @@
...
+
+We pass the ember-data `store` to `makeServer` so that Mirage can generate its models from your ember-data models. We also import a scenarios module, which is simply a function that seeds the Mirage database with development data. Finally, enabling `server.logging` will log all intercepted requests and responses to the browser console, which is useful for debugging.
+
+## Caveats
+
+ember-mirage is still under active development. The setup described in this post is specific to Ember apps using Vite. If you're still on a classic build (without Vite), the setup will differ slightly. In particular, `import.meta.glob` is not available in classic builds. In that case you can use `ember-auto-import`'s `allowAppImports` combined with Webpack's `require.context` as an alternative for auto-importing your Mirage modules.
+
+## Wrapping up
+
+Migrating from ember-cli-mirage to ember-mirage requires a bit more manual setup, but the result is a Vite and Embroider compatible MirageJS configuration that stays close to the native MirageJS experience, with ember-mirage providing utilities to make migration easier for situations where there was a reliance on ember-cli-mirage.
```
marcoow, line 199: You **could** add sth. about how this shows how modern Ember is much more in line with standard JS and requires fewer "Emberisms". Also, being able to use e.g. Vite functionality like `import.meta.glob` is only possible due to the progress that was made in the last 1-2y.

marcoow (commented):
nice!

src/posts/2026-03-13-ember-mirage.md
```diff
@@ -0,0 +1,195 @@
...
+import { createServer } from "miragejs";
+
+export async function makeServer(config) {
+  return createServer({
+    ...config,
+  });
+}
+```
+
+One of the features in which ember-cli-mirage followed ember.js is a lot of files were placed in certain spots by convention. In order to provide a similar setup, we can use the `createConfig` utility provided by ember-mirage to load all our Mirage factories, fixtures, models, serializers and identity managers. We can use Vite's `import.meta.glob` to import all of them at once from their folders. The generated config can then be splatted into the config passed to the `createServer` call.
```
BlueCutOfficial, line 46: It looks like there is a missing piece of explanation regarding what that blog post is about. Is it talking about a fresh ember-mirage setup in a Vite app? or is it talking about a migration path from ember-cli-mirage to unblock Vite? If we try to unblock Vite then we are initially working in a classic build.

src/posts/2026-03-13-ember-mirage.md
```diff
@@ -0,0 +1,195 @@
...
+
+We pass the ember-data `store` to `makeServer` so that Mirage can generate its models from your ember-data models. We also import a scenarios module, which is simply a function that seeds the Mirage database with development data. Finally, enabling `server.logging` will log all intercepted requests and responses to the browser console, which is useful for debugging.
+
+## Caveats
+
+ember-mirage is still under active development. The setup described in this post is specific to Ember apps using Vite. If you're still on a classic build (without Vite), the setup will differ slightly. In particular, `import.meta.glob` is not available in classic builds. In that case you can use `ember-auto-import`'s `allowAppImports` combined with Webpack's `require.context` as an alternative for auto-importing your Mirage modules.
```
BlueCutOfficial: Related to my previous comment, this needs to be stated way sooner. Instead of a post-section "Caveat", that could be a pre-section "Prerequisite" or an additional note in the section introducing `import.meta.glob`. Ideally I think the blog post should have a clear introduction about the scenario, especially if we are not likely to recommend ember-mirage for brand new app that didn't use ember-cli-mirage in the first place.

src/posts/2026-03-13-ember-mirage.md
nickschot: I've expanded a little bit, it's where you'd pass your ember-data store instance when using mirage from the app during development. For tests it's looked up automatically by default.

src/posts/2026-03-13-ember-mirage.md
nickschot: It's both really (as far as people still setup new projects with Mirage I guess), the setup is generally the same and there's not really any reason to move stuff to a different place imo.

edit: the title of course strongly focusses on migrating as that's the most common use case there's gonna be

src/posts/2026-03-13-ember-mirage.md
nickschot: Agreed! I've moved the caveat bit (and rephrased) to a prerequisites section. I wouldn't necessarily not recommend ember-mirage, it's just that you can selectively choose to use certain bits you want/need.

BlueCutOfficial (dismissed):

src/posts/2026-03-13-ember-mirage.md
BlueCutOfficial: > as far as people still setup new projects with Mirage I guess

This is actually the core of the problem IMO. By releasing this blog post as it is now, aren't we encouraging people to setup `ember-mirage` in brand new projects? I used to love working with `ember-cli-mirage`, but the last version of MirageJS was released in 2023, and in our latest assessments we have been explicitly telling teams with no/light `ember-cli-mirage` setup to go with agnostic solutions like [MSW](https://mswjs.io/). I feel like there is a conflict between what we communicate to our clients and what this blogpost is demonstrating.

`ember-mirage` remains a super relevant solution for Ember teams with a heavy `ember-cli-mirage` setup who need to figure out their way to Vite. But this means the _migration_ scenario is more important than the _new setup_ scenario, so this is what the blog post should demonstrate with clear code snippets, and nothing prevent us to say "By the way, if you're on Vite already and you want to go with `ember-mirage`, then `import.meta-glob`..."

> If you’re still on a Webpack/ember-auto-import based setup, you can use Webpack’s [`require.context`](https://webpack.js.org/guides/dependency-management/#requirecontext) combined with [ember-auto-import’s `allowAppImports`](https://github.com/embroider-build/ember-auto-import?tab=readme-ov-file#app-imports) feature instead.

This seems to be what the blog post should show first instead of one sentence that needs to be decoded (potentially hardly by the most junior part of the audience).

---

@nickschot is in vacation this week, 
cc @marcoow I would like to know your thought about this.

marcoow: Good point. The [intro](https://github.com/mainmatter/mainmatter.com/pull/2834/changes#diff-34e29ac530565910f8f4d2226fe7cd0d5e3531f9518699321522ddbf635d5341R7) suggests this is about upgrading scenarios to unblock Vite which I think makes sense. I think it would indeed make sense to not start with

> "This blog post assumes you already have a working Vite setup"

but the migration scenario in which you have a pre-Vite app. Also it might make sense to add something at the end saying that once you're on Vite, you can also use solutions that are backed by a larger community.

The [latest ember-mirage release](https://github.com/bgantzler/ember-mirage/releases/tag/v0.4.3-ember-mirage) was last summer btw – still doesn't invalidate the point that sth. like msw is the safer long-term bet.

mainmatterlinnbp: What can we do about this with @nickschot on vacation? Can someone else make the modifications and publish, or best we wait until Nick is back? He's not back until end of April unfortunately, and would be nice to get this out soon if we can.

marcoow: unless it has to go out now, we should wait for Nick and have him make those changes.

mainmatterlinnbp: We're a bit overdue for a blog post, but I'm hoping we can unblock Marine next week so that she can finalize the Warpdrive post, then it's no problem for this one to wait.

KevinBongart:
Superficial AI review about typos, grammar, vocabulary and punctuation: https://claude.ai/share/d91ebb02-6ec9-4ebf-8e5a-d26f2ee11958

src/posts/2026-03-13-ember-mirage.md
nickschot: So the difficult thing here is that the most straightforward migration scenario requires Vite, meaning you ideally want to do this as an integral part of the migration to Vite. Doing it any other way is unnecessary extra effort due to having to use Webpack. It _is_ possible to use Webpack but it's not great since things have to live in a different spot (from both OG ember-cli-mirage as well as Vite mirage) and would still need to be migrated again when moving to Vite. I can expand on that if we really want to I guess the target audience for that would only be projects that are somehow stuck on old Embroider. I also recall @mansona saying that we want Vite to be the focus/starting point for content nowadays since it's the current reality of Ember (even if projects aren't all there yet).

In the end the migration is mostly a light restructure and some configuration changes to keep things working as they were. I thought it'd be clearer to only explain that structure and keep it concise rather than add code samples for the classic setup and transform all of that into the new setup, but maybe I was wrong about that? In that case this would be the main point that should be clarified?

> The [latest ember-mirage release](https://github.com/bgantzler/ember-mirage/releases/tag/v0.4.3-ember-mirage) was last summer btw – still doesn't invalidate the point that sth. like msw is the safer long-term bet.

I don't think this matters as much as it's just some utility functions that don't really need maintenance. Mirage.JS being unmaintained is potentially more problematic of course (it also uses outdated but functional coding patterns). For a new setup, now that msw/data is no longer EOL/dead-end, that might make more sense but that's a separate story.

BlueCutOfficial: > but maybe I was wrong about that?

I wouldn't phrase it that way. This blog post was initially a written version of your talk, and your talk was good, but I think the situation changed since the conference. Now it's becoming quite clear that Mirage is unmaintained and we have changed our recommendation in the assessments. And that new "state of the art" has impact on what this blog post should communicate IMO -> migration over new setup.

If we align on that, then...

> I also recall @mansona saying that we want Vite to be the focus/starting point for content nowadays

Yes but that particular topic doesn't fit this general rule IMO, precisely because of the migration scenario being more consistent with our communication.

> the most straightforward migration scenario requires Vite, meaning you ideally want to do this as an integral part of the migration to Vite (...) I can expand on that if we really want to

Then yes, I would say this more explicitly. If my understanding is correct, the method relying on Webpack would then be a fallback for people whom Mirage setup is too heavy to include it in the Vite migration, is that correct? If it is, I would also say it more explicitly.

> If you’re still on a Webpack/ember-auto-import based setup, you can use Webpack’s [require.context](https://webpack.js.org/guides/dependency-management/#requirecontext) combined with [ember-auto-import’s allowAppImports](https://github.com/embroider-build/ember-auto-import?tab=readme-ov-file#app-imports) feature instead.

I still think this should have a proper snippet if possible, even if that adds a bit of complexity (could use some UI that highlight it's optional reading maybe).

nickschot: > Then yes, I would say this more explicitly. If my understanding is correct, the method relying on Webpack would then be a fallback for people whom Mirage setup is too heavy to include it in the Vite migration, is that correct? If it is, I would also say it more explicitly.

Why would it be too heavy? The bulk of the mirage files & tests isn't touched, just the configuration. I've had to implement the webpack way before and I'd still say it's an alternative less recommended path where you do have to move **all** of the mirage related files for the full setup which you then also all have to move back when you switch to Vite (not strictly necessary, but does give a cleaner separation).

> I still think this should have a proper snippet if possible, even if that adds a bit of complexity (could use some UI that highlight it's optional reading maybe).

It's not just a snippet, it's: file location (needs to be _within_ the app folder) meaning different paths, different way of including files since there's no glob support, different way of handling exclusion in production and more, which might be too much. If I have some extra time I could write that and see where it goes and we then decide if it's included or not.

---

I think I'll make the adjustment to put a bit more focus that this is mainly meant for migration, explaining that the blog examples are shown as a fresh app since it's more concise and applies to everyone using ember-cli-mirage, but that all aspects (e.g. route files, model files etc.) should be familiar to the audience already meaning it should be straightforward to apply to their (specific) case. In addition I'll use a little more words for the webpack path, but I think describing it in full distracts too much. Could be a follow-up if we feel the need.

BlueCutOfficial:
I had a new look at the introduction and prerequisites, it looks good to me. Just precising this happens after `ember-vite-codemod` + saying explicitly the fresh app is just a matter of simplification, these 2 things bring a lot more clarity IMO 👍

BlueCutOfficial (approved):

mansona (approved):

PR #2835, opened 2026-03-11
Nav bar redesign

umbrellarocket:


marcoow (changes requested):
I don't think this is quite ready yet:

* I can't interact with the contact form
* Should the menu have a bit more space maybe? It feels crammed on a regular sized viewport:

  Image

* The switch between the 2 menu variations when scrolling down doesn't seem to be quite right:

  https://github.com/user-attachments/assets/47403680-1a62-4585-9648-271080fbcefb

* Active state on case studies isn't right (or maybe it's not right on the current site?): https://deploy-preview-2835--objective-northcutt-37494c.netlify.app/cases/anondb/

* I ran into a bug with the expanded submenu:

  Image

* Does this indeed work well visually with all 4 of our colors present at once in the same element?

  Image

umbrellarocket:
@marcoow Thanks! I've arranged your comments by status below:

**I've fixed**
> Should the menu have a bit more space maybe? It feels crammed on a regular sized viewport:

Done, I've rejigged font-sizes to allow for more breathing room.

> Active state on case studies isn't right (or maybe it's not right on the current site?)

You're right, this was missing and is correct nav behaviour. I've added it in.

**Questions/comments**
> I can't interact with the contact form

What do you mean by this?

> Does this indeed work well visually with all 4 of our colors present at once in the same element?

It does not bother me. Technically, we've got two of 'our' colors there: aqua and purple. White and black are defaults that may not register as 'brand colors' at all.


**I will need dev help with**
> The switch between the 2 menu variations when scrolling down doesn't seem to be quite right:

This is due to the nav bar's position transitioning from `:absolute` to `:sticky` on scroll. All solutions i can find are quite hacky, so I'm not sure what the best course of action here is.

> I ran into a bug with the expanded submenu:

Definitely need another set of eyes here

marcoow:
> > I can't interact with the contact form
> What do you mean by this?

on https://deploy-preview-2835--objective-northcutt-37494c.netlify.app/contact/, I can't interact with the form anymore

marcoow:
> > Does this indeed work well visually with all 4 of our colors present at once in the same element?
> It does not bother me. Technically, we've got two of 'our' colors there: aqua and purple. White and black are defaults that may not register as 'brand colors' at all.

I'm not convinced tbh

umbrellarocket:
So, to summarize, I'm going to go ahead and request dev help with ~~three~~ two issues. 

**Issue 1. _Edit: fixed_**
> ~~The switch between the 2 menu variations when scrolling down doesn't seem to be quite right:~~

~~This is due to the nav bar's position transitioning from :absolute to :sticky on scroll. All solutions i can find are quite hacky, so I'm not sure what the best course of action here is.~~

https://github.com/user-attachments/assets/47403680-1a62-4585-9648-271080fbcefb

<br/><br/>
**Issue 2**
> I ran into a bug with the expanded submenu:
_Edit: the submenu positioning no longer works_

  Image
  
  <br/><br/>
 **Issue 3 _Edit:fixed_**
>~~I can't interact with the contact form~~

~~on https://deploy-preview-2835--objective-northcutt-37494c.netlify.app/contact/, I can't interact with the form anymore~~

<br/><br/>

**Issue 4**
Remove unused .js. From where I stand, everything before line `136: // Scrolling event listener for the sticky nav bar. Only appiles class="nav__scrolled" on breakpoints larger than breakpoint-m`
 is not necessary, can you pls double-check and remove if so?

marcoow:
I'm still not really convinced that "Other work" is a good title for the addtl. Menu item. I'd suggest Sth. Like "more" or slightly olayful "what else?" Or so

mainmatterlinnbp:
> I'm still not really convinced that "Other work" is a good title for the addtl. Menu item. I'd suggest Sth. Like "more" or slightly olayful "what else?" Or so

"More" sounds much less compelling to me, that's where I would assume to find things like open positions etc. "Other work" sounds more interesting for people looking to find out more about our work. At the end of the day it depends on what it links out to in the subnav. If we want to keep it more open we can go for "More".

marcoow:
I think what it links to should remain as is

mainmatterlinnbp:
> I think what it links to should remain as is

Looking at it again and considering they playfulness of the brand, I think More works well. It also leaves it open for us to add other topics there. Let's change to that 👍

umbrellarocket:
> > > Does this indeed work well visually with all 4 of our colors present at once in the same element?
> > > It does not bother me. Technically, we've got two of 'our' colors there: aqua and purple. White and black are defaults that may not register as 'brand colors' at all.
> 
> I'm not convinced tbh

@marcoow I roughly mocked up what a sticky menu would look like if we limited it to purple only, leaving only the 'Contact' button in aqua. The transition between the two states isn't perfect yet, but pls see how you feel about the concept: https://deploy-preview-2835--objective-northcutt-37494c.netlify.app/

_Edit:_ note that the 'More' popover menu is currently not fully operational, this will be fixed when positioning it via JS by @pichfl

umbrellarocket:
> > I think what it links to should remain as is
> 
> Looking at it again and considering they playfulness of the brand, I think More works well. It also leaves it open for us to add other topics there. Let's change to that 👍

This is done FYI

marcoow:
I think this works better

mainmatterlinnbp:
Agreed, this works much better 👍

umbrellarocket:
@marcoow Any chance you or @pichfl could take care of #2 and 4 in https://github.com/mainmatter/mainmatter.com/pull/2835#issuecomment-4040190962 so we can get this across the finish line?

marcoow:
Please check with @pichfl – I'm not sure when I'll find time and whether I'll be able to fix this at all

umbrellarocket:
@marcoow Pls give this a final review, it should be good to go.

marcoow (commented):
I think it's good, it just still jumps a bit when starting to scroll. I'm not sure whether that's intentional or not but it looks a bit strange in particular when the hover menu is visible:

https://github.com/user-attachments/assets/f36cea08-074f-458b-a1c4-ffe0f5852e4b

(the above is Safari but it happens in Chrome as well)

umbrellarocket:
> I think it's good, it just still jumps a bit when starting to scroll. I'm not sure whether that's intentional or not but it looks a bit strange in particular when the hover menu is visible:

@marcoow 
I've made two changes that solve the jumping issue
- set the position on desktop to remain `fixed` both on top-of-page and on-scroll
- harmonized the value of margin-top throughout


https://github.com/user-attachments/assets/5ebd8d77-b94f-40b2-aa16-170952a68162

marcoow (approved):

mainmatterlinnbp (approved):

PR #2839, opened 2026-03-17
Cookie banner

umbrellarocket:
TODO:
- [x] Edit our Privacy Policy text @marcoow 
- [x] _Edit: we decided to go with out-of-the-box copy_  ~~Edit the cookie banner copy. @mainmatterlinnbp~~ Screenshots of the current text below: 
image

- [ ] Post launch, confirm that the scripts are firing in Contentsquare (previously Hotjar) and GTM.


marcoow (commented):
We don't show ads so might not need preferences at all?

Image

Also again, styles need to be consistent with the rest of the site – no rounded corners, consistent hover states, etc. If we only show 2 options yes/no, it might also be easier to just roll our own cookie banner instead of using an external JS solution

umbrellarocket:
> We don't show ads so might not need preferences at all?

Google Tag is marketing; Contensquare is analytics. Pls correct me if I'm wrong @mainmatterlinnbp. 

> Also again, styles need to be consistent with the rest of the site – no rounded corners, consistent hover states, etc. 

Great, thanks! 

> If we only show 2 options yes/no, it might also be easier to just roll our own cookie banner instead of using an external JS solution

One thing I like about the current solution is that it's GDPR compliant out of the box.

mainmatterlinnbp:
@umbrellarocket is correct, it's Analytics for Hotjar and Marketing for the Google tag, so both is needed. I like the out-of-the-box solution for legal compliance reasons.

mainmatterlinnbp:
What is the status on this, anything needed before we go live? I'm working on the Google campaigns and would need the tag.

mainmatterlinnbp (approved):

umbrellarocket:
@marcoow Please give this a final once-over. @mainmatterlinnbp has indicated that it's a blocker for her. She would prefer if we merged soon and then dealt with the Privacy Policy changes.

src/assets/js/app.js
```diff
@@ -29,3 +32,64 @@ const logoList = document.getElementById("logo-list");
...
 bindSelectDropdowns();
+
+//Silktide consent manager
+window.silktideConsentManager.init({
+  prompt: {
+    position: "bottomLeft",
+  },
+  text: {
+    prompt: {
+      description: "<p>We use cookies to enhance your experience.</p>",
```
marcoow:
```suggestion
      description: "<p>We use cookies.</p>",
```

marcoow (approved):
just one change

marcoow (changes requested):
hm, actually no – alignment on the buttons is wrong on mobile:

Image

umbrellarocket:
@marcoow I've aligned the text to the left, which creates the following screenshot. Please note that having an arrow follow two lines of text directly is non-trivial to solve. Since only screens narrower than ca. 350 pixels will see this type of layout, I believe it's good enough for our purposes.

image

marcoow (approved):

PR #2850, opened 2026-03-25
Blog post: Redis Query Engine Triemap Port

tall-vase:
Working draft of the Triemap blog post. Putting the last known draft here before I address feedback on Notion.

Preflight checks before merging:

- [x] Diagrams added
- [ ] Author notes
- [ ] og image
- [ ] Final title
- [ ] Rename markdown file & assets folder to reflect date + final title.


tall-vase:
This needs a better way of communicating the speedups too. Percentages are all a bit broad and ">100% speedup" is very vague compared to "0.46x the original operation's speed."

In general though, this is ready for review @JonasKruckenberg

JonasKruckenberg:
I'll give this a look next Tuesday if that is alright with you!

tall-vase:
Works for me! I'll aim to get the diagrams in before then.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
+---
+
+---
+
+Redis wants to migrate the Redis Query Engine from C to Rust, and Mainmatter has been working with them on this project. They've rewritten some of their code to rust already, but this kind of work requires good strategy for both choosing candidates for rewrites and for matching the complex behavior and performance characteristics of the original C code.
```
hdoordt:
```suggestion
Redis wants to migrate the Redis Query Engine from C to Rust, and Mainmatter has been working with them on this project. They've rewritten some of their code to Rust already, but this kind of work requires good strategy both for choosing rewrite candidates and matching the complex behavior and performance characteristics of the original C code.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
...
+    void *data;
+    char label[];
+    //... here come the first letters of each child 
+    //... now come the children pointers
+} TrieMapNode;
+
+#pragma pack()
+```
+
+The core property of this type is that the fields, label, and the array of pointers to children take up a **single heap allocation**. This is really important for cache locality and minimizing pointer dereferences, but it also means the size of the type and some of the offsets of the fields of that type are not known at compile time.
```
hdoordt: I think it'd be good to state explicitly that the actual layout is not fully specified with this struct definition, as this is very much a dynamically sized type that is allocated dynamically on the heap. The comments add some info, but what needs to made more explicit is that:

- the length of `label` is stored in `labelLen`
- the number of "first letters of each child" is as well as the number of "children pointers" is stored in `numChildren`
- Each "first letter of each child" is stored a `char`

This is important as it shows that there is a lot more to it than the type definition shows, and that will need to be handled with care during implementation

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
...
+
+TODO: Diagram of the C memory layout, or pointing to the video.
+
+This is a complex type, and translating it to Rust is difficult. Translating it to safe Rust is impossible, but that doesn't mean we can't translate it _safely_.
```
hdoordt:
```suggestion
This is a complex type, and translating it to Rust is difficult. It cannot even be fully specified in C. Translating it to safe Rust is impossible, but that doesn't mean we can't translate it _safely_.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
...
+#[repr (C)]
+struct ChildRef<T> {
+    first_byte: u8, 
+    node: Node<T>,
+}
+```
+
+This implementation makes no attempt to match the performance characteristics of the C original. We have extra heap allocations for labels and children.
+
+What we get from this baseline is code that works correctly, knowledge of performance characteristics, and a team that has a deeper understanding of the original implementation.
```
hdoordt:
```suggestion
What we get from this baseline is code that works correctly, a well thought out external API, knowledge of performance characteristics, and a team that has a deeper understanding of the original implementation. 
```

hdoordt: Furthermore: this naive implementation allows us to already set up a good testing and benchmarking suite

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
...
+
+There are some differences between our Rust layout and the original C layout:
+
+1. Our optional leaf data sits at the end of the struct, rather than before the label and children.
+
+### DIY Dynamically Sized Types
+
+We want nodes to be a Dynamically Sized Type, like `&str` or `&[u8]`. Something whose size is not known at compile-time.
+
+Our ideal rust implementation would look something like this:
```
hdoordt:
```suggestion
Our ideal Rust implementation would look something like this:
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
...
+    label_len: u16, 
+    n_children: u8, 
+    label: [u8; self.label_len], 
+    children_first_bytes: [u8; self.n_children], 
+    children: [*Node<T>; self.n_children], 
+    data: Option<T>
+}
+```
+
+But we cannot express this in rust's type system. So instead, we split up the implementation into a stack type and a heap type. The stack type `Node<T>` stores the pointer, and the heap type `NodeHeader` stores our data as if it were `IdealNode<T>`.
```
hdoordt:
```suggestion
But we cannot express this in Rust's type system. So instead, we split up the implementation into a stack type and a heap type. The stack type `Node<T>` stores the pointer, and the heap type `NodeHeader` stores our data as if it were `IdealNode<T>`.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
...
+2. The layout is padded, unlike the packed original.
+
+We designed out layout to be as follows:
+
+TODO: Show the Rust Layout
+
+There are some differences between our Rust layout and the original C layout:
+
+1. Our optional leaf data sits at the end of the struct, rather than before the label and children.
+
```
hdoordt: TODO?

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
...
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+pub(super) struct NodeHeader {
+    pub label_len: u16,
+    pub n_children: u8,
+    // ...where are the other fields?
+}
+```
+
+Note that in `NodeHeader` we can't express everything we want to in the type system. Instead, we avoid defining them directly and make sure to manually allocate them ourselves.
```
hdoordt: Just like in the C implementation

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Allocating a Dynamically Sized Type
+
+The `alloc` function has the signature:
```
hdoordt:
```suggestion
The `std::alloc::alloc` function has the signature:
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
...
+
+## Porting Strategy: Managing Necessary Unsafe
+
+Once we have our `Layout`, we can call `alloc` to get our pointer. This memory is uninitialized, but we can initialize it manually. Managing this memory manually only gets gnarly once we want to split or merge nodes.
+
+In the final codebase we were finally left with **128** unsafe blocks and **21** unsafe methods.
+
+Managing this unsafe code requires care and understanding for how developers will interact with it. Preferably a developer won't interact with it at all unless they're a maintainer. 
+
+On top of all this, the client is moving to rust to minimize issues related to memory safety so it is important to build an API that can be trusted and can't be misused.
```
hdoordt:
```suggestion
On top of all this, the client is moving to Rust to minimize issues related to memory safety so it is important to build an API that can be trusted and can't be misused.
```

hdoordt: There's a couple more of these below. I'll stop calling them out explicitly to avoid cluttering

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
...
+
+To deal with these constraints, we have a core axiom of **no `unsafe` in the public API**, on top of trying to rely on tooling as much as possible.
+
+### Tooling & Patterns
+
+Our management of the unsafe code is a mix of good documentation practices for critical areas & automated tooling. We can't fully automate this process, otherwise it wouldn't be `unsafe`, but we can do our best to build and maintain certainty.
```
hdoordt:
```suggestion
Our management of the unsafe code is a mix of good documentation practices for critical areas & automated tooling. We can't fully automate this process, which is why we have to resort to `unsafe`, but we can do our best to build and maintain certainty.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
...
+
+1. **Miri**
+
+Miri runs your rust code in an interpreter against different memory models, checking for undefined behavior as it comes up. A great help when we're writing lots of unsafe code. 
+
+This only works on Rust code, so we can't apply this in other parts of Rust Redis where we cross FFI boundaries into C libraries, but we can use it here and anywhere else where all dependencies are rust.
+
+2. **Debug assertions**
+
+Debug assertions are checks that only run in debug builds, release builds do not pay the performance cost of them. Debug assertions let us check the invariants we want to build our API around at runtime and give good errors when those assertions fail.
```
hdoordt: They also further specify the invariants in code

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
...
+
+3. **Clippy Lints**
+
+Following [Jack Wrenn's guidance](https://jack.wrenn.fyi/blog/safety-hygiene/) we're using clippy lints to enforce two soft invariants:
+
+- 3.1: `undocumented_unsafe_blocks`: If an unsafe block does not have comments surrounding it, there will be a warning.
+
+- 3.2: `multiple_unsafe_ops_per_block`: If an unsafe block contains more than one unsafe operation, there will be a warning.
+
+This pushes developers in the direction of properly documenting the unsafe code they do right, operation by operation, to assure all invariants are documented.
```
hdoordt: And of course, we're denying warnings in CI

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,273 @@
...
+|Unsafe expressions|-|128|
+|Coverage|82.8%|95.6%|
+|Microbenchmarks|❌|✅|
+|Performance|(baseline)|25% to 100% improvement in core operations|
+
+The codebase did double in size, but with a move to a Rust codebase the unsafe operations have been confined to an area of the code a fraction of the size of the C codebase, where unsafe operations could have been happening anywhere.
+
+Test coverage was greatly improved, the last percentage points in coverage are in unreachable areas.
+
+Performance was also greatly improved, in part from changing the layout from Packed to Unpacked.
```
hdoordt:
```suggestion
Performance was also greatly improved, in part from moving away from the packed layout.
```

src/posts/2026-04-XX-redis-triemap.md
tall-vase: Good catch, may even bring it into a short paragraph.

src/posts/2026-04-XX-redis-triemap.md
tall-vase: I don't quite follow, could you elaborate?

tall-vase:
We need to title this one before I can sort an OG image for it. It's based on the "Rewrite, Optimize, Repeat: Our journey porting a triemap from C to Rust" talk, and I don't think an identical title would be appropriate but riffing off it makes sense. Looking for opinions.

mainmatterlinnbp:
@tall-vase can you please confirm that @LukeMathWalker's feedback from Notion has been resolved here? If so, please let us know. Thanks!

tall-vase:
The feedback I got from Luca was that the discussion of picking a good candidate from a complex module graph didn't flow well into the TrieMap details, as well as that this should be talking about the Redis Query Engine not Redis. I can confirm I've addressed this feedback, though a second set of eyes on that is always appreciated.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
+---
+
+---
+
+At Mainmatter we've been working to help Redis migrate the Redis Query Engine from C to Rust. They've ported some of their existing query engine code already, but this kind of work requires good strategy in both choosing good rewrite candidates and matching the complex behavior and performance characteristics of the original C code.
```
LukeMathWalker: This is not correct—they had not ported any of their Redis Query Engine code over to Rust. They had ported _other_ Redis modules (not C modules) over to Rust (namely, RediJSON).

Unless stated otherwise, I'd assume that the reader reads rewrite and interprets it as "big bang rewrite"/"new codebase".
We should also clarify here, front and centre, that we're doing an _incremental_ port, otherwise it's unclear why we need to choose good rewrite candidates.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+
+This blog post covers 
+
+## Case Study: Porting the TrieMap
```
LukeMathWalker: Re-iterating my feedback from Notion: how are we linking the intro to this section?

It's already much more focused, which I appreciate, but it doesn't quite flow "right" yet.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+
+A TrieMap is a key-value data structure that has a similar API as a Hashmap / Binary Tree Map, but optimized for compressing keys by their shared prefixes.
+
+![Diagram showing how a set of tuples are represented in a TrieMap.](/assets/images/posts/2026-04-XX-redis-triemap/trie.svg)
+
+We can easily imagine how to do this in Rust, we just need to store a data structure in the form:
```
LukeMathWalker:
```suggestion
We can sketch this out in Rust:
```

As a style guideline, never tell the reader that they can _easily_ do things or that they _just_ need to do X/Y/Z. We don't know how familiar they are with this kind of work and we don't want them to feel stupid if this doesn't seem/feel easy to do/imagine.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+struct TrieMapNode<T> {
+    label: Vec<u8>,
+    children: Vec<TrieMapNode<T>,
+    // The data that the label "maps" to.
+    // This is optional, because not all nodes have complete keys with associated data.
+    data: Option<T>,
+}
+```
+
+Very simple stuff! We can "cascade" down nodes from the root to complete the labels of leaves. The data field is optional as labels can branch without having an appropriate piece of data at the branch point: a map with only the key-value pairs `"Scarborough": 42` and `"Scaffolding": 451` doesn't have associated data for their shared parent `"Sca"`.
```
LukeMathWalker:
```suggestion
We can "cascade" down nodes from the root to complete the labels of leaves. The data field is optional as labels can branch without having an appropriate piece of data at the branch point: a map with only the key-value pairs `"Scarborough": 42` and `"Scaffolding": 451` doesn't have associated data for their shared parent `"Sca"`.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+
+At Mainmatter we've been working to help Redis migrate the Redis Query Engine from C to Rust. They've ported some of their existing query engine code already, but this kind of work requires good strategy in both choosing good rewrite candidates and matching the complex behavior and performance characteristics of the original C code.
+
+This blog post covers 
+
+## Case Study: Porting the TrieMap
+
+A TrieMap is a key-value data structure that has a similar API as a Hashmap / Binary Tree Map, but optimized for compressing keys by their shared prefixes.
+
+![Diagram showing how a set of tuples are represented in a TrieMap.](/assets/images/posts/2026-04-XX-redis-triemap/trie.svg)
```
LukeMathWalker, line 13: I believe we're transitioning too fast from "here's the data structure" to "let's discuss the nitty-gritty details of the implementation".
I'd rather have a dedicated section showing how the data structure grows/reshapes itself after each insertion, thus helping the reader build out their intuition as to how a TrieMap works. 
We can include a little disclaimer right under the heading that says something along the lines of "This section explains how a `TrieMap` works. If you're already familiar with this data structure, skip ahead to the <next section>".

My assumption is that few folks know (or remember) how a trie works by heart. I certainly didn't.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+    //... here come the first letters of each child 
+    //... now come the children pointers
+} TrieMapNode;
+
+#pragma pack()
+```
+
+The core property of this type is that the fields, label, and the array of pointers to children take up a **single heap allocation**. This is really important for cache locality and minimizing pointer dereferences, but it also means the size of the type and some of the offsets of the fields of that type are not known at compile time.
+
+Note how this type is not fully definable in C's type system, the two lines of comments here are fields that we can't define in a C struct definition and need to compute how to access at runtime. `labelLen` and `numChildren` let us do these computations.
```
LukeMathWalker: A small snippet of C code showing how C computes the offset of one of the dynamically-located fields would be helpful here.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+```rust
+pub unsafe fn alloc(layout: Layout) -> *mut u8
+```
+
+Rather than C's `malloc` which takes a length in bytes, `alloc` takes a `Layout` type that represents the size and alignment of an allocation.
+
+`Layout` exposes a fairly simple API, and we can define a layout to suit our complex needs at runtime.
+
+1. `Layout::new` produces a layout for a sized type.
+2. `Layout::array` produces a layout for an array of a sized type, with a runtime-or-const `n` elements.
```
LukeMathWalker:
```suggestion
2. `Layout::array` produces a layout for an array of a sized type, with `n` elements. 
```

I don't think the distinction of runtime vs const time evaluation matters at this point in the article.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+
+The `std::alloc::alloc` function has the signature:
+
+```rust
+pub unsafe fn alloc(layout: Layout) -> *mut u8
+```
+
+Rather than C's `malloc` which takes a length in bytes, `alloc` takes a `Layout` type that represents the size and alignment of an allocation.
+
+`Layout` exposes a fairly simple API, and we can define a layout to suit our complex needs at runtime.
```
LukeMathWalker: We should point out to the reader that `Layout`-based computations yield the same exact layout that the compiler would use for a struct with those fields annotated with `#[repr(C)]`, to ground their intuition.
As in, order matters too, for padding.

LukeMathWalker: A snippet of code doing the layout computation would probably help too.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+
+This composable API of `Layout` means it's fairly easy to build the `Layout` we need for our custom Dynamically Sized Type.
+
+## Porting Strategy: Managing Necessary Unsafe
+
+Once we have our `Layout`, we can call `alloc` to get our pointer. This memory is uninitialized, but we can initialize it manually. Managing this memory manually only gets gnarly once we want to split or merge nodes.
+
+In the final codebase we were finally left with **128** unsafe blocks and **21** unsafe methods.
+
+Managing this unsafe code requires care and understanding for how developers will interact with it. Preferably a developer won't interact with it at all unless they're a maintainer.
```
LukeMathWalker: The developer/maintainer distinction doesn't apply here. The project isn't really "open contributions", so all developers working on it are maintainers. 
It's more about "consuming this unit of code in another module" vs "changing how this unit of code works".

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+
+Once we have our `Layout`, we can call `alloc` to get our pointer. This memory is uninitialized, but we can initialize it manually. Managing this memory manually only gets gnarly once we want to split or merge nodes.
+
+In the final codebase we were finally left with **128** unsafe blocks and **21** unsafe methods.
+
+Managing this unsafe code requires care and understanding for how developers will interact with it. Preferably a developer won't interact with it at all unless they're a maintainer. 
+
+On top of all this, the client is moving to Rust to minimize issues related to memory safety so it is important to build an API that can be trusted and can't be misused.
+
+To deal with these constraints, we have a core axiom of **no `unsafe` in the public API**, on top of trying to rely on tooling as much as possible.
```
LukeMathWalker: *When possible. 
We have plenty of `unsafe` in public APIs elsewhere, if needed/proper. Let's avoid absolutes.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Tooling & Patterns
+
+Our management of the unsafe code is a mix of good documentation practices for critical areas & automated tooling. We can't fully automate this process, which is why we have to resort to `unsafe`, but we can do our best to build and maintain certainty.
```
LukeMathWalker: I'm not sure I follow what "We can't fully automate this process, which is why we have to resort to `unsafe`" means/implies.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+
+1. **Miri**
+
+Miri runs your rust code in an interpreter against different memory models, checking for undefined behavior as it comes up. A great help when we're writing lots of unsafe code.
+
+This only works on Rust code, so we can't apply this in other parts of Rust Redis where we cross FFI boundaries into C libraries, but we can use it here and anywhere else where all dependencies are rust.
```
LukeMathWalker: Rust Redis?

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+
+2. **Debug assertions**
+
+Debug assertions are checks that only run in debug builds, release builds do not pay the performance cost of them. Debug assertions let us both specify and check the invariants we want to build our API around at runtime, and give good errors when those assertions fail.
```
LukeMathWalker: It's not about "release builds" vs "debug builds". It's more about "release artefacts"—i.e. they will be stripped out in the shared library we ship to users.
But we do use them in optimised builds when running tests, see https://github.com/RediSearch/RediSearch/blob/c14fa3cb10c4b56779349f62635fe9444e1d0c55/src/redisearch_rs/Cargo.toml#L213

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+
+Heavy use of debug assertions ties nicely with our use of Miri, whose errors can be overly esoteric or academic. Rust assertions are much easier to contextualize.
+
+3. **Clippy Lints**
+
+Following [Jack Wrenn's guidance](https://jack.wrenn.fyi/blog/safety-hygiene/) we're using clippy lints to enforce two soft invariants:
```
LukeMathWalker: What is a "soft" invariant?

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+
+- 3.1: `undocumented_unsafe_blocks`: If an unsafe block does not have comments surrounding it, there will be a warning.
+
+- 3.2: `multiple_unsafe_ops_per_block`: If an unsafe block contains more than one unsafe operation, there will be a warning.
```
LukeMathWalker: I'd flip this to their positive counterpart, to make them easier to understand.
E.g. 
```
- 3.1: `undocumented_unsafe_blocks`: every `unsafe` block must have a dedicated `// SAFETY` comment attached to it.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+
+This pushes developers in the direction of properly documenting the unsafe code they do write, operation by operation, to assure all invariants are documented.
+
+### Safety Comments
```
LukeMathWalker, line 271: We should weave a link to https://github.com/rust-lang/rfcs/pull/3842 in this section as the "ideal" way of managing this via tooling.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,275 @@
...
+
+At this layer we're developing methods and functions that perform unsafe operations but have API surfaces which are much more specialized for the tasks we're using them for. This is where we try to keep most of the internal API of the TrieMap.  
+
+**Top**: The Safe API
+
+Finally, we have the safe API surrounding the data structure that we want end users to actually use. This doesn't expose any `unsafe`, as per our requirements.
+
+## Conclusion
+
+This implementation is now in production in the Redis Query Engine! We can note & measure some key changes in the codebase from this transition.
```
LukeMathWalker:
```suggestion
This implementation is now in production in Redis Query Engine! We can note and measure some key changes in the codebase from this transition.
```

LukeMathWalker:
I've done a first review pass, mostly focused on structure and flow, with only some minor style comments here and there.

src/posts/2026-04-XX-redis-triemap.md
hdoordt: We're saying

> If an unsafe block contains more than one unsafe operation, there will be a warning.

Which is great and all, but warnings can be ignored. That's why we configured CI such that it will fail if there are any warnings

src/posts/2026-04-XX-redis-triemap.md
tall-vase: I'm using it here as "invariants outside of rust's own formal tools" but given we're still applying an analytical tool to prevent this you're right to point out there's not really a difference in this case.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,352 @@
+---
+
+---
+
+At Mainmatter we've been working to help Redis migrate the Redis Query Engine from C to Rust. This kind of work requires good strategy in both choosing good rewrite candidates and matching the complex behavior and performance characteristics of the original C code.
+
+This blog post covers the process of porting the Redis Query Engine's TrieMap implementation to Rust, including the challenges faced with making unsafe code maintainable and designing our own Dynamically Sized Type.
```
LukeMathWalker:
```suggestion
Mainmatter has been working with Redis to migrate Redis Query Engine from C to Rust, module by module. Matching the performance characteristics of the original C code is a non-negotiable requirement: no one likes a slower database, even if it's written in Rust.

This blog post focuses on Redis Query Engine's TrieMap, one of the first C modules we ported over to Rust. In particular, we go over the implementation challenges of designing a custom Dynamically Sized Type (DST) and the maintenance approach we picked for the unsafe code required by a custom DST.
```

Let's mention the key things (incrementality, the importance of perf) without stressing aspects we won't dive into (module picking strategy).

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,352 @@
...
+
+## What Is a TrieMap
+
+A TrieMap is a key-value data structure that has a similar API to a Hashmap / Binary Tree Map, but optimized for compressing keys by their shared prefixes.
```
LukeMathWalker:
```suggestion
A TrieMap is a key-value data structure with an API similar to that of a Hashmap. Keys must be sequence-like types (in our case, byte slices); values can be arbitrary types. TrieMap leverages shared key prefixes to reduce its memory footprint, de facto compressing the key set.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,352 @@
...
+
+![Diagram showing how a set of tuples are represented in a TrieMap.](/assets/images/posts/2026-04-XX-redis-triemap/trie.svg)
+
+In the above diagram, taken from [the talk this blog post is based on](https://www.youtube.com/watch?v=XklFGy3aUX4), we can see that each node has:
+
+1. A prefix label that this node and all its children share.
+2. Optional data (emoji, here).
+3. Child nodes (each with an associated first character.).
+
+We can "cascade" down nodes from the root to complete the labels of leaves. The data field is optional as labels can branch without having an appropriate piece of data at the branch point: a map with only the key-value pairs `"Scarborough": 42` and `"Scaffolding": 451` doesn't have associated data for their shared parent `"Sca"`.
```
LukeMathWalker: Let's use a practical example based on the diagram. E.g.

```
Each key stored in the TrieMap can be recovered by walking down from the root node. For example, <Key X> is stored as <Segment A> + <Segment B> + <Segment C>.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,352 @@
...
+
+We hold onto the first byte of each child's label because, by nature of how shared prefixes are always consolidated by the ancestors of a node, each child is guaranteed to have a unique first byte. This can be leveraged for faster searches within the data structure.
```
LukeMathWalker:
```suggestion
We hold onto the first byte of each child’s label as a performance optimisation: when performing searches (e.g. find all the keys with a given prefix), we can determine which children to visit with minimal pointer chasing.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,352 @@
...
+    // This is optional, because not all nodes have complete keys with associated data.
+    data: Option<T>,
+}
+```
+
+This does what we need it to, at least structurally, so maybe we can stop here? End of the blog post everyone, we can all can go home! Not so fast. A trained eye can see how this formulation of a TrieMap can cause problems. Let's take a step into the real world, where this TrieMap doesn't meet the performance requirements of the Redis Query Engine.
+
+## Porting Redis Query Engine's TrieMap
+
+Redis Query Engine has its own C implementation of a TrieMap designed around the performance characteristics required for the real world. Let's take a look at it:
```
LukeMathWalker:
```suggestion
Let’s take a look at the node layout of the C implementation we’re trying to replace:
```

"real world" is a bit weak and vague. We can just omit it.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,352 @@
...
+    void *data;
+    char label[];
+    //... here come the first letters of each child 
+    //... now come the children pointers
+} TrieMapNode;
+
+#pragma pack()
+```
+
+The core property of this type is that the fields, label, and the array of pointers to children take up a **single heap allocation**. This is really important for cache locality and minimizing pointer dereferences, but it also means the size of the type and some of the offsets of the fields of that type are not known at compile time.
```
LukeMathWalker:
```suggestion
All node components are laid out next to each other, in a single heap allocation. There is no further indirection. This helps with cache locality and minimizes the number of pointer dereferences, but it also implies that the size of the type and the offsets of some of its fields are not known at compile time.
```

tall-vase: I would like to hold onto the 'property' framing, or at least something that points out how peculiar this is for a data structure design. But I'll commit this.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,352 @@
...
+We can use this to manually build a runtime-defined layout that suits our Dynamically Sized Type's needs. If we look at our DST's layout needs again:
+
+1. `label_len` (static size, `u16`)
+2. `n_children` (static size, `u8`)
+3. `label` (non-static size, `label_len` * `u8`)
+4. `children_first_bytes` (non-static size, `n_children` * `u8`)
+5. `children` (non-static size, `n_children` * `Node<T>`)
+6. `data` (static size, `Option<T>`)
+
+We have 3 non-static fields and 3 static fields. So to build up our layout we can do the following:
```
LukeMathWalker: `static` is a very charged word in Rust and it may be misleading here. Can we find another way to distinguish?
E.g. `size known at compile-time` vs `dynamic size`

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,352 @@
...
+
+Note how long this safety comment is, with 128 unsafe blocks in the codebase this could become an overwhelming amount of information very quickly. This is where the next component of our unsafe management strategy comes in.
+
+### Abstractions That Are Hard to Misuse
+
+Unsafe doesn't let users or maintainers do "anything", but it does let them get away with misusing the tools they're given in ways that make software systems unpredictable.
+
+To strengthen systems built on unsafe foundations, we start by **Layering Abstractions** in a way that gradually reduces the concerns of a maintainer.
+
+The end goal is to have a outwardly safe API, and a internal implementation that restricts what mistakes a maintainer to specific areas of the code. Core, most-unsafe operations are built upon by still-unsafe abstractions that limit the operations the maintainer can perform, and those abstractions are built upon by more-safe abstractions.
```
LukeMathWalker: This doesn't read very well. Can we find a way to make it flow better?

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,352 @@
...
+
+![List of microbenchmark results, showing an across the board improvement in execution times. Lowest increase is the loading function, at 0.88 original speed. Largest increase is leaf insertion, at 0.33 original speed. Most results sit around 0.45-0.75 times original speed.](/assets/images/posts/2026-04-XX-redis-triemap/speedup.svg)
+
+</details>
+
+One thing that stands out is how the codebase doubled in size. Moving to a rust codebase sometimes comes with the expectation of a more expressive, smaller codebase. With the move to Rust the unsafe operations have been confined to an area of the code a fraction of the size of the original C codebase, where unsafe operations could have been happening anywhere. The abstractions built to confine the unsafe operations mean there has been a bulking out of the module in terms of lines of code, but the payoff is that only ~128 of those lines need to be kept under extra scrutiny. 
+
+Test coverage was greatly improved, the last percentage points in coverage are in unreachable areas.
+
+Performance was also greatly improved, in part by moving away from the packed layout.
```
LukeMathWalker: This is too strong. Let's limit ourselves at "Perfomance improved as well, ...". 
We got numbers and we should redirect to those without too many qualifiers.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,352 @@
...
+
+What we get from a rewrite is more than just a codebase translated to a new language. A rewrite is an opportunity for an engineering team to relearn and re-contextualize the core problems that existing code exists to solve, to learn what's important about the building blocks of the software they're developing.
+
+A rewrite is also an opportunity to reassess assumptions. In the original implementation we had a packed layout, we questioned that assumption and went with a different direction, as well as making other small changes in the memory layout of the data structure. Challenging these assumptions paid off, as our new implementation is faster and has a negligible increase in memory use.
+
+We could have stopped at our naive implementation, failing the client on performance. We could have wrapped large areas of the codebase in unsafe, failing the client on maintainability. We could have come in with a new data structure entirely, failing the client on the performance assumptions made in the rest of their codebase.
```
LukeMathWalker:
```suggestion
We could have stopped at our naive implementation, failing the client on performance. We could have wrapped large areas of the codebase in unsafe, failing the client on maintainability. 
```

A new data structure would have been perfectly OK had we had one that satisfied the required boxes, so let's drop that entirely.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,352 @@
...
+} TrieMapNode;
+
+#pragma pack()
+```
+
+All node components are laid out next to each other, in a single heap allocation. There is no further indirection. This helps with cache locality and minimizes the number of pointer dereferences, but it also implies that the size of the type and the offsets of some of its fields are not known at compile time.
+
+Note how this type is not fully definable in C's type system, the two lines of comments after `label` are fields that we can't define in a C struct definition and need to compute how to access at runtime. `labelLen` and `numChildren` let us do these computations.
+
+```c
```
tall-vase, line 65: This needs closer scrutiny (@LukeMathWalker ? @hdoordt ?) as my attempts at getting an MVP working properly mostly ended in segfaults.

src/posts/2026-04-XX-redis-triemap.md
LukeMathWalker: I wouldn't say it's peculiar, it's fairly common optimization, but it's rather unwieldy to achieve it in this specific instances due to the dynamic nature of some of the data.

src/posts/2026-04-XX-redis-triemap.md
tall-vase: The nomenclature for "known at compile time" is "static" and "size known at compile-time" is very unwieldy. The lifetime/keyword connotations could be avoided by clarifying immediately that we don't mean those instead, but I'll think about alternatives.

LukeMathWalker: > The nomenclature for "known at compile time" is "static" 

What's the source for that? AFAIK, it would be `Sized` (as in https://doc.rust-lang.org/std/marker/trait.Sized.html), but that doesn't make it clearer to folks who haven't had to work with these issues before.

src/posts/2026-04-XX-redis-triemap.md
LukeMathWalker: This wasn't resolved @tall-vase.

src/posts/2026-04-XX-redis-triemap.md
LukeMathWalker: It is tricky to do it right, you need to account for the packed layout too and the fact that the children pointers may not be aligned.
Easier to quote the original code here: https://github.com/RediSearch/RediSearch/blob/dcf009a2327240b24d6efcf100fed577b8a5eef0/deps/triemap/triemap.c#L18

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,352 @@
...
+
+In the final codebase we were finally left with **128** unsafe blocks and **21** unsafe methods.
+
+Managing this unsafe code requires care and understanding for how maintainers will interact with it. On top of all this, the client is moving to Rust to minimize issues related to memory safety so it is important to build an API that can be trusted and can't be misused.
+
+To deal with these constraints, we have a normative value of **as little `unsafe` in the public API as possible**, on top of trying to rely on automated tooling as much as possible.
+
+### Tooling & Patterns
+
+Our management of the unsafe code is a mix of good documentation practices for critical areas & automated tooling. We can't fully automate the unsafe code management process, if we could then it wouldn't be unsafe, but we can do our best to build and maintain certainty through what tools and practices we can bring together.
```
LukeMathWalker:
```suggestion
Our management of the unsafe code is a mix of good documentation practices for critical areas and automated tooling. We can't fully automate the unsafe code management process, if we could then it wouldn't be unsafe, but we can do our best to build and maintain certainty through what tools and practices we can bring together.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,352 @@
...
+
+Heavy use of debug assertions ties nicely with our use of Miri, whose errors can be overly esoteric or academic. Rust assertions are much easier to contextualize.
+
+3. **Clippy Lints**
+
+Following [Jack Wrenn's guidance](https://jack.wrenn.fyi/blog/safety-hygiene/) we're using clippy lints to enforce two invariants:
+
+- 3.1: `undocumented_unsafe_blocks`: every `unsafe` block must have a dedicated `// SAFETY` comment attached to it.
+
+- 3.2: `multiple_unsafe_ops_per_block`: If an `unsafe` block contains more than one `unsafe` operation, there will be a warning.
```
LukeMathWalker: Same as the other bullet point, let's flip it to the "normative" version—i.e. every `unsafe` block must contain a single `unsafe` operation.

src/posts/2026-04-XX-redis-triemap.md
tall-vase: Aha, this was eaten by a merge resolve on my end.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+
+In the above diagram, taken from [the talk this blog post is based on](https://www.youtube.com/watch?v=XklFGy3aUX4), we can see that each node has:
+
+1. A prefix label that this node and all its children share.
+2. Optional data (emoji, here).
+3. Child nodes (each with an associated first character.).
+
+We can "cascade" down nodes from the root to recover the labels of leaves, as in the above diagram we recover `"bike": 🚲` from the empty root, the "bi" node, and the "ke" node. 
+
+The data field for each node is optional as labels can branch without having an appropriate piece of data at the branch point: a map with only the key-value pairs `"Scarborough": 42` and `"Scaffolding": 451` doesn't have associated data for their shared parent `"Sca"`.
```
LukeMathWalker:
```suggestion
The data field for each node is optional as labels can branch without having an associated payload at the branch point: a map with only the key-value pairs `"Scarborough": 42` and `"Scaffolding": 451` doesn't have associated data for their shared parent `"Sca"`.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+    label: Vec<u8>,
+    children_first_bytes: Vec<u8>,
+    children: Vec<TrieMapNode<T>>,
+    // The data that the label "maps" to.
+    // This is optional, because not all nodes have complete keys with associated data.
+    data: Option<T>,
+}
+```
+
+This does what we need it to, at least structurally, so maybe we can stop here? End of the blog post everyone, we can all can go home! Not so fast. A trained eye can see how this formulation of a TrieMap can cause problems. Let's take a step into the real world, where this TrieMap doesn't meet the performance requirements of the Redis Query Engine.
```
LukeMathWalker:
```suggestion
This does what we need it to, at least structurally, so maybe we can stop here? End of the blog post everyone, we can all can go home! Not so fast. A trained eye can see how this formulation of a TrieMap can cause problems. Let's take a step into the real world, where this TrieMap doesn't meet the performance requirements of Redis Query Engine.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+        // The label will be a null-term string, so we always advance by at least one past.
+        + (n->labelLen + 1) 
+        // We also need to skip past the 'first character of each child' data.
+        + n->numChildren
+        // And now we have the address of the children, which we typecast to 'TrieMapNode **'
+    ));
+}
+```
+
+The above is an example of how we might access the children of a type like this. We know the lengths of each dynamic part of the datatype, so we can compute the offset. The original, pre-rust implementation of this can be found [here](https://github.com/RediSearch/RediSearch/blob/dcf009a2327240b24d6efcf100fed577b8a5eef0/deps/triemap/triemap.c#L18).
```
LukeMathWalker:
```suggestion
The above is an example of how we might access the children of a type like this. We know the lengths of each dynamic part of the datatype, so we can compute the offset. The original, pre-Rust implementation of this can be found [here](https://github.com/RediSearch/RediSearch/blob/dcf009a2327240b24d6efcf100fed577b8a5eef0/deps/triemap/triemap.c#L18).
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+}
+
+#[repr (C)]
+struct ChildRef<T> {
+    first_byte: u8, 
+    node: Node<T>,
+}
+```
+
+This implementation makes no attempt to match the performance characteristics of the C original, so we have extra heap allocations for labels and children to maintain the simplicity.
```
LukeMathWalker:
```suggestion
This implementation makes no attempt to match the performance characteristics of the C original, so we have extra heap allocations for labels and children to keep things simple.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+
+Building off of this naive implementation, we can design an external API that can be tested and benchmarked. This working, tested, naive version and refresh of what the original was doing lowers the cognitive load when it comes to the team implementing a later version.
```
LukeMathWalker:
```suggestion
Building off of this naive implementation, we can design an external API that can be tested and benchmarked. This working, tested, naive version reduces cognitive load for the team when implementing a more optimized version later on.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+
+We didn't focus on performance, because that wasn't what this initial information-gathering pass was about. But we might as well make the comparisons we can so we know how important the optimization work will be:
```
LukeMathWalker:
```suggestion
We didn't focus on performance, because that wasn't what this initial information-gathering pass was about. But we might as well make the comparisons so we know how important the optimization work will be:
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+
+![](/assets/images/posts/2026-04-XX-redis-triemap/violin-chart.svg)
+
+Our implementation was twice as slow as the original C implementation, but about half as slow as an off-the-shelf crate from crates.io. Our implementation was also far less consistent in its speed, whereas the C implementation had very little variance. All the instances of `Vec` we used also meant the memory usage of the naive implementation was double that of the original.
```
LukeMathWalker:
```suggestion
Our implementation was twice as slow as the original C implementation but about twice as fast as an off-the-shelf implementation from crates.io. Our implementation was also far less consistent in its speed, whereas the C implementation had very little variance. All the instances of `Vec` we used also showed that the naive implementation's memory usage was double that of the C original.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+
+This first step in the porting strategy left us with 1. Total test coverage of what we were working on 2. Knowledge of how subpar the performance of this naive TrieMap implementation and 3. A team that has a deeper understanding of the original implementation and the decisions behind it.
```
LukeMathWalker:
```suggestion
This first step in the porting strategy left us with:
1. Total test coverage of what we were working on 
2. Knowledge of how subpar the performance of this naive TrieMap implementation is
3. A team that has a deeper understanding of the original implementation and the decisions behind it.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+The layout we ended up designing is as follows:
+
+![](/assets/images/posts/2026-04-XX-redis-triemap/rust-layout.svg)
+
+There are a couple of differences between our Rust layout and the original C layout:
+
+1. The data structure is padded, so depending on the number of children there may be unused space between the end of `children_first_bytes` and `children`.
+2. Our optional leaf data sits at the end of the struct, rather than before the label and children.
+
+How are we supposed to implement this in Rust? We do it with designing our own Dynamically Sized Types.
```
LukeMathWalker:
```suggestion
How are we supposed to implement this in Rust? We do it with designing our own Dynamically Sized Type (DST).
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+
+### DIY Dynamically Sized Types
+
+We want nodes to be a Dynamically Sized Type, like `str` or `[u8]` (note the lack of `&`). Something whose size is not known at compile-time.
```
LukeMathWalker:
```suggestion
We want nodes to be a DST, like `str` or `[u8]` (note the lack of `&`). Something whose size is not known at compile-time.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+pub unsafe fn alloc(layout: Layout) -> *mut u8
+```
+
+Rather than C's `malloc` which takes a length in bytes, `alloc` takes a `Layout` type that represents the size and alignment of an allocation.
+
+A trivial `Layout` for a sized type `T` can be generated via `Layout::new::<T>()`, but `Layout` also exposes a relatively simple to use API:
+
+1. `Layout::new` produces a layout for a sized type (as already stated).
+2. `Layout::array` produces a layout for an array of a sized type, with `n` elements. 
+3. `Layout::extend` lets us "add" one layout to another, returning that composed layout and an offset for the rhs layout relative to the original layout. Order matters here.
```
LukeMathWalker:
```suggestion
3. `Layout::extend` lets us "add" one layout to another, returning that composed layout and an offset for the right-hand side layout relative to the original layout. Order matters here.
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+
+We can use this to manually build a runtime-defined layout that suits our Dynamically Sized Type's needs. If we look at our DST's layout needs again:
```
LukeMathWalker:
```suggestion
We can use this to manually build a runtime-defined layout that suits our DST. If we look at what our DST layout needs again:
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+
+1. `label_len` (static size, `u16`)
+2. `n_children` (static size, `u8`)
+3. `label` (dynamic size, `label_len` * `u8`)
+4. `children_first_bytes` (dynamic size, `n_children` * `u8`)
+5. `children` (dynamic size, `n_children` * `Node<T>`)
+6. `data` (static size, `Option<T>`)
```
LukeMathWalker:
```suggestion
1. `label_len` (known size, `u16`)
2. `n_children` (known size, `u8`)
3. `label` (dynamic size, `label_len` * `u8`)
4. `children_first_bytes` (dynamic size, `n_children` * `u8`)
5. `children` (dynamic size, `n_children` * `Node<T>`)
6. `data` (known size, dynamic offset, `Option<T>`)
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+
+We have 3 dynamically-sized fields and 3 statically-sized fields (note that by "static" we mean "known at compile-time" not the `'static` lifetime or the `static` keyword). So to build up our layout we can do the following:
```
LukeMathWalker:
```suggestion
We have 3 dynamically-sized fields and 3 fields with sizes known at compile-time. So to build up our layout we can do the following:
```

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
...
+
+And now that we have a way to express our layout needs at runtime, we can manually call `alloc` to retrieve a pointer.
+
+## Porting Strategy: Managing Necessary Unsafe
+
+Once we have our `Layout`, we can call `alloc` to get our pointer. This memory is uninitialized, but we can initialize it manually. Managing this memory manually only gets gnarly once we want to split or merge nodes.
+
+In the final codebase we were finally left with **128** unsafe blocks and **21** unsafe methods.
+
+Managing this unsafe code requires care and understanding for how maintainers will interact with it. On top of all this, the client is moving to Rust to minimize issues related to memory safety so it is important to build an API that can be trusted and can't be misused.
```
LukeMathWalker: Let's be consistent: either we write `unsafe` in monospaced font everywhere, or we omit the "`" everywhere.

src/posts/2026-04-XX-redis-triemap.md
```diff
@@ -0,0 +1,351 @@
+---
```
LukeMathWalker, line 1: A few general notes: we need links to external resources we mention (e.g. `HashMap`) as well as a link to the Rust implementation of our TrieMap (ideally early on in the article).

LukeMathWalker (approved):

marcoow:
Let's just use the talk's title for now: https://eurorust.eu/2025/talks/rewrite-optimize-repeat/

mainmatterlinnbp:
Here's a new title from Luca (FYI @marcoow): "Lesson learned while porting a TrieMap from C to Rust", which should reflect the content well.

PR #2858, opened 2026-05-14
feat: update "Truly Native Apps with Svelte?" blog post

paoloricciuti:
This updates the blogpost with the new information about the custom renderer api.


mainmatterlinnbp:
Thanks @paoloricciuti! Reading this, I realize it makes sense to hold off a bit here until the custom renderers are actually out and we can link to the next blog post. Otherwise we'll have to do this twice anyways. Also, it'd be great if this could just be a banner on top of the post or something so that it's clear it's an update. If not a banner then some other sort of visual representation that makes it clear that this is not just part of the original blog post.

PR #2860, opened 2026-05-25
chore: add mainmatter content review skill

paoloricciuti:
This adds the skill @KevinBongart wrote to the mainmatter.com repo so we can just run it locally when we are writing some blog post


chesedo:
This skill now has a home here -> https://github.com/mainmatter/claude-plugins

PR #2868, opened 2026-06-08
Update Blog consulting banners 

avollrath:
https://www.notion.so/mainmatter/Optimize-consulting-banners-on-the-MM-blog-37264e58ddfa8077a8f2c001f39e6d52?source=copy_link

### DIFF
--- a/src/assets/css/components/_cta-banner.scss
+++ b/src/assets/css/components/_cta-banner.scss
@@ -97,7 +97,7 @@
 
 .service-cta--blog {
   .cta-banner__wrapper {
-    padding: 2rem 6rem;
+    padding: 4rem 6rem;
   }
 
   .cta-banner__main {
--- a/src/assets/css/components/_cta-banner.scss
+++ b/src/assets/css/components/_cta-banner.scss
@@ -103,7 +103,6 @@
   .cta-banner__main {
     max-width: none;
     padding: 0;
-    
   }
 
   .cta-banner__title {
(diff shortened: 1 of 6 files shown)


mainmatterlinnbp:
Thanks André, this looks great! Just two minor things:

- The consulting page links don't really look clickable when they're bold, as a user I wouldn't understand that it's a link. Let's drop the bold and instead use underline. It might also make sense to only make the keywords clickable, as such: We can help with <u>Svelte training</u>, <u>Svelte 5 migrations</u>, <u>testing</u> and more. That'll make it even more obvious that this is a link out to the addressed cases. 
- Let's drop the "today" from the CTA
- A comma was missing from the Svelte banner. I just added it. It should be "We can help with Svelte training, Svelte 5 migrations, testing, and more."

avollrath:
@mainmatterlinnbp Thanks for the feeback, all changes here:
https://deploy-preview-2868--objective-northcutt-37494c.netlify.app/blog/2025/06/25/web-components-with-svelte/

mainmatterlinnbp:
Thanks for sharing this André, overall, this looks a lot cleaner and more clickable. The three separate links though didn't quite work out visually the way I'd hoped. Let's use one link per banner instead after all, in the following style:
- Svelte: We can help with <u>Svelte training, Svelte 5 migrations, testing</u>, and more.
- Rust: We can help with </u>complex C-to-Rust migrations, reliability issues, and skill gaps</u>. 
- Ember: We can help with <u>migrating to Vite, future-proofing your codebase, and multi-framework architectures</u>. 

Besides that we're good to go, thanks for your great work here! 👍

mainmatterlinnbp (approved):
Looks great! I think these are great improvements. Excited to see how the banner performs.

PR #2877, opened 2026-06-22
Add upskilling week event

umbrellarocket:


.eleventy.js
```diff
@@ -65,6 +65,10 @@ module.exports = async function (eleventyConfig) {
    */
   dayjs.extend(customParseFormat);
 
+  eleventyConfig.addFilter("monthDay", function (date) {
+    return dayjs(date).format("MMMM D");
+  });
```
marcoow, line 70: This is good but we should update most other upcoming entries (e.g. EuroRust) as well then

marcoow (approved):

mainmatterlinnbp (approved):

PR #2897, opened 2026-07-23
Add Rust Commercial Network badge

avollrath:
- add the RCN logo to the homepage marquee
- add the RCN badge to the Rust consulting page (in We Are Rust Experts section)


marcoow (commented):
Hm, just thinking that if we have this here, we could also add the fact that we're a Rust Foundation member and a trusted training provider - not sure it's worth adding the logo at all now…

What do you think @mainmatterlinnbp?

Image

mainmatterlinnbp:
Agree, it makes sense to include the RFTT badge, Rust foundation and Commercial network all together, on the homepage, mainmatter.com/rust and /training/rust.

marcoow:
I'm not actually sure about that. E.g. on the training page, our foundation or commercial network membership do not add anything on top of the TT program badge. On the Rust page, if we include the 3 badges, we could/should also include the EuroRust, 100 exercises, and maybe other badges… Or we just mention the trusted training program and the commercial network in the text as we do with e.g. EuroRust now.

mainmatterlinnbp:
> I'm not actually sure about that. E.g. on the training page, our foundation or commercial network membership do not add anything on top of the TT program badge. On the Rust page, if we include the 3 badges, we could/should also include the EuroRust, 100 exercises, and maybe other badges… Or we just mention the trusted training program and the commercial network in the text as we do with e.g. EuroRust now.

I understand the reasoning around RFTT. I'm not sure I follow however what's special about the Commercial network badge per se? It's not widely known, and I see no benefit for us as to really mention it in more places than with the other badges on the homepage.

marcoow:
let's just mention it in the text on the rust consulting page – it might not be widely known (yet) but it certainly sounds good which is a value in its own right

avollrath:
image

So no badges at all but a mention of the Commercial network in the text of the "We Are Rust Experts" section?

marcoow:
I'd say so, yes

src/rust-consulting.njk
```diff
@@ -95,6 +95,13 @@ og:
             maintaining several other
             <a href="#open-source" class="link-black">open-source&nbsp;projects</a>.
           </p>
+          <p class="medium">
+            We are also a founding member of the
```
avollrath: @marcoow 
Removed all badges and added the RCN mention to the text instead. Please check whether the copy needs any adjustments.

src/rust-consulting.njk
```diff
@@ -95,6 +95,13 @@ og:
             maintaining several other
             <a href="#open-source" class="link-black">open-source&nbsp;projects</a>.
```
marcoow:
```suggestion
            <a href="#open-source" class="link-black">open-source&nbsp;projects</a>. We are also a founding member of the <a href="https://rustfoundation.org/rust-commercial-network/" class="link-black"
              >Rust Commercial Network</a
            >, where we work with ecosystem members to solve shared problems and
            advance Rust adoption.
```

src/rust-consulting.njk
```diff
@@ -95,6 +95,13 @@ og:
...
           </p>
+          <p class="medium">
+            We are also a founding member of the
+            <a href="https://rustfoundation.org/rust-commercial-network/" class="link-black"
+              >Rust Commercial Network</a
+            >, where we work with companies and community members to share real-world experience and
+            advance Rust adoption.
+          </p>
```
marcoow:
```suggestion
```

Let's add this to the previous paragrph.

marcoow (approved):

PR #2898, opened 2026-07-24
Unify workshop card content widths

avollrath:
Removed the full-width layout for workshop cards without dates or prices.

This gives all workshop descriptions a consistent width and improves alignment across the training pages.

### DIFF
--- a/src/assets/css/components/_workshops-landing.scss
+++ b/src/assets/css/components/_workshops-landing.scss
@@ -127,11 +127,6 @@
   pointer-events: auto;
 }
 
-.workshop-card--full .workshop-card__main,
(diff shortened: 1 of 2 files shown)


marcoow:
Isn't this a general issue with the website that affects all kinds of pages. Or did I misunderstand you?

marcoow (approved):

avollrath:
Hm, maybe we misunderstood each other? 🤔

No other pages use a similar two-column layout, so this is specific to the training pages. That’s why I thought it made sense to address this small change directly.

This is unrelated to the inconsistent container widths on the website, if that’s what you were thinking of?

PR #2906, opened 2026-08-17
Add more Rust workshops

marcoow:
* https://github.com/mainmatter/rust-proc-macro-workshop
* https://github.com/mainmatter/async-rust-workshop
* https://github.com/mainmatter/advanced-rust-workshop

## TODO

- [x] add `og:image`s for each workshop
- [x] replace the hero and foot images for each workshop with sth. more fitting


marcoow:
There are still some TODOs for @avollrath but could you already review the contents @gdesmott and @hseeberger ?

src/workshops/rust-procedural-macros.md
```diff
@@ -0,0 +1,77 @@
+---
+title: "Procedural Macros: extend the compiler, one token at a time"
+tags: "rust"
+format: "1 day"
+subtext: "Bookable for teams – on-site or remote"
+description: This 1-day workshop will make you comfortable writing procedural macros in Rust. We cover all relevant topics from advantages and disadvantages of Rust, the kinds of procedural macros, toos like syn and quote, as well as error handling and testing.
```
gdesmott:
```suggestion
description: This 1-day workshop will make you comfortable writing procedural macros in Rust. We cover all relevant topics from advantages and disadvantages of Rust, the kinds of procedural macros, tools like syn and quote, as well as error handling and testing.
```

gdesmott (commented):
Looks good except the typo.

Maybe the topics list is a bit too long? We could squash some to get closer to the actual chapter splits, see https://rust-exercises.com/proc-macro/

marcoow:
@gdesmott sure, happy to do that but ideally you can make some concrete suggestions. I'm not actually familiar with the materials.

src/workshops/async-rust.md
```diff
@@ -0,0 +1,67 @@
...
+  <p>You have written async Rust before. You know what <code>async fn</code> and <code>.await</code> do, you have used Tokio, and you have at some point watched a program do something other than what you expected it to. This workshop is about the advanced async topics that come after the syntax: who owns your state, what happens when a future is dropped halfway through, and what your server does when it is asked for more than it can deliver.</p><p>You will turn the example <code>minidb</code>, a small in-memory key-value store, into a networked service. By the end of the workshop you will have added support for a line protocol over TCP, and built an architecture where one task owns the data and everything else asks it for what it needs. With that setup, a slow client cannot starve a fast one, an overloaded server says so instead of falling over, and a restart picks up where the last one left off.</p><p>The workshop is designed for developers who have written async Rust before. If you are new to Rust altogether, start with our <a href="/training/learn-rust-starting-from-scratch/">"Learn Rust, starting from scratch"</a> workshop for an introduction to the language.</p>
```
hseeberger:
```suggestion
  <p>You are comfortable with Rust: ownership, traits, <code>Arc</code>, and what <code>Send</code> means. Async is the part you are less sure of, whether you have written an <code>async fn</code> and left it there or watched a Tokio service in production do something other than what you expected. The workshop makes the vocabulary precise first, then spends its time on the questions that come after the syntax: who owns your state, what happens when a future is dropped halfway through, and what your server does when it is asked for more than it can deliver.</p><p>You will turn the example <code>minidb</code>, a small in-memory key-value store, into a networked service: a length-prefixed protocol over TCP with several requests in flight per connection, one task owning the data and everything else asking it, durability pluggable behind a trait, and tests that run the whole server on a simulated network. A slow client cannot starve a fast one, an overloaded server says so instead of falling over, and a restart picks up where the last one left off.</p><p>The workshop is designed for developers who are comfortable writing Rust and want to stop guessing when it comes to async. If you are new to Rust instead, start with <a href="/training/learn-rust-starting-from-scratch/">Learn Rust, starting from scratch</a>.</p>
```

src/workshops/async-rust.md
```diff
@@ -0,0 +1,67 @@
...
+      We will cover Tokio's paused clock, which makes a thirty second timeout testable in microseconds, <code>tokio::io::duplex</code> for exercising a connection handler without opening a socket, and asserting on properties rather than on scheduling order. You will also instrument the server with <code>tracing</code> and test the resulting spans as structured data.
+
+
+  - title: Durability across restarts
+    text: >
+      You will add a write-ahead log, recording every mutating request before it is applied, batching the syncs into a group commit so that a busy server does not make one trip to the disk per request, and replaying the log on startup so that a restart preserves the data.
+
+
+leads:
+  - handle: algo_luca
```
hseeberger, line 94: Should that be hseeberger?

src/workshops/advanced-rust.md
```diff
@@ -0,0 +1,67 @@
...
+      We will cover static and dynamic dispatch, <code>dyn</code> compatibility and the signatures it rules out, and how to keep a generic function thin so that monomorphisation does not duplicate its body once per type. You will also seal a trait, and we will cover when a closed set is better expressed as an enum.
+
+
+  - title: PhantomData, variance, and brands
+    text: >
+      We will cover what <code>PhantomData</code> does to size, auto traits, drop checking, and variance. You will give a handle that owns its data the borrowing behaviour of a reference, then use a branded lifetime and a higher-ranked bound so that keys built against one store cannot be used with another.
+
+
+leads:
+  - handle: algo_luca
```
hseeberger, line 64: hseeberger?

src/workshops/async-rust.md
```diff
@@ -0,0 +1,67 @@
...
+
+
+  - title: Testing async code
+    text: >
+      We will cover Tokio's paused clock, which makes a thirty second timeout testable in microseconds, <code>tokio::io::duplex</code> for exercising a connection handler without opening a socket, and asserting on properties rather than on scheduling order. You will also instrument the server with <code>tracing</code> and test the resulting spans as structured data.
+
+
+  - title: Durability across restarts
+    text: >
+      You will add a write-ahead log, recording every mutating request before it is applied, batching the syncs into a group commit so that a busy server does not make one trip to the disk per request, and replaying the log on startup so that a restart preserves the data.
```
hseeberger, line 60: The topic list covers only half of the material; here is the full agenda. The existing topics are unchanged apart from the `Pin` sentence, which now says how far the fundamentals go on first contact and that the later chapters return to the poll level.

```suggestion
topics:
  - title: The Tokio runtime
    text: >
      We will cover what Tokio adds on top of the language, why a future does nothing until something polls it, and why <code>.await</code> is a suspension point rather than a call, along with as much of <code>Future</code>, <code>poll</code>, and <code>Pin</code> as reasoning about the runtime demands; a later chapter returns to them at the poll level.


  - title: Tasks
    text: >
      <code>tokio::spawn</code> turns a future into a task, which is not a thread and does not behave like one. We will cover the <code>Send + 'static</code> requirement, running work concurrently with <code>join!</code> and <code>JoinSet</code>, and moving work that cannot yield onto <code>spawn_blocking</code> so that it does not occupy the runtime's worker threads.


  - title: Building the server
    text: >
      You will turn <code>minidb</code> into a TCP server that speaks a line protocol, covering the accept loop, framing with <code>BufReader</code> and <code>lines()</code>, and one task per connection. We will also cover which errors should end a single connection and which should not reach the accept loop.


  - title: Who owns the state
    text: >
      Every connection is a task, and each of them needs access to the same store. You will build both answers, <code>Arc&lt;Mutex&lt;Store&gt;&gt;</code> and an actor task that owns the data outright, and we will cover which mutex to use and where its guard may not be held. The section ends with a criterion benchmark comparing the two approaches under contention.


  - title: Cancellation
    text: >
      In async Rust, cancelling work means dropping a future, which can happen at any suspension point. We will cover <code>timeout</code>, <code>select!</code>, and <code>CancellationToken</code>, the state a dropped future leaves behind, and cancel safety: why <code>next_line</code> can be dropped safely and a hand-written read loop cannot.


  - title: Backpressure
    text: >
      An unbounded queue grows until the process runs out of memory. We will cover bounded <code>mpsc</code> channels, the difference between waiting, refusing, and dropping work when a queue is full, and limiting concurrent connections with a <code>Semaphore</code>, where the point at which the permit is acquired determines what the limit protects.


  - title: Shutdown and supervision
    text: >
      Without a shutdown path, a deploy terminates requests that are still in flight. We will cover graceful shutdown with <code>CancellationToken</code> and <code>TaskTracker</code>, the ordering that drains connections before the store task, and how to decide what should happen when a task fails without being asked to stop.


  - title: Testing async code
    text: >
      We will cover Tokio's paused clock, which makes a thirty second timeout testable in microseconds, <code>tokio::io::duplex</code> for exercising a connection handler without opening a socket, and asserting on properties rather than on scheduling order. You will also instrument the server with <code>tracing</code> and test the resulting spans as structured data.


  - title: Durability across restarts
    text: >
      You will add a write-ahead log, recording every mutating request before it is applied, batching the syncs into a group commit so that a busy server does not make one trip to the disk per request, and replaying the log on startup so that a restart preserves the data.


  - title: AsyncRead and AsyncWrite
    text: >
      We will go underneath all of Tokio's IO, to the traits it is built on. We will cover the <code>poll_read</code> and <code>poll_write</code> contract and <code>ReadBuf</code>, and you will write an IO wrapper of your own that counts every byte passing through the server, dropping into the connection handler without changing a signature.


  - title: Codecs and framing
    text: >
      You will replace the hand-rolled line loop with a length-prefixed frame, implementing <code>Decoder</code> and <code>Encoder</code> and driving them with <code>Framed</code>. The same codec then carries the write-ahead log, and a restriction the old framing forced on values disappears. This also closes the loop on cancel safety: a partial frame lives in the codec's buffer, not on a dropped future's stack.


  - title: Streams, sinks, and pipelining
    text: >
      We will cover the <code>Stream</code> and <code>Sink</code> traits and splitting a connection into independently owned halves. You will then let one connection have several requests in flight at once, choosing the in-flight bound and keeping the responses in order deliberately rather than by accident.


  - title: Async functions in traits
    text: >
      You will make the store's durability pluggable behind a trait with async methods, and meet the compile error that made async fn in traits famous: the spawned actor demands <code>Send</code> futures the trait does not promise. We will cover <code>impl Future</code> in return position as the fix, and what <code>dyn</code> does and does not allow.


  - title: Diagnosing a stuck runtime
    text: >
      A blocked worker thread does not announce itself. You will find one with <code>tokio-console</code>, and we will cover the runtime's metrics as well as the bridges between sync and async code, <code>block_in_place</code> and <code>Handle::block_on</code>, and the ways they go wrong quietly.


  - title: Testing against a hostile network
    text: >
      You will run the whole server on a simulated network with <code>turmoil</code>: injected latency, dropped connections, and partitions, all deterministic and all inside a unit test, using the same abstractions that made the store pluggable.
```

src/workshops/async-rust.md
```diff
@@ -0,0 +1,67 @@
+---
+title: "Async Rust: concurrency you can reason about"
+tags: "rust"
+format: "2 days"
+subtext: "Bookable for teams – on-site or remote"
+description: A 2-day workshop for developers who have written async Rust and want to stop guessing, covering Tokio, cancellation, backpressure, and graceful shutdown.
```
hseeberger: The audience here should match the intro, and the list drops the two chapters with the most time in them.

```suggestion
description: A 2-day workshop for Rust developers who want to stop guessing about async, covering Tokio, state ownership, cancellation, backpressure, graceful shutdown, and durability.
```

hseeberger (changes requested):

src/workshops/advanced-rust.md
marcoow: we'll stick with the full-time team for now

src/workshops/rust-procedural-macros.md
```diff
@@ -0,0 +1,77 @@
...
+
+
+  - title: Attribute macros
+    text: >
+      Unlike a derive, an attribute macro replaces the item it is attached to. We will cover wrapping a function, re-emitting the original item on failure, and parsing arguments with <code>darling</code>'s <code>FromMeta</code> to build <code>#[retry(times = 3, delay_ms = 100)]</code>.
+
+
+  - title: "Capstone: a state machine macro"
+    text: >
+      You will build one macro that uses everything: code generation over an enum, custom attributes declaring transitions and the initial state, invalid transitions rejected at compile time, and hygienic identifiers.
```
gdesmott:
```suggestion
topics:
  - title: Introduction
    text: >
      Rust has three kinds of procedural macros, and you have used all of them already. We will look at derive, function-like, and attribute macros in the wild, expand <code>thiserror</code> with <code>cargo expand</code> to see what they generate, and compare macros against plain functions and <code>macro_rules!</code> so you know when a procedural macro is the right call – and when it is not.


  - title: The proc-macro toolkit
    text: >
      Time to get your hands dirty. You will set up a proc-macro crate, write a minimal derive macro, and learn what a macro actually receives and returns: tokens and token streams, why <code>proc-macro2</code> makes your code testable, parsing Rust code with <code>syn</code>, and generating it with <code>quote</code>. We will also cover the re-export pattern used by <code>serde</code> and <code>thiserror</code>, and how to test a macro with <code>trybuild</code>.


  - title: Derive macros
    text: >
      Where the real work happens: handling every struct field layout and enums, absolute paths and hygienic identifiers so your output compiles in anyone's crate, error reporting from <code>panic!</code> to <code>compile_error!</code> to properly spanned <code>syn::Error</code>s, container and field attributes by hand and then with <code>darling</code>, and generics with lifetimes and <code>where</code> clauses.</code>.


  - title: Function-like macros
    text: >
      Whatever sits between a macro's delimiters is yours to interpret, even if it is not valid Rust. We will cover why <code>println!</code> has to be a macro, how to parse arbitrary token input, how to define your own syntax, and when to prefer this over a declarative macro. You will finish by building a small DSL: a <code>routes!</code> macro.


  - title: Attribute macros
    text: >
      Unlike a derive, an attribute macro receives the annotated item and returns its replacement. We will cover the parse-tweak-re-emit loop on a function, graceful error handling by re-emitting the original item to avoid cascading errors, and parsing attribute arguments with <code>darling</code>'s <code>FromMeta</code> to build a <code>#[retry(times = 3, delay_ms = 100)]</code> attribute.


  - title: "Putting it all together"
    text: >
      A capstone that draws on the whole day: a <code>#[derive(StateMachine)]</code> macro. Code generation over an enum's variants, custom helper attributes declaring the initial state and the allowed transitions, invalid transitions rejected at compile time with the error spanned on the offending token, and generated identifiers that never clash.

```

gdesmott: @marcoow : something like this then?

PR #2913, opened 2026-09-18
feat: local visual regression testing blog post

paoloricciuti:
[RENDERED VERSION](https://deploy-preview-2913--objective-northcutt-37494c.netlify.app/blog/2026/10/01/local-visual-regression-testing/)


src/posts/2026-10-01-local-visual-regression-testing.md
```diff
@@ -0,0 +1,293 @@
...
+    platform: linux/amd64
+    working_dir: /app
+    volumes:
+      - .:/app
+      - /app/node_modules
+```
+
+This `docker-compose.yml` does the following things:
+
+- It pins the platform to linux/amd64.
```
BobrImperator: That shouldn't be necessary. The `Dockerfile` playwright always uses `ubuntu` IIRC.
It could be somethign I don't understand about Mac however

src/posts/2026-10-01-local-visual-regression-testing.md
```diff
@@ -0,0 +1,293 @@
...
+- It pins the Playwright version (we would need to manually keep this in sync whenever we update our dependencies).
+- It mounts two volumes: one that keeps our project in sync with `/app` so that we can make changes to the project without having to rebuild the image, and another that keeps the container's `node_modules` around after the container has been shut down. This keeps the `node_modules` from our machine from leaking into the container.
```
BobrImperator: I think we should recommend using a bind volume for the node_modules instead. Using `/app/node_modules` will result in the module being recreated over and over I believe and at some point `docker system prune -a` would remove hundreds of gigabytes 😅 

```yaml
    volumes:
      - .:/app
      - /tmp/visual-regression:/app/node_modules
```
This should send the Container's `app/node_modules` to the semi-durable `/tmp` directory so it will stay durable until system reboots or gets rid of it another way.

src/posts/2026-10-01-local-visual-regression-testing.md
```diff
@@ -0,0 +1,293 @@
...
+:root {
+  /** other styles */
+  --font-display: ui-monospace, "SF Mono", "Cascadia Code", "JetBrains Mono",
+    Consolas, monospace;
+  --font-body: -apple-system, BlinkMacSystemFont, "Segoe UI", "Helvetica Neue",
+    Arial, sans-serif;
+}
+```
+
+Do you notice something? Yup, a lot of those are system fonts on macOS, which are missing on Linux. This is why you need consistency.
```
BobrImperator: It's not just the fonts. The same browser and the same fonts will render differently on different platforms either way. Which is why visual testing tools must capture the baseline images for each platform separately. Which in turn is why we're running in Docker so there's only 1 stable platform to test against.
I don't know the exact details why is that but hardware architectures themselves are inconsistent so it is what it is 😅

paoloricciuti: Oh yeah this was just referring to what happened as one of the reasons why you need consistencies...if it's not clear I can rephrase that.

BobrImperator: I take no issue with that :+1: Only the fact it kind of reads as if "system fonts" were the problem while unfortunately there's a plethora of things that _inevitably_ cause the discrepancies :sweat_smile:

src/posts/2026-10-01-local-visual-regression-testing.md
paoloricciuti: In this case, the important bit is `amd64` to make sure that the architecture also matches otherwise it would use `arm64` an Apple Silicon...maybe we could specify it in the bullet point tho.

src/posts/2026-10-01-local-visual-regression-testing.md
paoloricciuti: In the command we run with `--rm` so the volume is actually destroyed when the container is unmounted. However, we could also clarify this in the bullet point, so people don't mistakenly remove that and ends up with a bunch of node modules lol

src/posts/2026-10-01-local-visual-regression-testing.md
paoloricciuti: Yeah will try to change it so that it's clear it's not just that

src/posts/2026-10-01-local-visual-regression-testing.md
```diff
@@ -0,0 +1,293 @@
...
+description: "Give your agents a window into your application and yourself peace of mind"
+autoOg: true
+tagline: <p>Give your agents a window into your application and yourself peace of mind</p>
+---
+
+The first time I tried the `frontend-design` skill from Anthropic, I was amazed by the designs it could generate. When working on a greenfield personal project, I can rarely produce nice designs because, while I can reproduce a design when given one, coming up with a design of my own isn't my strongest suit.
+
+Luckily, when doing serious work, you'll always have a designated designer, and there's a reason for that: design consistency matters in a product. Having a dashboard or a public-facing page constantly changing makes your users' experience jarring.
+
+This is why tools like Percy and Chromatic exist: on every PR they'll diff the updated UI against the last known UI and let the designer stamp their approval on the changes.
```
marcoow: That's not the main use in my experience? The main use is detecting design regressions?

src/posts/2026-10-01-local-visual-regression-testing.md
```diff
@@ -0,0 +1,293 @@
...
+
+In an era where the changes were made by a human who kept a browser page open to check their work before opening a PR, this was fine (maybe not the best experience, but definitely fine). But something changed somewhat recently: we now build with our agents. A few of them have vision capabilities, and when properly set up they can also "keep a browser open", but both of these capabilities are different from a human looking at the page as a whole. We need something more: we need to make the loop tighter.
```
marcoow: Even a human who keeps a browser open during work cannot detect regressions that way as those would appear **elsewhere** (e.g. on other pages). In order to preview UI changes, I'd guess the agent can use Chrome developer tools to e.g. take a screenshot or use some kind of screen recording tool and analyze the video?

marcoow (commented):
I didn't read the entire post yet but the intro seems to indicate some fundamentally different understanding of what visual regression testing is for?

paoloricciuti:
> I didn't read the entire post yet but the intro seems to indicate some fundamentally different understanding of what visual regression testing is for?

I just needed a way to introduce the argument, but I think you are right, I should expand on that to also include the bigger benefits of catching regressions in an automated way.

paoloricciuti:
@marcoow @BobrImperator pushed a few commits to rephrase a few things

src/posts/2026-10-01-local-visual-regression-testing.md
```diff
@@ -0,0 +1,293 @@
+---
+title: "Local visual regression testing"
+authorHandle: paoloricciuti
+tags: [mcp, ai]
+customCta: "global/svelte-cta.njk"
+bio: "Paolo Ricciuti, Senior Software Engineer"
+description: "Give your agents a window into your application and yourself peace of mind"
+autoOg: true
+tagline: <p>Give your agents a window into your application and yourself peace of mind</p>
```
marcoow, line 9: Let's find a better intro that summarizes what the text is about

src/posts/2026-10-01-local-visual-regression-testing.md
```diff
@@ -0,0 +1,293 @@
...
+
+Luckily, when doing serious work, you'll always have a designated designer, and there's a reason for that: design consistency matters in a product. Having a dashboard or a public-facing page constantly changing makes your users' experience jarring.
+
+This is why tools like Percy and Chromatic exist: on every PR they'll diff the updated UI against the last known UI and let the designer stamp their approval on the changes while simultaneously giving you a red check in case your change introduced some regressions.
+
+In an era where the changes were made by a human who kept a browser page open to check their work before opening a PR, this was fine (maybe not the best experience, but definitely fine). But even then a change introduced by mistaken in a different page was difficult to catch and that we build with our agents the problem is even bigger. A few of them have vision capabilities, and when properly set up they can also "keep a browser open", but even if your agent has both of these capabilities you still have no guarantee that a change would not go unnoticed. We need something more: we need to make the loop tighter.
```
marcoow: I still don't think this is a good entry – the point of visual regression testing is that it's hard to know whether a particular change you make only changes the UI of your system in the way you intend or has other, unintended changes. It does not matter for that whether the design comes from a Claude skill or a human designer.

src/posts/2026-10-01-local-visual-regression-testing.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Visual regression testing
+
+As we said, tools like Percy and Chromatic are good when it comes to a human signing off on the work after it's finished. However, there are a few problems with these tools:
```
marcoow, line 32: I don't think that's the main point – see bove

marcoow (commented):
The main point of visual regression testing really is regressions – hence the name

src/posts/2026-10-01-local-visual-regression-testing.md
paoloricciuti: I don't think that is the point that is expressed tho...it's a way to start the post but not the only point no? IMHO the main point there is

> Luckily, when doing serious work, you'll always have a designated designer, and there's a reason for that: design consistency matters in a product.

More specifically 

> design consistency matters in a product

The fact that I start by talking about claude designs it's irrelevant to the point no?

src/posts/2026-10-01-local-visual-regression-testing.md
paoloricciuti: A human signing off on the work after it finished is not necessarily the designer here: it's someone that verify that there's no regressions thanks to Chromatic/Percy no?

marcoow: So the point of visual regression testing tools is not to review a design that was implemented and check it was good but to detect regressions which are unintended changes, typically on other pages/screens than the one that you actually worked on.

src/posts/2026-10-01-local-visual-regression-testing.md
marcoow: > The fact that I start by talking about claude designs it's irrelevant to the point no?

I think so tbh – it doesn't matter here where a design comes from, it's about understanding what visually changes as a consequence of a code change (in some cases you would not actually expect anything to visually change but then the visual regression testing tool tells you there's a change anyway)

src/posts/2026-10-01-local-visual-regression-testing.md
BobrImperator: > So the point of visual regression testing tools is not to review a design that was implemented and check it was good

I don't think that's what that paragraph says. In general how it usually works with Percy integrations and such (which is what this describes I think) is once you're done with a PR and it gets merged, then a visual snapshot is created to later regression test against (what you're saying).

And on a related note: With Playwright which can run locally, you can have a workflow where you create the design, take a snapshot locally and can use it to validate your further changes e.g. when refactoring a view and such.

I'd not get hooked up on the following though:
> typically on other pages/screens than the one that you actually worked on

paoloricciuti: > So the point of visual regression testing tools is not to review a design that was implemented and check it was good but to detect regressions which are unintended changes, typically on other pages/screens than the one that you actually worked on.

I personally think they are both...but what I mean is that even checking for visual regressions in separate pages is "a human signifying this off" because if there's a change you have to approve that it was intended

src/posts/2026-10-01-local-visual-regression-testing.md
paoloricciuti: >  think so tbh – it doesn't matter here where a design comes from, it's about understanding what visually changes as a consequence of a code change (in some cases you would not actually expect anything to visually change but then the visual regression testing tool tells you there's a change anyway)

I agree about this: but me mentioning a "claude design" is not meant to say that it only works with claude. It's just a way to introduce the argument. If it's confusing I can change it tho

marcoow: I think it's slightly confusing as it points in a different direction

paoloricciuti:
@marcoow I've rewritten the intro if you want to take a look

marcoow:
I still don't get it tbh. If you want to start with a tangible example for what visual regression testing catches, it'd be sth. like you intend to make a small visual change on one page but without you being aware, that changes something on all other pages as well.

src/posts/2026-10-01-local-visual-regression-testing.md
```diff
@@ -0,0 +1,303 @@
...
+
+This is why it's important to have a good visual regression story. We want to catch unintended regressions before actually shipping them to our users.
+
+This is why tools like Percy and Chromatic exist: on every PR they'll diff the updated UI against the last known UI and let the designer/developer stamp their approval on the changes while simultaneously giving you a red check in case your change introduced some regressions.
+
+In an era where the changes were made by a human who kept a browser page open to check their work before opening a PR, this was fine (maybe not the best experience, but definitely fine). But even then a change introduced by mistaken in a different page was difficult to catch and now that we build with our agents the problem is even bigger. A few of them have vision capabilities, and when properly set up they can also "keep a browser open", but even if your agent has both of these capabilities you still have no guarantee that a change would not go unnoticed. We need something more: we need to make the loop tighter.
```
marcoow: > In an era where the changes were made by a human who kept a browser page open to check their work before opening a PR, this was fine

No, the point has always been that you cannot test all possible regressions manually (which is the point of all regression testing – it's prohibitively expensive to test for them manually).

src/posts/2026-10-01-local-visual-regression-testing.md
```diff
@@ -0,0 +1,303 @@
...
+
+## Visual regression testing
+
+As we said, tools like Percy and Chromatic are good when it comes to a human signing off on the work after it's finished. However, there are a few problems with these tools:
+
+- The feedback loop is slow: you have to finish your work, commit, open a PR, wait for the diff to happen, visit the page (which might not even be optimized for agents), and only then can you see what changed.
+- With the productivity agents add, taking that many screenshots can become costly, and you don't want to slow down just because of cost.
```
marcoow: can we have a back-of-the-envelope calculation?

marcoow (commented):
I'd be interested in some thoughts around how this can effectively be made part of the feedback loop. How can an agent look at and understand the visual differences and decide which ones are intended and which ones aren't?

src/posts/2026-10-01-local-visual-regression-testing.md
paoloricciuti: That's what the next line is about no?

> But even then a change introduced by mistaken in a different page was difficult to catch and now that we build with our agents the problem is even bigger.

src/posts/2026-10-01-local-visual-regression-testing.md
paoloricciuti: It really depends on the volume of the website and increase in PR open but I'll try to calculate something

paoloricciuti:
> I still don't get it tbh. If you want to start with a tangible example for what visual regression testing catches, it'd be sth. like you intend to make a small visual change on one page but without you being aware, that changes something on all other pages as well.

What you don't get? Isn't this line talking exactly about that?

> This is why it's important to have a good visual regression story. We want to catch unintended regressions before actually shipping them to our users.

Or would you introduce the argument with an example like that?

src/posts/2026-10-01-local-visual-regression-testing.md
marcoow: Yes but this is still wrong:

> In an era where the changes were made by a human who kept a browser page open to check their work before opening a PR, this was fine

Visual regression testing tools solve a problem that exists in the same way when humans write the code.

marcoow:
I find the intro confusing. A website that looks bad and/or is design by AI is not necessarily related to the topic of visual regressions and testing for them. It can be completely free of visual regressions – of course it can also have some but my point is starting the post with this is confusing as it's not clear how that relates to the topic.

If you want to start with a tangible example (which I understand and agree makes sense), why not start with an example for a visual regression?

paoloricciuti:
> If you want to start with a tangible example (which I understand and agree makes sense), why not start with an example for a visual regression?

My point is: why do you need visual regression? Because consistency in a website it's important. But I guess I can start with an example where a change in one part changes an unrelated part

marcoow:
I think consistency is related but not the main point – the main point is you want to know what you're changing and the problem with regressions is you're changing something that you're unaware of that you're changing it

paoloricciuti:
@marcoow new try...I'll try to add a few numbers tomorrow

src/posts/2026-10-01-local-visual-regression-testing.md
```diff
@@ -0,0 +1,293 @@
...
+
+But then you find out that in a different page someone used `variant="login"` because it looked fine. And it was until you changed it. This is just an example of how things can go wrong if you don't have proper visual regression tests.
+
+This is why tools like Percy and Chromatic exist: on every PR they'll diff the updated UI against the last known UI and let the designer/developer stamp their approval on the changes while simultaneously giving you a red check in case your change introduced some regressions.
+
+Now that we build with our agents the problem is even bigger. A few of them have vision capabilities, and when properly set up they can also "keep a browser open", but even if your agent has both of these capabilities you still have no guarantee that a change would not go unnoticed. We need something more: we need to make the loop tighter.
+
+## Visual regression testing
+
+As we said, tools like Percy and Chromatic are good when it comes to a human signing off on the work after it's finished. However, there are a few problems with these tools:
```
marcoow: Can Percy and Chromatic be used by agents at all? E.g. do they have MCP servers or CLIs?

src/posts/2026-10-01-local-visual-regression-testing.md
```diff
@@ -0,0 +1,293 @@
...
+
+But wait, do you notice something? There's a change we didn't ask for. Claude inadvertently removed the glitch effect around the "Pixel Perfect" header when we only asked for a change in the accent color.
+
+That's the power of visual regression testing.
+
+We can now ask Claude to revert that change and run the tests again.
+
+As expected, the tests fail again (since the accent color did change), but if we take a look at the diff, it's now a much more reasonable one.
+
+![a diff of the expected page vs the current took by playwright tests](/assets/images/posts/2026-10-01-local-visual-regression-testing/diff-after.png)
```
marcoow: Can we change the screenshot so that the main diff you want to show isn't cut off?

src/posts/2026-10-01-local-visual-regression-testing.md
```diff
@@ -0,0 +1,293 @@
...
+
+Now that we've verified the changes are the ones we expected, we can run `npx playwright test --update-snapshots` to update the old snapshots and have a new ground truth for our homepage.
+
+The best part is that these are just files on disk, which means that the agent can already autonomously read them to figure out what the problem is.
```
marcoow: Can we have a special review subagent that looks at these files? How does that work?

marcoow (commented):
I think the intro works now and is less confusing

src/posts/2026-10-01-local-visual-regression-testing.md
paoloricciuti: So Percy has an MCP, Chromatic does not, I've asked a friend that works there, and he said he'll get back to me. Obviously, both can work with a browser MCP to access the page if they have access to it (so it could technically work anyway)...do you think we should add info about this or is just to confirm the statement?

src/posts/2026-10-01-local-visual-regression-testing.md
paoloricciuti: Agents generally do this by themselves with no need of special subagents...obviously depending on the setup it could help having a specialized subagent but I don't think it's necessary.

src/posts/2026-10-01-local-visual-regression-testing.md
marcoow: You could say that at least Chromatic can't be used by agents directly at all (at least not straight forward so you can't really make it part of the agent's feedback loop at all.

src/posts/2026-10-01-local-visual-regression-testing.md
marcoow: It's not necessary but review subagents seem to be pretty common? Also, I don't think this is trivial anyway – obviously for a lot of changes you'd expect some of the screenshots to change but the question is which ones and more specifically which parts of which ones (e.g. only the buttons, not the headline in your example)? I don't think just telling the agents "make sure the visual changes in the screenshots make sense in relation to the code changes you made" is going to be it?

paoloricciuti: > It's not necessary but review subagents seem to be pretty common?

They definitely are but this is exactly why I don't think they fit in the blog post: it's not really related to visual regression testing. We can obviously hint that you can write a subagent with specific instructions to look at the diff and find out if the change is expected or not.

> Also, I don't think this is trivial anyway – obviously for a lot of changes you'd expect some of the screenshots to change but the question is which ones and more specifically which parts of which ones (e.g. only the buttons, not the headline in your example)? I don't think just telling the agents "make sure the visual changes in the screenshots make sense in relation to the code changes you made" is going to be it?

Tbf I think it should be enough: those are effectively tests and models are well-trained to make the tests pass so when they will fail they will figure out why they are failing. Newer models are also pretty good at inferring from the original prompt what's the expected output. And tbf there's not much more you can do in the subagent that is not explaining better "make sure the visual changes only relate the original request".

That said we can defintely add a paragraph that talks about the subagent if you prefer.

src/posts/2026-10-01-local-visual-regression-testing.md
paoloricciuti: I'll wait until they get back to me but then yes I'll write that if they say they don't have a way

